PR #3, opened 2015-07-31
add button to copy version info in first level about box

chozabu:


chozabu:
@csoler - may as well poke you on this one too, I'm happy for it to be refused if it is not a desired feature.

PR #7, opened 2015-08-02
Fix Debian build scripts to build libsqlcipher together with Wheezy, …

heini:
…Jessie and Ubuntu Precise packages.

This does essentially the same as in former times for all distribtions (I know of) that don't have native libsqlcipher packages.


csoler:
not needed for precise, since the ppa has an extra ppa dependency (guardianproject ppa), which provides sqlcpher as a package

For jessie, it's needed.

heini:
Doen's seem to get merged, closing.

PR #8, opened 2015-08-02
Auto version and packaging

chozabu:
Not sure this works for deb/buntu package - may not be best method. builds on previous PR


chozabu:
Closing, this patch is somewhat outdated, though it's concepts still hold:
- have one script to get version numbers - so there is not duplicated code in lib+gui+deb-pkg
- don't alter files under VCS - makes it trivial to have correct version numbers in dev builds.
- default to inserting version numbers in build

Issue #13, opened 2015-08-05
Should 4096 bits PGP Key Length be the default?


ghost:
Considering that Retroshare is moving toward working with TOR networking, it make sense to set PGP Key Length for the highest allowable bit security possible despise some of it's disadvantages like increasing computational cost and storage size. What your thought on this?

electron128:
i'm closing this, see #15

PR #15, opened 2015-08-05
A Proposal Pull Request for Default PGP Key Length

ghost:
It's a pullrequest for issue #13


csoler:
2048 bits is the current standard for PGP and is by far sufficient for securing your data. 3072 and 4096 are available as options for the most paranoid users, but shouldn't be the default.

PR #18, opened 2015-08-05
Use separate build directory for building

AsamK:
qmake supports out of directory builds. That way all generated files are contained in a single build directory and not mixed with the source code.

This also allows to have multiple build directories, e.g. buildDebug and buildRelease.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,4 @@ qrc_*.cpp
 ui_*.h
 Makefile.*
 *~
-
+/build/
--- a/README.txt
+++ b/README.txt
@@ -17,14 +17,15 @@ Compilation on Ubuntu/Debian
 
 3 - compile
 
-   # cd trunk
-   # qmake CONFIG=debug
+   # mkdir build
+   # cd build
+   # qmake CONFIG=debug ../trunk
    # make
 
    => the executables produced will be 
 
-         trunk/retroshare-gui/src/Retroshare
-         trunk/retroshare-nogui/src/retroshare-nogui
+         build/retroshare-gui/src/Retroshare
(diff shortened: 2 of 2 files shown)


thunder2:
I use shadow build too but outside the source tree so you don't need to add the build directory to .gitignore.

AsamK:
That’s possible, too, also qtcreator uses a build directory outside the src tree by default.
When building myself, I prefer however to use a build directory inside the tree so all is contained in one folder.

As the README suggests to create an additional parent folder, I could change the commit to create a build folder next to the trunk folder.
However I'd still like to add the build directory to .gitignore ;)

csoler:
I don't really see the point in building in a separate directory. .o and .ui already go into a separate directory. What would you gain beyond complicating things a little more?

csoler:
As you say, it's possible to tweak qtcreator for that, which anyone who wants this can do easily.
As far as debian packaging is concerned, changing the build directory would need to carefully update all scripts.

AsamK:
I'm not suggesting to change it for building packages, just for users who download the source and want to compile it themselves. Also deleting all generated files is as easy as "rm -r build".
git clean could do that too, but you would have to be careful not to delete wanted files.

csoler:
make distclean, from trunk, does exactly this.

sehraf:
some others may call mixing source and build dir complicate - it's good practice to separate source code and build directory 

but in the end it's a matter of taste

AsamK:
yeah, it's a matter of taste. Though I really don't like mixing build and source directory ;)

AsamK:
Closing as it won't be merged and doesn’t apply cleanly anymore.

PR #21, opened 2015-08-07
Store lobby history on disk

AsamK:
RS didn't store the lobby history on disk, so it was unavailable after a restart.
This seems to have been intentional, however after removing the appropriate if clause, saving the history works without problems.
The only thing is that, the history can only be accessed after rereceiving the lobby id from another node and joining it. But that’s another issue, that lobbies are not saved.

The second commit adds resolving of nicks from the gxsid, when restoring the chat lobby history or viewing it in the history browser.

### DIFF
--- a/libretroshare/src/pqi/p3historymgr.cc
+++ b/libretroshare/src/pqi/p3historymgr.cc
@@ -115,11 +115,6 @@ void p3HistoryMgr::addMessage(const ChatMessage& cm)
         item->sendTime = cm.sendTime;
         item->recvTime = cm.recvTime;
 
-        if (cm.chat_id.isLobbyId()) {
-			// disable save to disc for chat lobbies until they are saved
-			item->saveToDisc = false;
-		}
-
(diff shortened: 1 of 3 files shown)


SarcoRS:
> The second commit adds resolving of nicks from the gxsid, when restoring the chat lobby history or viewing it in the history browser.

Is this also about retroactively replacing IDs with nicknames in lobby windows, when somebody joins the lobby, posts something, but his name gets resolved only later and it's very confusing to figure out who posted that without explicitly looking at IDs of people in the lobby?

AsamK:
No, that's not done, I guess that it would be difficult to safely replace the IDs in the chat windows retroactively, as that's just a text box with some formatting. What could be done is sending some "ID is now known as Nick" message, when resolving is done, but that's not done by this patch.

However with this patch, the history browser will show you all messages with resolved nicks, even those that weren't resolved in the main lobby window before.

PR #25, opened 2015-08-13
jensters patch on pqissl.cc

cavebeat:


chozabu:
I think we are better off with this patch than without?
Must admit - have not looked in detail - but if it stops RS from reading an invalid socket, that sounds good!

csoler:
I merged it manually to add some additional changes.

cavebeat:
https://github.com/RetroShare/RetroShare/commit/2e08dde32be689e33f755f944abcf12a3c223445

PR #32, opened 2015-08-14
libresapi - Forums

chozabu:
the python test code:

```
print "FORUM"
r = requests.put("http://0.0.0.0:9090/api/v2/forums")
for f in r.json()['data']:
    if f['own']:
        print f
```

results in:

```
FORUM
{u'own': True, u'name': u'another test forum', u'visible_msg_count': u'0', u'pop': u'0', u'pgp_linked': False, u'group_status': u'0', u'parent_grp_id': u'00000000000000000000000000000000', u'author_id': u'6f312e1c10163d6261e02ada6067bbb9', u'id': u'720959e1d78edd39d8682c07fc5a8fe1', u'description': u'ignore me?'}
```

So, a webUI could now get a forum list :)

### DIFF
--- a/libresapi/src/api/ApiServer.cpp
+++ b/libresapi/src/api/ApiServer.cpp
@@ -225,6 +225,7 @@ public:
     ApiServerMainModules(ResourceRouter& router, StateTokenServer* sts, const RsPlugInInterfaces &ifaces):
         mPeersHandler(sts, ifaces.mNotify, ifaces.mPeers, ifaces.mMsgs),
         mIdentityHandler(ifaces.mIdentity),
+        mForumHandler(ifaces.mGxsForums),
         mServiceControlHandler(rsServiceControl), // TODO: don't use global variable here
         mFileSearchHandler(sts, ifaces.mNotify, ifaces.mTurtle, ifaces.mFiles),
         mTransfersHandler(sts, ifaces.mFiles),
@@ -238,6 +239,8 @@ public:
                                    &PeersHandler::handleRequest);
         router.addResourceHandler("identity", dynamic_cast<ResourceRouter*>(&mIdentityHandler),
                                    &IdentityHandler::handleRequest);
+        router.addResourceHandler("forums", dynamic_cast<ResourceRouter*>(&mForumHandler),
+                                   &ForumHandler::handleRequest);
         router.addResourceHandler("servicecontrol", dynamic_cast<ResourceRouter*>(&mServiceControlHandler),
                                    &ServiceControlHandler::handleRequest);
         router.addResourceHandler("filesearch", dynamic_cast<ResourceRouter*>(&mFileSearchHandler),
@@ -250,6 +253,7 @@ public:
 
     PeersHandler mPeersHandler;
     IdentityHandler mIdentityHandler;
+    ForumHandler mForumHandler;
     ServiceControlHandler mServiceControlHandler;
     FileSearchHandler mFileSearchHandler;
     TransfersHandler mTransfersHandler;
--- a/libresapi/src/api/ApiServer.h
+++ b/libresapi/src/api/ApiServer.h
@@ -5,6 +5,7 @@
 #include "ApiTypes.h"
 #include "PeersHandler.h"
 #include "IdentityHandler.h"
+#include "ForumHandler.h"
 #include "ServiceControlHandler.h"
 #include "StateTokenServer.h"
 #include "FileSearchHandler.h"
--- a/libresapi/src/api/ForumHandler.cpp
+++ b/libresapi/src/api/ForumHandler.cpp
@@ -1,0 +1,87 @@
+#include "ForumHandler.h"
+
+#include <retroshare/rsgxsforums.h>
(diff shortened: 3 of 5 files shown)


chozabu:
@sehraf Blast, I see what you mean. I looked at my diff using meld, and it was fine, but I see on github and CLI, the whole file looks different. I'll tidy and force push over this commit

chozabu:
@sehraf fixed - looks like my qtcreator changed line endings to unix style automatically or something

chozabu:
updated  python test code for getting messages:

```
print "\nFORUM"
print "own groups"
fid = 0
fname = ""
r = requests.put("http://0.0.0.0:9090/api/v2/forums")
for f in r.json()['data']:
    if f['own']:
        print f
        fid = f['id']
        fname = f['name']
print fid

print "\n messages in " + fname + "("+fid+")"
r = requests.put("http://0.0.0.0:9090/api/v2/forums/"+fid)
print r.json()['data']
```

and output

```
FORUM
own groups
{u'own': True, u'name': u'another test forum', u'visible_msg_count': u'0', u'pop': u'0', u'pgp_linked': False, u'group_status': u'0', u'parent_grp_id': u'00000000000000000000000000000000', u'author_id': u'6f312e1c10163d6261e02ada6067bbb9', u'id': u'720959e1d78edd39d8682c07fc5a8fe1', u'description': u'ignore me?'}
720959e1d78edd39d8682c07fc5a8fe1

messages in another test forum(720959e1d78edd39d8682c07fc5a8fe1)
[{u'name': u'TestPostTitle', u'thread_id': u'0000000000000000000000000000000000000000', u'id': u'4a10f13e7ba2ab6ddb26b89f7a4f7547a2907bfb', u'parent_id': u'0000000000000000000000000000000000000000', u'orig_msg_id': u'4a10f13e7ba2ab6ddb26b89f7a4f7547a2907bfb', u'author_id': u'6f312e1c10163d6261e02ada6067bbb9', u'group_id': u'720959e1d78edd39d8682c07fc5a8fe1', u'message': u'Test Post Message!\n;)'}]
```

chozabu:
OK guys - very, very unfinished forum bridge http://31.220.117.61:8080/#?forumID=9db8353a0388c8c48982633f9ebc821e 
Javascript is very flawed.
requires this PR

server is here:
https://github.com/chozabu/RsSideSite

electron128:
It looks like you copied from old code. I came to the conclusion that something like while(is_token_ready(token)) sleep(1); is a pain. There is something better in GxsResponseTask, it handles token polling and identity fetching.

It works like this: you subclass GxsResponseTask and implement gxsDoWork(). In the Handler return an instance of your subclass. The ApiServer calls GxsResponseTask::doWork() which checks if all tokens are ready and if all identies are fetched. If this is the case, it calls gxsDoWork(). In gxsDoWork() you tell GxsResponseTask which tokens to poll and which identities to fetch in the next tick. If you are done, you call done() to prevent further ticking.

Read IdentityHandler.cpp for an example.

chozabu:
@electron128 I was looking at identityHandler.cpp as an example!
Looking closer, I guess you mean the method used in handleOwn instead of handleWildcard?

the method like "while(is_token_ready(token)) sleep(1); " seems short and simpler than GxsResponseTask - which also looks very geared towards GXSIDs?

If you require it as an update before merging - I'll give it a go, though I'm happy with the method I understood just by skimming the code!

Really, it'd be nice if GXS had an interface that optionally does not seem async...

electron128:
Yes, IdentityHandler::handleOwn() is a good example for GxsResponseTask. handleWildcard is not.

GxsResponseTask does two things for us:
1. wait for gxs tokens
2. fetch identities
1. comes from gxs and from the fact that the webserver has only one thread. I want to have all waiting code in ApiServer::handleRequest(), it is then easier to later move the waiting to the http server to have parallel request processing.
2. comes from gxs and from the design decision to resolve all identity ids to names. I think it is better for clients if they get names. For example, you can render the JSON output to html with no further requests. This makes the HTML render step very simple.

GxsResponseTask is meant to make life easier, and to reuse more code.
You can do it the way you want. The while(!is_token_ready()) sleep(); method bites you as soon as you want to have identity names, or if you want to have a channel post + comments + identity names.

chozabu:
@electron128 those points sound like fairly clear advantages!
Though, I'm still not totally sure how to use GxsResponseTask - been looking  at it for longer than it would take me to implement libres-posted the "old" way  already....
Tempted to wait and see an example from you in either forums/posted/channels.

On a semi-related note, I could do with grabbing you to see what I am doing wrong with POST requests... retroshare://forum?name=Interacting with libresapi&id=9db8353a0388c8c48982633f9ebc821e&msgid=8e4d885031e7e11c7f462c267d86ce39a317f7ef

chozabu:
@electron128 how would you feel about merging this code for now so we have some forum access from libresapi?
And thanks for pointing out my mistake with post requests! Must be in json format.

csoler:
@electron: i let you merge this one when you think this is ready.

PR #37, opened 2015-08-18
friendlist import export

sehraf:
- export/import friend-list to/from a file 
- peers and groups are saved/loaded
- using QSettings
- unencrypted for manual editing
- (important) new functions have doxygen comments

_you can find the progress (unsquashed) here https://github.com/sehraf/RetroShare/commits/pr-friendlist_import_export_

### DIFF
--- a/retroshare-gui/src/gui/common/FriendList.cpp
+++ b/retroshare-gui/src/gui/common/FriendList.cpp
@@ -28,6 +28,7 @@
 #include <QDateTime>
 #include <QPainter>
 
+#include "rsserver/rsaccounts.h"
 #include "retroshare/rspeers.h"
 
 #include "GroupDefs.h"
@@ -126,6 +127,8 @@ FriendList::FriendList(QWidget *parent) :
     connect(ui->actionHideOfflineFriends, SIGNAL(triggered(bool)), this, SLOT(setHideUnconnected(bool)));
     connect(ui->actionShowState, SIGNAL(triggered(bool)), this, SLOT(setShowState(bool)));
     connect(ui->actionShowGroups, SIGNAL(triggered(bool)), this, SLOT(setShowGroups(bool)));
+    connect(ui->actionExportFriendlist, SIGNAL(triggered()), this, SLOT(exportFriendlistClicked()));
+    connect(ui->actionImportFriendlist, SIGNAL(triggered()), this, SLOT(importFriendlistClicked()));
 
     connect(ui->actionSortByName, SIGNAL(triggered()), this, SLOT(setSortByName()));
     connect(ui->actionSortByState, SIGNAL(triggered()), this, SLOT(setSortByState()));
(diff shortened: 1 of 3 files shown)


csoler:
This is a nice and useful feature. The only problem now IMHO is that the file format is quite hard to get with, especially if you want to add/remove info from it.
I suggest to make it an xml file. See how xml files are parsed in e.g. RsCollectionFile.cpp. It's quite easy to parse them using Qt.

sehraf:
I personally hate XML because it takes up a huge amount of space for storing little information. I agree though that QSettings isn't easy to read for a human. Another option would be JSON (the ExportImportPlugin and libresapi are using JSON).
Anyway when you want XML i'll change it.

csoler:
xml also has one big advantage: it's very easy to ensure backward
compatibility when you update the format with new fields.

sehraf:
same goes for QSettings and JSON ;) - anyway i'll update the code to use XML

csoler:
There's something weird in the base64 keys. They seem to contain "&#xa;" which is suspicious to me. Is that for the character end of line ? if so, just remove them before entering them in the file.

Another suggestion: use 
<root>
   <location=" [location name ] " certificate=" [cert base 64] " sslid=" [sslid ] "  pgpId=" [pgp id] " />
   ...
</root>

I mean, don't call "pubkey" the certificate. It's misleading.

I guess it'll be perfect in the next version -)

sehraf:
Do you have any example for " in the certificate? I cannot reproduce it (or i may misunderstand you). " is not a valid character for base64 so this shouldn't be happening . Or do you mixed up an empty location name?
I can't guarantee the next version to be perfect but i'll do my best :+1:

chozabu:
@sehraf I think cyril means there is a space ( ) not a double quotation mark (")?

csoler:
what I mean is that the base64 that is in the file is wrong because it contains html-encoded line feeds:

pubkey="CQEGAf8AAAJWmQENBFNDpiEBCADHJY3YpbaGt0n3yUxBr6fX6V2l9+lEFrx97mE9&#xa;fBesJRgRgxPJtuzU8DKb+v0tw...

See the "&#xa;' in the middle of the line?

csoler:
Oups! When I copy it, the github website converts it into a line feed. So what I meant by " " is 
" & # x a ; "   without the spaces.

csoler:
there's plenty of them in the base64 encoded lines of the xml file that is generated.

chozabu:
paste with 4 spaces at line start to use "code mode" and avoid markdown parsing

```
pubkey="CQEGAf8AAAJWmQENBFNDpiEBCADHJY3YpbaGt0n3yUxBr6fX6V2l9+lEFrx97mE9 fBesJRgRgxPJtuzU8DKb+v0tw.
+-=/
#&asd
#3
```

like that

PR #39, opened 2015-08-22
Add /usr/lib64 to plugin directories

SarcoRS:
Solution for https://github.com/RetroShare/RetroShare/issues/33

PR #44, opened 2015-08-23
Improve qmake: "make install" support and ability to change install prefix

### LINKED ISSUE #33: Load plugins from /usr/lib64
Reference point in code: https://github.com/RetroShare/RetroShare/blob/master/libretroshare/src/rsserver/rsinit.cc#L1254

What needs to be done:

Check if `/usr/lib` and `/usr/lib64` are symlinked (either `/usr/lib` is a symlink pointing to `/usr/lib64`, or `/usr/lib64` is a symlink pointing to `/usr/lib`). This is done to avoid redundancy in the next step.

If `/usr/lib` and `/usr/lib64` are **not** symlinked, then prepend `/usr/lib64` to the list of directories where RetroShare can read plugins from.

Something like:

```
+ plugins_directories.push_back(std::string("/usr/lib64/retroshare/extensions6/")) ;
  plugins_directories.push_back(std::string("/usr/lib/retroshare/extensions6/")) ;
```

AsamK:
I've tried to improve the qmake build system for RS. The first goal was to make "make install" work, so users can easily install RS system wide from source, without building packages.
The second goal was to make the install prefix (usually /usr) and library path (usually /usr/lib) configurable. The difficulty there was to make the RS code aware of those paths, so RS will look for sounds, stylesheets and plugins in the configured directories and not the default paths. I achieved this by creating DATA_DIR and LIB_DIR defines in the libretroshare.pro that will be picked up in the c++ files.

libretroshare already had an install section, that installed some header files and libretroshare.a, however the header list is completely outdated (it's from 2010). I think in its current state this is not useful so I disabled installation of those files. If someone wants to figure out which header file are necessary for plugin development they can update the list and reenable the installation, might be cool for a retroshare-dev package.

My last commit "Make file names consistent with packaging" is probably a bit more controversial. It changes the names of the output executables to "RetroShare06" and "RetroShare06-nogui", the names which are used in the official debian/ubuntu packages. It also changes the names of the .desktop and .png files to match the names in the package.
With those changes the installation produced by "make install" is the same as installing the debian package. Another advantage is that packagers for other distributions can easily produces packages that are consistent with the debian packages.
I have updated the debian rules file and the windows .nsi installer file to reflect those changes. Both debian package and windows installer produce the same result as before so end-users won't see any changes.
@thunder2 please have a look at the .nsi file, if those changes are ok, i haven't tested them
@csoler please have a look at the changes in debian rules file, i haven't tested them

Please tell me what you think or what I should improve.
### Usage:

```
    # A clean rebuild is necessary when changing the prefix
    make distclean
```
- Default install in /usr
  
  ```
  qmake && make && sudo make install
  ```
- Install in /usr/local
  
  ```
  qmake PREFIX=/usr/local && make && sudo make install
  ```
- Default install in /usr with plugins in /usr/lib64 (this fixes #33 in a nice way and obsoletes #39)
  
  ```
  qmake LIB_DIR=/usr/lib64 && make && sudo make install
  ```
- Optionally for packagers
  
  ```
  qmake PREFIX=/usr LIB_DIR=/usr/lib64
  make
  make INSTALL_ROOT=${PKGDIR} install
  ```


TheSentry:
Would it also be possible to implement a target that doesn't compile the plugins? The current "make" tries to build all and fails if it can't build VoIP for example. Which is really unnecessary on my Raspi.

AsamK:
@TheSentry You could put something like this in RetroShare.pro:

```
NO_PLUGINS {
        SUBDIRS -= plugins/plugins.pro
}
```

and then call "qmake CONFIG=NO_PLUGINS" to build without plugins. Not sure if that's the best way to do it, though.

TheSentry:
@AsamK Thanks, it's been a while since I had to deal with Makefiles. I'll try that after I got RS compile on my Raspi again. At the moment the linker runs out of memory.

chozabu:
not tested it, but looks good :+1:

SarcoRS:
Support for `/usr/lib64` is my wet dream. If this is going to be accepted, it indeed obsoletes https://github.com/RetroShare/RetroShare/issues/33. Being able to move all the stuff that will be packaged with just one `make install` instead of manually copying each file is another wet dream of any packager.

@AsamK, I've just one question. I had a look at the changes to `README.md` made in this pull request and I only see `PREFIX` slightly mentioned there. You wrote great docs in the very first post of this pull request talk. Please include as much as you can from the `Usage` section into the README. Particularly the last two use cases that show customization is what I would very much like to see in the README.

cavebeat:
i like that :+1: 

btw: ping @csoler :bangbang:

sehraf:
:+1:  for merging

csoler:
nice work! Just two things I'm not sure of:
1 - why did you remove the cp -r ${workdir}/data ... from makeSourcePackage.sh / these files will be needed when compiling the package.
2 - when DATA_DIR is not set, you should issue an error, otherwise the compiled code will look at the wrong regions, which is potentially a security whole. So something like this in one of the file where you use it:
#ifndef DATA_DIR
#error DATA_DIR variable not set. Cannot compile.
#endif

AsamK:
1. In the last commit I moved the data directory from /build_scripts/Debian+Ubuntu/ to /, because the redhat build script uses the same data and also other distros need it. That's why I removed the cp, because the data folder is already where it is expected.
2. I'll do that

csoler:
... or use a default value. I suppose you put that in the .pro

PR #48, opened 2015-08-23
Add retro share link open file when exists

PhenomRetroShare:
When the file in the link already exists, RS open it.
Test it in chat lobby. First the file was downloaded and then, if you click a new time on RsLink, the file is opened.


AsamK:
That's useful :+1: 

Just one thing about git usage: could you use "git rebase" instead of "git pull" to base your changes on the latest master branch? That would eliminate the unnecessary merge commits here.

chozabu:
@AsamK I don't mind seeing merge commits in master - maintains just a little more information

PR #58, opened 2015-08-28
Adapt debian/ubuntu package files to use "make install"

AsamK:
This PR removes the manual install commands from debian/rules and replaces them with "make install".
It also removes the unused "builddir" section and adds a new "configure" section that calls qmake.
Finally it updates the *.install files to put installed files in the correct split packages, and adds a call to "dh_install" in "binary_arch" section to actually use the *.install files.


csoler:
I don't see how that can actually work.
The different packages should contain the correct files before the packager is collecting them. That is precisely what the install commands are doing, in each appropriate directory (one per package). if you remove that, the created packages will miss the files right?

AsamK:
No, that's what dh_install now does. It takes the package specific .install files and installs the specified files from the common installation folder "debian/tmp" to the package specific folder.
e.g.: in retroshare06-nogui.install the line "debian/tmp/usr/bin/RetroShare06-nogui" installs the file to "debian/retroshare06-nogui/usr/bin/RetroShare06-nogui"

http://manpages.ubuntu.com/manpages/precise/man1/dh_install.1.html

PR #62, opened 2015-08-29
Qt5 fix

hunbernd:
Some functions were missing on Options-->General dialog under Windows with Qt5, because Q_WS_ macros are not supported. I replaced them with Q_OS_ macros that are supported by both Qt 4 and 5.

I replaced the macros for OSX and Linux too, but I could test it only on Windows.

Also I've found some unnecessary limitations on BandwidthGraphWindow.

### DIFF
--- a/retroshare-gui/src/gui/AboutDialog.cpp
+++ b/retroshare-gui/src/gui/AboutDialog.cpp
@@ -761,13 +761,13 @@ void AboutDialog::on_copy_button_clicked()
 		verInfo+=QSysInfo::prettyProductName();
 	#endif
 #else
-	#ifdef Q_WS_X11
+	#ifdef Q_OS_LINUX
 	verInfo+="Linux";
 	#endif
-	#ifdef Q_WS_WIN
+	#ifdef Q_OS_WIN
 	verInfo+="Windows";
 	#endif
-	#ifdef Q_WS_MACX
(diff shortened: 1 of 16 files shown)


thunder2:
Thank you.
What is the reason to use Q_OS_WIN32 instead of Q_OS_WIN?

hunbernd:
Q_OS_WIN is not available in Qt4 (or just missing from the documentation)
Q_OS_WIN32: the name a little bit misleading, but it is included in both 32 and 64 bit versions

thunder2:
I found the define in Qt 4.8.7 and Qt 5.5.0.

Qt 4.8.7 qglobal.h:
  #if defined(Q_OS_WIN32) || defined(Q_OS_WIN64) || defined(Q_OS_WINCE)
  #  define Q_OS_WIN
  #endif

Qt 5.5.0 qsystemdetection.h:
  #if defined(Q_OS_WIN32) || defined(Q_OS_WIN64) || defined(Q_OS_WINCE) || defined(Q_OS_WINRT)
  #  define Q_OS_WIN
  #endif

hunbernd:
Then it seems like it is just missing from the documentation here: http://doc.qt.io/qt-4.8/qtglobal.html

I edited my commits to use Q_OS_WIN

thunder2:
Thank you. Merged.

PR #66, opened 2015-08-31
More qmake improvements

AsamK:
This PR further improves the qmake build files: The main features:
- Possiblity to install subprojects (gui/nogui) without compiling unneeded subprojects (good to compile nogui on a raspberrypi)
- Add .pri include file for common stuff of all .pro files, so far it contains the default values for the installation paths
- Reduce duplicate code in gui/nogui, by moving LIBS*=… directives to libretroshare (see http://doc.qt.io/qt-5/qmake-advanced-usage.html#libdepend) only done for linux, as i can't test on other platforms
- Treat missing sqlcipher as error, and allow building without it only by using "qmake CONFIG+=NO_SQLCIPHER" (in which case it doesn't link to sqlcipher even if it exists on the system)

For more info see the commit messages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,12 +27,12 @@ addons:
     project:
       name: "RetroShare/RetroShare"
       description: "RetroShare Build submitted via Travis CI"
-    build_command_prepend: "qmake; make clean"
+    build_command_prepend: "qmake CONFIG+=NO_SQLCIPHER; make clean"
     build_command:   "make -j 4"
     branch_pattern: coverity_scan
 
 before_script:
-  - qmake
+  - qmake CONFIG+=NO_SQLCIPHER
 
 #script: make
 script: if [ "${COVERITY_SCAN_BRANCH}" != 1 ]; then make ; fi
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Compilation on Linux
 3. Compile
    ```bash
    cd trunk
-   qmake CONFIG=debug
+   qmake CONFIG+=debug
    make
    ```
 
@@ -55,7 +55,7 @@ For packagers
 -------------
 Packagers can use PREFIX and LIB\_DIR to customize the installation paths:
 ```bash
-qmake PREFIX=/usr LIB_DIR=/usr/lib64
+qmake PREFIX=/usr LIB_DIR=/usr/lib64 "CONFIG-=debug" "CONFIG+=release"
 make
 make INSTALL_ROOT=${PKGDIR} install
 ```
--- a/README.md
+++ b/README.md
@@ -51,6 +51,17 @@ Compilation on Linux
          /usr/bin/RetroShare06
          /usr/bin/RetroShare06-nogui
 
+Compile only retroshare-nogui
+-----------------------------
+If you want to run RetroShare on a server and don’t need the gui and plugins,
+you can run the following commands to only compile/install the nogui version:
+
+```bash
+qmake
+make retroshare-nogui
+sudo make retroshare-nogui-install_subtargets
+```
+
 For packagers
 -------------
 Packagers can use PREFIX and LIB\_DIR to customize the installation paths:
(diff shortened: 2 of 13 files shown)


AsamK:
Fixed travis build by adding CONFIG+=NO_SQLCIPHER to the travis config. The travis build doesn't have sqlcipher and that is now an error unless NO_SQLCIPHER is specified.

thunder2:
Windows compile is broken with the commits d116f40d6ea9a7c992b0deda0f472f6ac6b27eb2 "Move openpgpsdk to common section of libretroshare.pro" and 829c2f396eb19cde5acb1f92a7ebeb310f63d949 "Move LIBS from gui/nogui linux section to libretroshare".
The libs are needed in the pro file of the binary (gui, nogui) not in the libretroshare.pro.

AsamK:
that's what config+=create_prl and link_prl should be doing, automatically adding the LIBS to all projects that use the library (http://doc.qt.io/qt-5/qmake-advanced-usage.html#libdepend) the docs don't mention anything that this doesn't work on windows.
can you try running qmake -r to regenerate all makefiles?

sehraf:
using qmake with '-r' did the trick on Arch Linux (using Qt4)

thunder2:
I already use -r to build rs and it doesn't work with Qt 4 and Qt 5. I tried it with and without shadow build.
qmake ..\RetroShare.pro -r CONFIG+="release"

AsamK:
can you check that the file libretroshare/src/lib/libretroshare.prl is created in the build directory, and that it contains the necessary LIBS?

unfortunately i don't have an RS build environment set up on windows to test it.

thunder2:
The file exists.

QMAKE_PRL_BUILD_DIR = <path to>/libretroshare/src
QMAKE_PRO_INPUT = libretroshare.pro
QMAKE_PRL_TARGET = libretroshare
QMAKE_PRL_CONFIG = include_source_dir lex yacc warn_on debug uic resources incremental_off thread_off windows debug DebugBuild Debug build_pass warn_on link_prl copy_dir_files debug_and_release debug_and_release_target precompile_header debug stl exceptions rtti mmx 3dnow sse sse2 largefile debug DebugBuild Debug build_pass staticlib bitdht create_prl gxs upnp_miniupnpc debug DebugBuild Debug build_pass no_autoqmake static staticlib
QMAKE_PRL_LIBS = ../../libbitdht/src/lib/libbitdht.a ../../openpgpsdk/src/lib/libops.a -lbz2

AsamK:
Strange, the file is created correctly, just not picked up for the gui/nogui builds. I found the same issue here: https://bugreports.qt.io/browse/QTBUG-12901 no real solution, but they say it works with qt 5.3.

if it doesn't work with qt 5.3 either, i guess you should just readd

```
LIBS += ../../libbitdht/src/lib/libbitdht.a ../../openpgpsdk/src/lib/libops.a -lbz2
```

to the gui/nogui .pro files in the win32 section.

thunder2:
The name of the prl file with Qt 4 and Qt 5 is wrong. It's retroshare.prl and not libretroshare.prl (I was wrong in the previous post).

The link with Qt 4 still fails because the order of the libraries is wrong. I don't know why, but ws2_32 needs to be in LIBS after libbitdht, crypto after libretroshare. Maybe we can add these libraries to the prl file too?
In Qt 4 the libs from prl are added at the end and in Qt 5 at the beginning of LIBS, that's why RS can be linked with Qt 5 when I rename the prl file.

thunder2:
In the sources of qmake I found, that "TARGET_PRL = libretroshare" fixes the name.

AsamK:
Good find with TARGET_PRL, I just set up a windows build here, and that fixed the build with qt5.
To add the ws2_32 to the prl file, just move the "LIBS+=-lws2_32" from retroshare-(no)gui.pro to the win32 section of libretroshare.pro. The bitdht lib in this file is added before the win32 section so this should work

thunder2:
Solved link issues by using "QMAKE_LFLAGS += -Wl,--start-group".
Now the order of the libraries doesn't matter.

PR #68, opened 2015-09-01
add option to disable updates to systray tooltip

chozabu:
In KDE5/QT5 - the tooltip is updated using DBUS
This happens once a second - and includes sending the icon, which seems a little heavy for dbus.

This PR just adds an option to disable systray tooltip updates

### DIFF
--- a/retroshare-gui/src/gui/MainWindow.cpp
+++ b/retroshare-gui/src/gui/MainWindow.cpp
@@ -678,6 +678,7 @@ void MainWindow::updateTrayCombine()
 }
 
 void MainWindow::toggleStatusToolTip(bool toggle){
+    if(!toggle)return;
     QString tray = "RetroShare\n";
     tray += "\n" + nameAndLocation;
     trayIcon->setToolTip(tray);
--- a/retroshare-gui/src/gui/MainWindow.cpp
+++ b/retroshare-gui/src/gui/MainWindow.cpp
@@ -677,12 +677,19 @@ void MainWindow::updateTrayCombine()
     updateFriends();
 }
 
+void MainWindow::toggleStatusToolTip(bool toggle){
+    QString tray = "RetroShare\n";
+    tray += "\n" + nameAndLocation;
+    trayIcon->setToolTip(tray);
+}
+
 void MainWindow::updateStatus()
 {
     // This call is essential to remove locks due to QEventLoop re-entrance while asking gpg passwds. Dont' remove it!
     if(RsAutoUpdatePage::eventsLocked())
         return;
-
+    if(Settings->valueFromGroup("StatusBar", "DisableSysTrayToolTip", QVariant(false)).toBool())
+        return;
     float downKb = 0;
     float upKb = 0;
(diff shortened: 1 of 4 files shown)


retroshare-gui/src/gui/MainWindow.cpp
```diff
@@ -677,12 +677,19 @@ void MainWindow::updateTrayCombine()
     updateFriends();
 }
 
+void MainWindow::toggleStatusToolTip(bool toggle){
```
AsamK, line 680: toggle parameter is unused, i think this function should only set the tooltip if toggle == true

retroshare-gui/src/gui/settings/AppearancePage.cpp
```diff
@@ -142,6 +143,7 @@ bool AppearancePage::save(QString &errmsg)
 	Settings->setValueToGroup("StatusBar", "HideSound", QVariant(ui.checkBoxHideSoundStatus->isChecked()));
 	Settings->setValueToGroup("StatusBar", "HideToaster", QVariant(ui.checkBoxHideToasterDisable->isChecked()));
 	Settings->setValueToGroup("StatusBar", "ShowSysTrayOnStatusBar", QVariant(ui.checkBoxShowSystrayOnStatus->isChecked()));
+	Settings->setValueToGroup("StatusBar", "DisableSysTrayToolTip", QVariant(ui.checkBoxDisableSysTrayToolTip->isChecked()));
```
AsamK, line 146: I think pMainWindow->toggleStatusToolTip(ui.checkBoxDisableSysTrayToolTip->isChecked()); should be called here, otherwise the status could have been overwritten by MainWindow::updateStatus() in the meantime

chozabu:
Thanks for the feedback @AsamK ! I'll try those suggestions out now

chozabu:
Done&Done

AsamK:
:+1:

chozabu:
@csoler Any chance of taking a glance at this? without it, RS+KDE5 cause so much dbus activity it seems to cause serious problems...

Every few ~5 seconds, 14,000 lines (mostly re-sending icon) are sent over dbus
http://chozabu.net/autopush/dbuslog.txt <-- this 1mb file is a dbus log over a ~5 seconds with RS open
and this: http://chozabu.net/autopush/dbuslognors.txt about ~15 seconds with no retroshare - the only activity was me switching away from the terminal and back

Sure, in many ways it is technically a QT/KDE bug - but one that may well not be fixed... this is an easy workaround for now and would mean I can use RS without my taskbar grinding to a halt.

csoler:
it should be found why so much data is sent to the tray! That shouldn't happen. If the tray icon is re-sent every 5 secs, that's a bug.

chozabu:
@csoler it's due to updating the tooltip with the current bandwidth.

It would be a tiny amount of data, but there is also an icon in the tooltip,
Even that would not matter so much - but QT5+KDE5 update the systray using DBUS.
DBUS is not intended to take this much data - so it, and KDE5 start freaking out!

PR #72, opened 2015-09-01
Fix Windows compilation of libresapi with error:

PhenomRetroShare:
\libresapi\src\api\ForumHandler.cpp:108: erreur : 'u_int32_t' was not
declared in this scope
                 KeyValueReference<u_int32_t>
vis_msg("visible_msg_count", grp.mMeta.mVisibleMsgCount);
                                   ^


AsamK:
fixed by thunder seven seconds before you ;) c3fff396249abd85d7f0fd0efe3e294dca473675

csoler:
we'd rather use uint32_t, as everywhere else in the code.

PR #73, opened 2015-09-02
Fix subtle bug in tlv deserialization of TLV_TYPE_KEYSIGNATURETYPE

AsamK:
The main sympton of this bug was receiving no more forum/channel posts.

When using -O2, the variable currType is optimized out (it's declared inside the while loop), so the signature is added to the keySignSet with the wrong sign type. Then when verifying the
data, the signature is not found and verification fails.

With debug builds (-O0) this worked, that's why not every one was affected
by this.

This bug was introduced in 2012 (19e856c2a8eb56c5e43e977aac27a3dbc79ab2d0)
It surfaced only now, because before commit 2b8eafa1dbfd252b6391640bd04f2e3e92ec4bc1 from after RC2, libretroshare was always built with CONFIG+=debug, so the error was not observable.


csoler:
Impressive! I actually get a great deal of uninitialized memory reads which I cannot locate in the deserialisation code. Could be related.

In the unpatched code, currType is potentially not intialized though.

PR #86, opened 2015-09-03
Fix Windows 7 32b Compilation from scratch.

PhenomRetroShare:
And add scripts to build Libs on branch.
Good for beginners...


hunbernd:
I recommend to use libssh 0.6.5 instead. It can be built without the fix.

csoler:
libssh is not used anymore. Why would you need it?

hunbernd:
In this case, neither protobuf is needed.

csoler:
protobuf is indeed not needed anymore. I don't think that these two are included in the compilation anyway.

csoler:
@Phenom: thx for all this. But please, try to make separate pull requests in the future when you're committing changes for different subjects. Huge changes in the code like this are very hard to review.

PhenomRetroShare:
But I have to do it in same as they are linked...

csoler:
the video preview looks terrible. What happenned??

csoler:
seems that it comes from the _lastTimeToShowFrame thing. When removed, video looks ok again.
I also found a bug: clicking cancel while video preview is running causes a crash inside Qt. I didn't find the reason. 

Could you plz separate that PR at least in one block for VOIP, fixing the problems above, so that I can merge it? All the rest that deals with windows compilation can only be reviewed by thunder.

csoler:
I merged the VOIP part.

PR #89, opened 2015-09-04
remove VOIP pre-build in travis

chozabu:
Not properly tested - but this should prevent travis from building VOIP plugin.
This means we should not constantly get "build failed" messages - but a more long term solution is to setup a compatible version of libavcoded on buntu 12.04

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ addons:
     branch_pattern: coverity_scan
 
 before_script:
+  - sed -i '/VOIP/d' ./plugins/plugins.pro
   - qmake CONFIG+=NO_SQLCIPHER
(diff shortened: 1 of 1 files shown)


chozabu:
Looks like the build passed - line disabling VOIP compilation ran here: https://travis-ci.org/RetroShare/RetroShare/builds/78728644#L923

AsamK:
:+1:
until someone sets up travis to use a more recent libav version VOIP should be disabled, so travis is useful for other PRs.

AsamK:
This can be closed now. #92 fixed VOIP to compile with older libav versions.

chozabu:
Closing :)

PR #93, opened 2015-09-05
Disable pegmarkdown build by default, only used by disabled wiki

AsamK:
pegmarkdown is only used by the gui parts of the wiki, which is currently disabled, so building it by default is not necessary. Also with the gui disabled, the wiki parts in libretroshare are unused so I disabled those, too. Both can be enabled with CONFIG+=wikipoos.

The wiki gui was bitrotting a bit, using a no longer existing function (rsWiki->updated()), so I ported it to use RsGxsUpdateBroadcastPage. That makes it compile and work as good as it ever has.

### DIFF
--- a/RetroShare.pro
+++ b/RetroShare.pro
@@ -7,7 +7,6 @@ SUBDIRS += \
         libbitdht \
         libretroshare \
         libresapi \
-        pegmarkdown \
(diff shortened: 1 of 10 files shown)


cavebeat:
nice

PR #106, opened 2015-09-13
Create Debian source without re-cloning

heini:
Reworked the makeSourcePackage.sh script, so that it creates the source package directly from the checked out revision, w/o re-cloning the repository into a subdirectory. It now works similar to the package creation script for RedHat/Fedora.

### DIFF
--- a/build_scripts/Debian+Ubuntu/control.jessie
+++ b/build_scripts/Debian+Ubuntu/control.jessie
@@ -2,7 +2,7 @@ Source: retroshare06
 Section: devel
 Priority: standard
 Maintainer: Cyril Soler <csoler@users.sourceforge.net>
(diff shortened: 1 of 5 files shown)


csoler:
I don't understand what you gain with this. It's so easy to include local commits unknowingly doing that, that I consider it a bad solution.

heini:
What's bad about local commits? OTOH, why should I need to clone the repository again, when I already cloned it to get at the build scripts?

csoler:
Because I'm the guy who does the packaging for ubuntu, and I want to make sure that what I'm packaging is the unaltered trunk every time. The clone operation is performed with -depth 1, which is minimal, so I prefer to keep it that way.

heini:
What about adding an option for doing it in-place?

PR #108, opened 2015-09-15
Adding I2P support to RS 

sehraf:
Similar to Tor I2P provides hidden services and socks5 proxies to reach those.
This patch set adds I2P support to RetroShare.

It does:
- enable libretroshare to manage more than one in-proxy (tor + i2p)
- libretroshare will pick the right proxy based on the address to connect to
- generalize the gui - wherever appropriated - and add I2P support to hidden nodes (a hidden node can either be a Tor or an I2P hidden node)

Also:
- existing proxy code and gui functions (to check connectivity) are reused and expanded
- adding more proxies is easier now
- new functions have doxygen comments

### DIFF
--- a/libretroshare/src/pqi/p3linkmgr.cc
+++ b/libretroshare/src/pqi/p3linkmgr.cc
@@ -2026,11 +2026,16 @@ void  p3LinkMgrIMPL::locked_ConnectAttempt_ProxyAddress(peerConnectState *peer, 
 	peerConnectAddress pca;
 	pca.addr = proxy_addr;
 
-	if (mPeerMgr->hiddenDomainToHiddenType(domain_addr) == RS_HIDDEN_TYPE_I2P)
+	switch (mPeerMgr->hiddenDomainToHiddenType(domain_addr)) {
+	case RS_HIDDEN_TYPE_I2P:
 		pca.type = RS_NET_CONN_TCP_HIDDEN_I2P;
-	else
+		break;
+	case RS_HIDDEN_TYPE_TOR:
+	default:
 		/* default tor */
 		pca.type = RS_NET_CONN_TCP_HIDDEN_TOR;
+		break;
+	}
 
 	//for the delay, we add a random time and some more time when the friend list is big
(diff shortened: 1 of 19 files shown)


csoler:
will review that tonight! congrats.

csoler:
It seems to work very well. I have some suggestions for improvement:
- the tooltip in I2P socks proxy should document that I2P does not provide a socks proxy by default. A new I2P tunnel of type "socks 3/4/5" must be created for that. With TOR this comes for free.
- more documentation should be provided in the hidden node network->hidden service tab in order to setup a I2P hidden service. Now it refers to "documentation", which is vague. What basic steps do we need ? It took me a while to figure it out. Having the right keywords in a sentence would help.
- the test button still shows "tor incoming ok" => should be "hidden service incoming ok", at least while the hidden service type is not recognised yet.
- when adding a hidden certificate, the i2P address does not show up. It should be inside the certificate. Actually it does not for TOR either. It would be nice to show it.

csoler:
About the code:
- changing the value of RS_NET_CONN_UDP_DHT_SYNC was dangerous, but apparently it's not used ;-) so it should be fine.
- the default I2P proxy port is set to 9051. I think that the field in the GUI should have a tooltip saying that this is to be set by you. Otherwise it's confusing.

sehraf:
Thx for reviewing! I'll change the things you have pointed out.

sehraf:
Things are fixed now except for the certificate part. Feel free to comment on the explanations for setting up I2P server/client stuff  (since i'm better at coding than at writing tutorials :relaxed:)

sehraf:
certificate is fixed too: it will always show the external addr. for non hidden nodes or the hidden addr. for hidden nodes.

One small problem remains: -> settings -> hidden page -> the plainTextEdit for proxies has a yellow background color - how do i remove this? Or can it stay like this?

mhh the latest patch (7c7e4f86dc5d41e7557db1e0a1515d03b359f6fb) seems to create conflict in retroshare-gui/src/gui/connect/ConnectFriendWizard.ui - since it's rather small you can easily resolve it :wink:

PR #116, opened 2015-09-25
Format ISO Received time for chat message in WebUI.

PhenomRetroShare:


AsamK:
:+1:

electron128:
I think it would be better to format the time on the client side, because we could then respect the client time zone.
It would be nice to show relative time, like "5 minutes ago".
Does someone know a JavaScript library for this?
(Would be nice if the library could speak multiple languages.)

csoler:
@electron128: I leave this one on you. It's apparently not 100% ready.

PhenomRetroShare:
Yes it is finished.
Use web client date format and time offset.

electron128:
Commit 1 would be ok, but i found it is much easier to do this in JavaScript:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date
Commit 2 is not ok, because it deletes some files which i want to keep. Those files appear unused, because they are not meant to run in the browser. They are made for nodejs.
I added formatting of the time in a75cbcd7b3b82933be38276acd7491437d3c6303 with a one-liner:

``` javascript
new Date(msg.send_time*1000).toLocaleTimeString()
```

I want to avoid pulling in more React dependencies, because i may drop React.
Thanks for your work on this Phenom, i hope you can understand that i prefer the solution which uses standard JavaScript functions.

PR #117, opened 2015-09-27
Remove some garbage from pqissl

G10h4ck:
That code was actually doing nothing a part from wasting CPU cicles instantiating a mutex just to return false...


csoler:
The two log output are actually useful I think. For the rest of it, I totally agree.

G10h4ck:
readded the outputs

csoler:
Will take care of this asap. Sorry for the delay.

PR #119, opened 2015-10-01
Add Broadcast in list of lobbies.

PhenomRetroShare:
Before you can only write text on it if unread messages are in.


csoler:
Please do not "(void)" every unused parameter! It create nop code that makes the source harder to read => bad. Instead, put the params in comments   bool function(uint32_t  /\* param_name */)

PR #173, opened 2015-11-05
show or hide ID chooser in message box depending on destination of msg

mr-alice:
This should solve the ambiguous situation caused when an identity is required to send a message that is not intended to another identity. Hope it helps.


mr-alice:
sorry for the mess. I had to do this multiple times before getting it right.

PR #176, opened 2015-11-08
Webui download enhancements

hunbernd:
- FileStreamer returns correct content-type header based on file extension
- MediaPlayer on webui supports video files too
- Some minor enhancements

The content types are loaded from /etc/mime.types file, on Windows this file should be copied into the same folder as RetroShare06.exe

### DIFF
--- a/libresapi/src/api/ApiServerMHD.cpp
+++ b/libresapi/src/api/ApiServerMHD.cpp
@@ -7,6 +7,7 @@
 #include <algorithm>
 
 #include <util/rsdir.h>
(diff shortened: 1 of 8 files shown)


electron128:
thanks for this pull request.

There was only one issue with this line:
    unsigned int i = line.find_first_of("\t ");
It does not work on my system, because sizeof(unsigned int)=4 and sizeof(size_t)=8.
std::string::npos of type size_t does not fit into unsigned int, and we have an infinite loop.
I will fix this on master.

PR #179, opened 2015-11-20
latest OSX patches

sirjenster:
This should  make the final changes to enable one to build the gui and nogui by simply adding dependencies. dependencies presently located in /usr/local but can be changed later.

Still need to work on build instructions, build script, and add(but not needed) application icon and other resources..

In the distant future WILL need serious HELP with  plugin support and MAC_IDLE fix. perhaps native zero-configuration.


csoler:
I don't understand why you need the private/public thing in p3grouter.h
Same for the parameter values in help() of argstream.

sirjenster:
missing help() change

sirjenster:
try try again

sirjenster:
trying again to updste retroshare.cc

PR #189, opened 2015-12-07
Chat enhancements

hunbernd:
> green
> /me
> Saving images posted in chat or forums
> Spoiler text

### DIFF
--- a/retroshare-gui/src/gui/chat/ChatStyle.cpp
+++ b/retroshare-gui/src/gui/chat/ChatStyle.cpp
@@ -272,6 +272,7 @@ QString ChatStyle::formatMessage(enumFormatMessage type, const QString &name, co
 	QString styleOptimized ;
 	QString errorMsg ; int errorLine ; int errorColumn ;
 	QString messageBody = message ;
+	me = me || message.trimmed().startsWith("/me ");
 	if (doc.setContent(messageBody, &errorMsg, &errorLine, &errorColumn)) {
 		QDomElement body = doc.documentElement();
 		if (!body.isNull()){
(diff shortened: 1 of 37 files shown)


csoler:
Hi! Thx for this. I have a few suggestions (from me and other persons in the dev lobby):
- the "save image" feature checks that what's pointed is actually an image since it warns if it's not. Could you show the "save image" menu entry only when it is an image instead?
- the hard-coded color for the highlighted text is a problem. Is there a way to ask Qt to highlight the text in a generic way instead? Apparently there's something in RS to do that already in HandleRichText.cpp.
- the spoiler text seems not to work. The menu indeed creates a "_spoiler_ link, but then when the forum messages or chat messages containing it are displayed, they just show it as non clickable text.
- when using "/me" in chat, the text for my own name in the chat window becomes "*" instead of my name.

hunbernd:
- Save image: OK
- It is OK if it is coded into the stylesheets of different color schemes? Or should I use a color picker button somewhere in settings. It effects chats, forums and messages too, so where should I put the button?
- Put you cursor on the _SPOILER_, you will see it in the tooltip. The link thing is needed only, because QTextEdit.toHtml does not export the _title_ attribute, so I have to put the hidden information into somewhere else, then put it back into the tooltip. I don't want to keep it clickable, because it is not backward compatible, tooltip is.
- /me: it is intended to work like this. An another option is to hide the name field completely, but it is not possible, without creating a copy of all html files in qss/gui/chat with the name fileld removed.

TheSentry:
Highlighting text by stylesheets sounds like a good idea. That way highlighting can be made to work with different color schemes, which is a mess right now.

csoler:
What I meant is that /me causes the title of your own posts to be "*" instead of your name in private chat. That's definitely a bug.

ghost:
What does ">greentext" mean? What does that commit do?
Setting a color in the stylesheet will probably not work for the "skin with no name" because with that skin RS is using the default system colors that can be anything. But I do not know which color we are talking about.

hunbernd:
> What does ">greentext" mean? What does that commit do?

It is used in forums to mark quoted text. This commit changes the colour of the lines that start with > to green. Example: https://en.wikipedia.org/wiki/Template:Greentext

> Setting a color in the stylesheet will probably not work for the "skin with no name"

RS uses a built-in stylesheet, which is loaded always, even when no other colour scheme selected

hunbernd:
- Now greentext's color is defined in qss.default and can be overridden by color schemes.
- Also I fixed the thing with save image menu.
- SPOILER was not broken. It is possible that Linux users have to wait more for the tooltip to show up.

I still have to figure out what to do with /me.
Anyway if the other features OK now, I can separate the /me feature and create a new pull request with the rest.

defnax:
yes the features are ok for me, they dont want the /me?, better /me seperated.
seams skype has this /me too:
https://support.skype.com/en/faq/FA10042/what-are-chat-commands-and-roles

hunbernd:
Fixed issues with /me

PR #207, opened 2015-12-16
V0.6 gxs debug

csoler:
In depth review of GXS net service code. In summary:
- time stamp IDs of subscribed groups so that they are not forgotten
- finxed timing issues
- documentation of gxsnetservice
- improved debugging information to allow to print data onf each service/peer/group combination


csoler:
@other devs: do not merge this yet. I have created it for testing. It needs some more work: bandwidth limitation and one bug to fix in faulty signed group propagation.

PR #212, opened 2015-12-20
V0.6 traffic optim

csoler:


csoler:
Groups packets in pqistreamer just before sending to friends. This apparently improved the crypto overhead due to paddign of small data packets.

PR #219, opened 2015-12-23
Fix pqihandler::UpdateRates() to get more accurate.

PhenomRetroShare:
From Jolavilette


csoler:
I'll take care of this asap. Just give me the time to fully understand what it's doing ;-)

PR #224, opened 2015-12-28
Fix GenCertDialog masking PGP Key password dialog at end.

PhenomRetroShare:
Sometimes, at end, when creating new key, the dialog stay behind.


csoler:
Is there a way to put the passwd box in front instead of hiding the rest? That seems like a dangerous way to fix the problem. One option is to make the passwd box modal (it should, for security) and the gen cert dialog non modal (it doesn't need to be). Then the passwd will always be on top of the gen cert dialog.

PhenomRetroShare:
But the GenCertDialog close directly after in all case. No need to see it...

PR #237, opened 2016-01-03
p3ServiceControl: Optimize checkFilter(...), solve memory leak removi…

G10h4ck:
p3ServiceControl:

Optimize checkFilter(...) there was a map lookup which result was used only for debugging, moved the map loockup inside debug guard

solve memory leak removing mSerialiser which was allocated but never used or deleted

code cleanup solved a bunch of compiler warning plus very long lines plus indentation

### DIFF
--- a/libretroshare/src/pqi/p3servicecontrol.cc
+++ b/libretroshare/src/pqi/p3servicecontrol.cc
@@ -3,7 +3,7 @@
  *
  * 3P/PQI network interface for RetroShare.
  *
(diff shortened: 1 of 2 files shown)


csoler:
I cannot see the changes in the .cc file on github.

csoler:
Besides, big cosmetic changes like this:
- are a pain to review
- drop ownership of lines, which means it's hard to find the right person to ask for a particular piece of information
- should not be an excuse to put your name in the beginning of the file (wtf?)

PR #242, opened 2016-01-07
Code cleanup

G10h4ck:
As I read code I need to understand in order to finish IPv6 implementation and  multiple connection I clean it up according to recommended coding style so me and others reading the code can understand it easier, this will also make easier the future REAL changes needed for IPv6

https://github.com/RetroShare/RetroShare/wiki/Coding

### DIFF
--- a/libretroshare/src/chat/p3chatservice.cc
+++ b/libretroshare/src/chat/p3chatservice.cc
@@ -1249,8 +1249,8 @@ void p3ChatService::saveDone()
 
 RsSerialiser *p3ChatService::setupSerialiser()
 {
-	RsSerialiser *rss = new RsSerialiser ;
-	rss->addSerialType(new RsChatSerialiser) ;
+	RsSerialiser *rss = new RsSerialiser;
+	rss->addSerialType(new RsChatSerialiser);
 	rss->addSerialType(new RsGeneralConfigSerialiser());
 
 	return rss ;
--- a/libretroshare/src/pqi/p3linkmgr.cc
+++ b/libretroshare/src/pqi/p3linkmgr.cc
@@ -73,21 +73,23 @@ const int p3connectzone = 3431;
 
 //#define P3CONNMGR_NO_TCP_CONNECTIONS 1
 
-const uint32_t P3CONNMGR_TCP_DEFAULT_DELAY = 3; /* 2 Seconds? is it be enough! */
-const uint32_t P3CONNMGR_UDP_DEFAULT_DELAY = 3; /* 2 Seconds? is it be enough! */
+const uint32_t P3CONNMGR_TCP_DEFAULT_DELAY = 3;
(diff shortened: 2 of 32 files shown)


sehraf:
on which basis do you indent functions? see
https://github.com/G10h4ck/RetroShare/commit/a334ca27f5e9345601be4aa44a9a8fafdd799164#diff-fb6ee67d1dbf6322e01fc01097306619R174
https://github.com/G10h4ck/RetroShare/commit/a334ca27f5e9345601be4aa44a9a8fafdd799164#diff-fb6ee67d1dbf6322e01fc01097306619R181
https://github.com/G10h4ck/RetroShare/commit/a334ca27f5e9345601be4aa44a9a8fafdd799164#diff-fb6ee67d1dbf6322e01fc01097306619R186

G10h4ck:
Thanks you for spotting those are actually alignment errors, my editor QtCreator changed spaces to tabulators on save so it looked good on my editor but wrong on GH I am fixing it right now to prevent this i have found this useful

http://stackoverflow.com/questions/14387920/qt-creator-indentation-changes-at-save

Those parameters should have one indentation so 1 tabulator and the rest alignment spaces

G10h4ck:
@sehraf check it now ;)

sehraf:
looks good :+1: 

though i dislike the different indention here (though i understand why it is like it is) 
maybe this can be unified?

https://github.com/RetroShare/RetroShare/commit/f8ba08a49f3dc13cfd76b64bed6144137b79f2c6#diff-fb6ee67d1dbf6322e01fc01097306619R310
https://github.com/RetroShare/RetroShare/commit/f8ba08a49f3dc13cfd76b64bed6144137b79f2c6#diff-fb6ee67d1dbf6322e01fc01097306619R313

G10h4ck:
That's because if I put the first parameter on the first line the alignment of the second would make it overflow the 80 column, while I understand your point I am doing this only if necessary because it feels an overkill to use a newline everywhere also if not necessary.

csoler:
I don't see myself the point in re-indenting everything in a big commit. It drops author ownership of lines, and makes it easy to insert a nasty bug in a huge commit. These two reasons should be enough to not do it, except in places where you actually add something. 

Among your changes, I like the replacement of mutex calls by the define, which allows to measure time in debug mode. Quite useful.

sehraf:
indention broken:
https://github.com/RetroShare/RetroShare/pull/242/files#diff-9eae8aa965a70a84681659dcd9ce205cR119
https://github.com/RetroShare/RetroShare/pull/242/files#diff-9eae8aa965a70a84681659dcd9ce205cR223
https://github.com/RetroShare/RetroShare/pull/242/files#diff-9eae8aa965a70a84681659dcd9ce205cR304
https://github.com/RetroShare/RetroShare/pull/242/files#diff-9eae8aa965a70a84681659dcd9ce205cR321

PR #247, opened 2016-01-10
Switching Debian Stretch builds to QT5.

heini:
Did a succesfull build against QT5 on Stretch today with these changes.

### DIFF
--- a/build_scripts/Debian+Ubuntu/control.stretch
+++ b/build_scripts/Debian+Ubuntu/control.stretch
@@ -1,0 +1,44 @@
+Source: retroshare06
+Section: devel
+Priority: standard
+Maintainer: Cyril Soler <csoler@users.sourceforge.net>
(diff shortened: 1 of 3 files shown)


sehraf:
since Qt4 support ended last year (19.12.2015) i'd suggest using Qt5 everywhere.

heini:
I'd guess that depends on availability of QT5 for particular (older) distribution versions.

electron128:
ping @csoler : please have a look at this. If you want to merge this, you maybe want to merge before next release.

sehraf:
maybe we can switch all OS's that support it  to Qt5 for the next RC?

csoler:
I'm ok with the switch to Qt5, but I insist on being able to keep Qt4 when Qt5 not available. Some people use RS on home servers or NAS that might not have Qt5.

csoler:
why do you need a new rules file btw?

heini:
Because this is for Stretch only. Don't know whether it's possible to have conditionals in the rules file to distinguish between releases.

PR #248, opened 2016-01-10
sqlcipher needs to be compiled with -fPIC on RHEL/Fedora.

heini:
Without this flag, I get linker errors on some RHEL family distributions.


electron128:
thx heini!

PR #254, opened 2016-01-16
Automatic [De]Serialization

G10h4ck:
Implemented automatic [de]serialization based on RsAutoSerializable and RsAutoTlvSerializable, this is a radical change for RetroShare but retro-compatibility is kept and things has been made so the porting to the new [de]serialization is easy and can be progressive


csoler:
This is nice! Could you please split this in two? One part for de-serialisation and the changes related to it, the rest for the code cleaning? Thx

G10h4ck:
I would have preferred to get all those clean-up as bonus but we already discussed about that so here it goes the PR without the bonus clean-up

PR #260, opened 2016-01-27
Add buttons to send message in plain text and without EmoteIcon.

PhenomRetroShare:


csoler:
This can be useful to many users. Could you plz, before I merge:
- resolve conflicts
- make two buttons better looking? (use an appropriate icon instead of text)
  Also, the NE button seems to have no effect. When I use it, I can still see the emoticons in the windows where the cat is sent.

csoler:
actually, as I understand it, the emoticons is just the way some text is displayed on the receiver side. So you cannot really "disable" them when sending messages right?

PhenomRetroShare:
All are here:
https://github.com/RetroShare/RetroShare/pull/260/files#diff-09d28d1bfe72ae398a40fa0a95b848e7R877
If receiver is not updated, yes it's not working.

For the icon, defnax may change this after.

PhenomRetroShare:
Done

PhenomRetroShare:
I've added Icons.

PR #261, opened 2016-01-27
Improved message shown over lobby names

PhenomRetroShare:


csoler:
this is too much of a hack: the day we change the tooltip, the change you're proposing is not working anymore. Could you find a way to do the tooltip thing when inserting the message instead? It's also bad to do the tooltip in pone place and replace it in another place. thx!

PR #269, opened 2016-02-07
Save last state of OpMode status bar droplist and restore it at start.

PhenomRetroShare:
Add colored style sheet to OpModeStatus ComboBox.
And defined default color when stylesheet files are not found.

Still get Bad effect for Combo with GTK++ or WindowsVista and Standard, but get it too without this commits, only adding these lines to \retroshare-gui\src\gui\qss\stylesheet\qss.default:
OpModeStatus {
    background: #CCFFCC;
}

### DIFF
--- a/retroshare-gui/src/gui/qss/stylesheet/qss.default
+++ b/retroshare-gui/src/gui/qss/stylesheet/qss.default
@@ -217,3 +217,23 @@ GenCertDialog QLabel#entropy_label
 {
 	qproperty-fontSizeFactor: 115;
 }
+
+/* OpModeStatus need to be at end to overload other values*/
+OpModeStatus {
+	qproperty-opMode_Full_Color: #CCFFCC;
+	qproperty-opMode_NoTurtle_Color: #CCCCFF;
+	qproperty-opMode_Gaming_Color: #FFFFCC;
+	qproperty-opMode_Minimal_Color: #FFCCCC;
+}
+OpModeStatus[opMode="Full"] {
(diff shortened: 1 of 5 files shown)


felisucoibi:
a very very usefull thing, i have to reset this every lanuch, thanks!

csoler:
Some work is needed in order to make this right:
- the dropbox looks weird. It does not have the same style/look than other widgets
- normal mode should not have a specific backgroud color, but only use the default
- fix the tooltip. There's a few TODO left in it.
- add one tooltip for each entry in the dropbox, or if not possible, change the tooltip to explain  the side effects of the current choice. Only "normal mode" should show all the information (like it does now).

I'm still convinced that this will raise far more bug reports than new happy users, since it is very easy to restart and forget that setting. The normal behavior of re-starting RS should be to leave it in a fully working state.

felisucoibi:
Every time we start the program we have to change the settings to "not anonymous", also because even if it's a direct friend tries to share trough anonymous tunnels so it occupy slots... if its too complicate and you think will generate problems, by default hide this options menu, so by default new installation is "normal mode" and can't a new user to change it, and enable it on advanced mode, and then show the different options, and save the last state so only experienced users will play with this and set for next session what they consider.

felisucoibi:
Hide and remember by default
https://github.com/RetroShare/RetroShare/issues/486
I thin i talked once about this with csoler and agreed this way.

csoler:
merged using cherry-picking.

PR #279, opened 2016-02-14
added openSSL option SSL_OP_SINGLE_DH_USE regarding CVE-2016-0701

cavebeat:
tl;dr The OpenSSL 1.0.2 releases suffer from a Key Recovery Attack on DH small subgroups. This issue got assigned CVE-2016-0701 with a severity of High and OpenSSL 1.0.2 users should upgrade to 1.0.2f. If an application is using DH configured with parameters based on primes that are not "safe" or not Lim-Lee (as the one in RFC 5114) and either Static DH ciphersuites are used or DHE ciphersuites with the default OpenSSL configuration (in particular SSL_OP_SINGLE_DH_USE is not set) then is vulnerable to this attack.

### DIFF
--- a/libretroshare/src/pqi/authssl.cc
+++ b/libretroshare/src/pqi/authssl.cc
@@ -334,6 +334,13 @@ static  int initLib = 0;
 	sslctx = SSL_CTX_new(SSLv23_method());
 	SSL_CTX_set_options(sslctx,SSL_OP_NO_SSLv3) ;
 
+	//SSL_OP_SINGLE_DH_USE 	CVE-2016-0701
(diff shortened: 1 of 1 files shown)


cavebeat:
http://intothesymmetry.blogspot.de/2016/01/openssl-key-recovery-attack-on-dh-small.html
https://www.openssl.org/news/secadv/20160128.txt
https://bugzilla.redhat.com/show_bug.cgi?id=1301845
http://arstechnica.com/security/2016/01/high-severity-bug-in-openssl-allows-attackers-to-decrypt-https-traffic/
https://tools.cisco.com/security/center/viewAlert.x?alertId=43320

PR #281, opened 2016-02-16
discovery: don't remove new added locations too soon

sehraf:
The current code adds new locations via discovery with a "last seen" time stamp set to 0. This leads to the problem that the new location gets cleaned up within the next 10 minutes when no connection is established. This can be to fast for DHT connections when both sides are behind NAT/firewall.

This patch increases the "protection" time to 1 day by adding new locations with a sound "last seen" value.

### DIFF
--- a/libretroshare/src/pqi/p3peermgr.cc
+++ b/libretroshare/src/pqi/p3peermgr.cc
@@ -2749,7 +2749,7 @@ bool p3PeerMgrIMPL::removeUnusedLocations()
 		std::map<RsPeerId, peerState>::iterator it;
 		for(it = mFriendList.begin(); it != mFriendList.end(); ++it)
 		{
-			if (now - it->second.lastcontact > RS_PEER_OLD_PEER)
+			if (now > it->second.lastcontact + RS_PEER_OFFLINE_DELETE)
 			{
 				toRemove.push_back(it->first);
 
--- a/libretroshare/src/pqi/p3peermgr.cc
+++ b/libretroshare/src/pqi/p3peermgr.cc
@@ -64,8 +64,6 @@ const uint32_t MIN_TIME_BETWEEN_NET_RESET = 		5;
 
 const uint32_t PEER_IP_CONNECT_STATE_MAX_LIST_SIZE =     	4;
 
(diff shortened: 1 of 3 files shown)


csoler:
I think that you could set it to time(NULL) directly. 

Also, a while ago, the removal of old locations always made sure that at least the most recent location was never deleted. Otherwise, the PGP key with no locations becomes useless anyway. That would be nice to re-instate this.

sehraf:
i'm not sure if it's a good idea to add them with time(NULL) since this will add old locations with a new time stamp. but if you want it you shall get it :smile: 

the second patch adds a check that at least one location is kept (untested)

csoler:
(minor), please do if(now > it->second.lastcontact + RS_PEER_OLD_PEER) instead of if(now - it->second.lastcontact > RS_PEER_OLD_PEER). I have had bad moments with differences between uint32_t / size_t. 

(more seriously) you always keep one location among the ones to be deleted. That's not what I meant. I meant kept at least one location among all locations. Also, you're doing a O(n^2) loop since you search each time you need to remove something. Just make a linear pass first to find the most recent location, then check that it's not the one you need to remove.

sehraf:
Ok the minor remark is addressed now (FYI: all three values are time_t)

You are right about the clean up code - i guess i'll have to rewrite it instead (I've removed the patch for now)

Any comments about the time stamp to use when adding a location via discovery? After thinking a bit more about the problem adding a second time stamp for "last time connected" might be the best solution after all (or bool). Otherwise you may never get rid of old locations since there might be a friend that shares it just before he will delete it and then you have it and will share it for another day (or whatever time window we will use)

csoler:
If we use a single TS for cleaning and deciding if we send in discovery, we need:
- set the delay for delete to be D1 (e.g. 90 days)
- only send in discovery if now < TS+D2   (with D2 equal to something like 30 days)
- set the TS when receiving via discovery to be "now - D2 - 1"
  Then disc-ed locations should be kept 60 days before being whipped  out, but won't be sent by discovery until connected.
  So I suppose that you were right in the first place. We just need T2 to be

sehraf:
using this approach sounds reasonable - i'll send in a patch as soon i find some spare time

PR #283, opened 2016-02-17
Fix connect in ChatMsgItem for avatar updateAvatar.

PhenomRetroShare:


thunder2:
You don't need to connect the changed signals because the AvatarWidget should handle the change of the avatar itself.

PhenomRetroShare:
Removed them as connected in Constructor.

thunder2:
Thank you

PR #285, opened 2016-02-17
Fix error "(EE) SSGxsChannelGroup::load() asked to load a null string.

PhenomRetroShare:
Weird." by creating a default serviceString if empty.

### DIFF
--- a/libretroshare/src/services/p3gxschannels.cc
+++ b/libretroshare/src/services/p3gxschannels.cc
@@ -535,26 +535,6 @@ bool p3GxsChannels::getChannelDownloadDirectory(const RsGxsGroupId & groupId,std
 
     /* extract from ServiceString */
     SSGxsChannelGroup ss;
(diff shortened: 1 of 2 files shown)


electron128:
I would just define "empty string" as a valid case and not being an error. Initialise SSGxsChannelGroup with default values at only one place, which is the constructor.

Can you make a new patch with the following things?
- fix debug output in p3GxsChannels::getChannelDownloadDirectory (same as current patch)
- initialise all members in SSGxsChannelGroup constructor to default values (same as current patch)
- in SSGxsChannelGroup::load(): first thing should be to check if string is empty. If yes, don't touch any members and just return true. We might want to print something if debug is enabled.

I'm surprised that there is no mutex in the whole p3GxsChannels class. We should fix this, but in a different patch. An we have to take care then to not deadlock ourself.

PhenomRetroShare:
Done.

electron128:
thank you Phenom!

PR #286, opened 2016-02-18
sqlcipher lib  was not visible for compiler

xcps:


csoler:
sqlcipher does normally not need to be installed. Retroshare will pull the libs from that directory and link to them statically.

Issue #250, opened 2016-01-10
In download queue window add a "Load .rscollection file"


felisucoibi:
In download queue window add a "Load .rscollection file", that should be exactly the same option that appears in your filelist to open .rscollection but in the right natural place, download queue, maybe a new tab there called "Download .rscollection" next to "Search" or just a button somewhere visible.

That is usefull, and also by now solves another bug we send that when you download a .rscollection it doesn't allow you to open it, just open the folder, that instead should open the donwload .rscollection menu with the directory strcuture....
Bests from the Spanish rs user database.

defnax:
done 

![Image](https://github.com/user-attachments/assets/29751a66-28a0-4915-b567-61f6d059eb69)

@csoler close

PR #297, opened 2016-03-01
Allow only one instance. If exists, pass to it arguments.

PhenomRetroShare:
For now, only allow to pass RsCollection file.

To test it, "open with" in your explorer and select your RS executable.


hunbernd:
Please do not limit RS to only one instance. Running multiple instances is useful for developing and testing.

One possible solution is to keep the RS instance running if no arguments specified.

Or do things based on  what kind of arguments specified. If the first argument is an existing file or RS link then pass it and close the current instance. Otherwise continue with normal operation, keep RS running and let it process the arguments, do not pass anything to the other instance.

sehraf:
i agree with hunbernd - only exception might be when clicking on a RS link

PhenomRetroShare:
Change it to only take care if autologged.
Because, when autologin is enable, you can't run two instances (lock file).
When autologin is disable, you can't know to whom send args.

hunbernd:
> you can't know to whom send args

What about making a checkbox in setting which turns on and off the QLocalServer. It shall be enabled by default. It is suitable for most of the users, but one can set which instance will receive the args if needed.

If it is enabled on multiple instances:
According to the documentation, it is not problem on Windows, multiple servers are allowed, but only one will receive the message. 
On Linux disable the server if one is already listening.

felisucoibi:
Yes, some how the .rscollection should be opened, and would be verny nice to right click, and open with rs, maybe like they say only not open a new instance if there is a parameter with the .rscollection file, also I sugested a button in downloads to open a .rscollection
https://github.com/RetroShare/RetroShare/issues/250

csoler:
It's really nice to have this working on linux! This PR has multiple problems however, which can all be fixed easily:
1 - do not ask the user to modify the rights on /usr/share/applications/retroshare06.desktop. The file should already contain the proper lines according to the installed files. If not, display a tooltip that says RS hasn't been installed using a package manager, so the file is missing.
2 - update-desktop-database should be called after package installation (needs to be added to postinst script for debian users). Dont' ask the user to run a "sudo something".
3 - what does the tooltip on the 2nd checkbox means?? I think it has an error, and I dont get the idea.

In the end, on linux at least, I don't see why this option shouldn't be available from installation instead of being allowed to be switched on/off. The second checkbox is useful of course, since it allows to set a preferred instance for receiving the orders. So I would say:

4 - on linux, remove the 1st checkbox. It's not needed. Just make it happen using properly installed installed files. If so, a label should replace the 1st checkbox to say wether it is possible or not to use it (in case the RS instance running cannot find the files)

Note: the second checkbox can be used to switch the service off already.

thx!

PhenomRetroShare:
Done.

csoler:
it seems to me that RS still complains that it cannot write to /usr/share/applications/... It should never have to write there. Did I miss something?

PhenomRetroShare:
You have to remove then pull my branch as I've amanded the commit.
I know it's bad, but I don't want to make a lot of commits for one PR.

defnax:
There is always a problem to take RS links from Browser on windows,
Add friend via rs link from browser, rs starts a new instance.
I think its same for download links?

RS links from Browser works only, if there is no RS instance running on Windows.
This is not fixed yet?

Normally links must be taked when RS is running, and not start new instance.

PhenomRetroShare:
With this branch and "Use Local Server" option selected:
If you click on RsLink when no instance is running, a new one is opened.
If you click on RsLink when an instance is running, this one receives the link and no more is created.
For "Register retroshare protocol", this (only in windows) register retroshare:// as URL protocol to open with RS. So if unchecked, nothing appends when you click on link. This option is activated by default in Linux and MacOS.

csoler:
sorry to be a pain but.... the branch pr/297 has conflicts (lots of them). I don't know what you've done but it's a bit messy. Could you e.g. make a new clean PR ? thx.

csoler:
I finally merged it. thx!

PR #313, opened 2016-03-19
Move decrypted_data initialization on p3gxstunnel.cc

PhenomRetroShare:
So It will be initialized only just before needed.
Thanks to DTS to report this cppcheck error.


csoler:
actually there's a better way to handle this, using RsTemporaryMemory. See commit 2e9d22c

PhenomRetroShare:
Doesn't know it, good.

PR #323, opened 2016-03-24
Webui

zeners:
new webui - developed in cooperation with electron


zeners:
added: starting download by pasting RS-links

PR #326, opened 2016-03-28
Fix error in compilation with last Cpp11 g++

PhenomRetroShare:


csoler:
We really cannot go down that path. Having these ifdef c++11 whenever we have a const in the code is simply not possible. What's is the problem exactly? Is the compiler reporting any guidelines on how to fix this?

PhenomRetroShare:
This is the error:
/libretroshare/src/pqi/pqiqosstreamer.h:37: erreur : 'constexpr' needed for in-class initialization of static data member 'const float pqiQoSstreamer::PQI_QOS_STREAMER_ALPHA' of non-integral type [-fpermissive]
   static const float    PQI_QOS_STREAMER_ALPHA      = 2.0 ;
                                                       ^

You can have a look to this: http://stackoverflow.com/questions/9141950/initializing-const-member-within-class-declaration-in-c

dtschmitz:
Might this be for the new branch of circle changes?? Is circles using qt5?  If so is phenom building with qmake-qt4 or qmake-qt5?  Am I barking up the wrong tree?...

...because, either an environment var should be set or /usr/bin/qmake should symlink to qtchooser which will look to ~/.config/qtchooser for a symlink from the appropriate conf (4.conf or 5.conf) in /etc/xdg/qtchooser:

`[dts@ThinkPad qtchooser]$ ls -l
total 0
lrwxrwxrwx 1 root root 25 Mar 28 13:14 default.conf -> /etc/xdg/qtchooser/5.conf
[dts@ThinkPad qtchooser]$
`

PhenomRetroShare:
I use last arch Qt5

dtschmitz:
Phenom, are the functions prototypes the same from cyril's circle merge?

PhenomRetroShare:
constexpr come from C++11

aereaux:
Would [this](https://doc.qt.io/qt-5/qtglobal.html#Q_DECL_CONSTEXPR) help resolve this issue?

csoler:
then this is simple to solve:
- in the .h, remove the "= something" part. 
- in the .cc, add the declaration of the value:
  const float pqiqosstreamer::PQI_QOS_STREAMER_ALPHA = 2.0;

PhenomRetroShare:
Can't get it working with RsReputations and RsServer as only header file...

zeners:
Phenom: what do you tried? csoler suggestion in #332

FanchTheSystem:
why not replace const by constexpr in header file ?

zeners:
@FanchTheSystem backward compatibility i guess, i only followed csolers suggestion

PR #337, opened 2016-04-03
Fixed compilation for clang.

merkeld:
Fixed compilation for clang. Compilation with gcc still works. Binarys with both compilers were tested.

### DIFF
--- a/retroshare-gui/src/gui/gxs/GxsGroupDialog.cpp
+++ b/retroshare-gui/src/gui/gxs/GxsGroupDialog.cpp
@@ -468,6 +468,9 @@ void GxsGroupDialog::updateFromExistingMeta(const QString &description)
     ui.distributionValueLabel->setText(distribution_string) ;
 
 	ui.idChooser->loadIds(0, mGrpMeta.mAuthorId);
+    
+    	if(!mGrpMeta.mAuthorId.isNull())
+		ui.idChooser->setChosenId(mGrpMeta.mAuthorId) ;
 
 	updateCircleOptions();
 }
@@ -552,7 +555,7 @@ bool GxsGroupDialog::prepareGroupMetaData(RsGroupMetaData &meta)
 		std::cerr << " Invalid GroupName";
 		std::cerr << std::endl;
 		return false;
-	}//if(name.isEmpty())
+	}
 
 	// Fill in the MetaData as best we can.
 	meta.mGroupName = std::string(name.toUtf8());
@@ -565,7 +568,7 @@ bool GxsGroupDialog::prepareGroupMetaData(RsGroupMetaData &meta)
 		std::cerr << " Invalid Circles";
 		std::cerr << std::endl;
 		return false;
-	}//if (!setCircleParameters(meta))
+	}
 
 	std::cerr << "void GxsGroupDialog::prepareGroupMetaData() meta.mCircleType: ";
 	std::cerr << meta.mCircleType << " Internal: " << meta.mInternalCircle;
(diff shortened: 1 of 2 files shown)


csoler:
why do you remove the nullstream in grouter??

merkeld:
`template<typename T> p3GRouter::nullstream& operator<<(p3GRouter::nullstream& ns,const T&) { return ns ; }` 
caused a compiling error because `class nullstream: public std::ostream {};` is private inside p3GRouter and the line is outside the scope of p3GRouter. My SDK showed no references to this line and compilation and testing went flawless so I removed it. I'm not sure if this code does what the author intended. It is like a function:

nullstream& function(nullstream& ns, const anytype x){
      return ns;
}

It has no connection to the class p3GRouter despite that nullstream is defined there. If << was a member function of nullstream the syntax would have been ... p3GRouter::nullstream::operator<< ... but p3GRouter:: stands in front of nullstream which just means, that the return and parameter type is defined in p3GRouter, but as nullstream is defined private Clang correctly rejects it, because non-member functions are not allowed to access privately defined classes.

Another solution would have been to place public: before and private: after the nullstream. Like it is done in the apple version (see #ifdef **APPLE**). This works also on gcc.

csoler:
it's only used when GROUTER_DEBUG is enabled in the .cc
Maybe it can be moved in the .cc as static and between two "ifdef GROUTER_DEBUG" as well.

merkeld:
No it is not used at all, because operator<< is just a function not a member of the nullstream class, so it does not overwrite anything in std::ostream. I think the author intentended the following which actually overwrites the << operator in ostream.

```
class nullstream: public std::ostream {
    template<typename T> p3GRouter::nullstream& operator<<(const T&){
                                            //only one parameter ^
                                            //because the 2nd is implicitly this
        return *this;
    }
};
std::ostream& grouter_debug() const
{
    static nullstream null; //Without overwritten << operator nullstream is just an ostream
    return _debug_enabled?(std::cerr):null;
}
```

This compiles well on clang and gcc. I'm going to commit it in my Nyfor/Retroshare repository.

csoler:
the author (me, in fact), intended to have an operator that does not print anything when applied to a nullstream. That is used when GROUTER_DEBUG in on and when multiple grouter are active simultaneously (in network simulation mode). I've been using it for ages and it should compile (it does on ubuntu 14.04 at least).

The way you wrote it is equivalent, I think,

merkeld:
No it is not equivalent, in the old solution grouter_debug()<<"Debug print." with _debug_enabled==false did indeed nothing in the sense, that it prints nothing to std::err. However you are wasting memory because null is in fact an ostream, because it inherits from ostream and the << operator is not overwritten, so every debug print is stored although you intended to drop everything.

Quote from https://en.wikibooks.org/wiki/C%2B%2B_Programming/Operators/Operator_Overloading

```
// binary operator as member function (see Vector2D:: before operator+)
Vector2D Vector2D::operator+(const Vector2D& right)const {...}
// binary operator as non-member function
Vector2D operator+(const Vector2D& left, const Vector2D& right) {...}
```

However after a test I see that my solution does also not override << because it is not virtual in the upper class and thus is not affected by dynamic binding. So I recommend to remove the nullstream completely. It is not used anyway because _debug_enabled is set to true in the p3GRouter constructor and never changed, so return _debug_enabled?(std::cerr):null; in grouter_debug() always gives a reference to std:err. The #ifdef GROUTER_DEBUG preprocessor directives already do what you wan't. However a correct nullstream would allow you to toggle between debug and nondebug mode at runtime by setting _debug_enabled, if you remove the #ifdef GROUTER_DEBUG directives. Right now you have two switches for enabeling debug prints GROUTER_DEBUG and _debug_enabled that is bad because you have to take care of 2 switches for controlling debugging.

I think this should work, but I favor the drop nullstream solution because #ifdef GROUTER_DEBUG already does the job:

```
std::ostream& grouter_debug() const
{
    static std::ostream null(0); //Open stream in failed state, so nothing gets written
    return _debug_enabled?(std::cerr):null;
}
```

PR #338, opened 2016-04-04
Fixed compilation for Clang.

merkeld:
Fixed compilation for Clang. Replaced nullstream with std::ostream null(0), a stream opened with invalid buffer, so that nothing gets written in it. This is a bit dirty I think it would be better to remove _debug_enabled and grouter_debug() and replace grouter_debug() with std::cerr, because debug prints can already switched on and of with GROUTER_DEBUG.

However I wanted to provide a solution which doesn't change too much, so I only removed the nullstream because it unnessessarily stored every debug print instead of dropping it and the operator<< function which is incorrect C++, because it accesses a private class.

Edit:
Link to the GCC-Bug which allows p3grouter.h to compile although it should be rejected: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=41437

I revoked my old pull request, here is the link to the old discussion: https://github.com/RetroShare/RetroShare/pull/337

### DIFF
--- a/libretroshare/src/grouter/p3grouter.h
+++ b/libretroshare/src/grouter/p3grouter.h
@@ -225,20 +225,13 @@ private:
     void handleLowLevelTransactionAckItem(RsGRouterTransactionAcknItem*) ;
 
     static Sha1CheckSum computeDataItemHash(RsGRouterGenericDataItem *data_item);
-#ifdef __APPLE__
-public:
-#endif
-    class nullstream: public std::ostream {};
 
     std::ostream& grouter_debug() const
     {
-        static nullstream null ;
-
+        static std::ostream null(0);
         return _debug_enabled?(std::cerr):null;
     }
-#ifdef __APPLE__
-private:
-#endif
+
     void routePendingObjects() ;
     void handleTunnels() ;
     void autoWash() ;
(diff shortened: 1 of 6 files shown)


csoler:
actually the use of this print/noprint option is to allow a specific GRouter instance to print or not print, when you use multiple instances in the same program (I use to do that in the network simulator to test message routing). So it cannot depend on a #ifdef.

I totally agree that whatever the solution is, it shoudn't waste resources for something that is not used in the released software.

merkeld:
You will still be able to switch debug prints on and off for certain instances at runtime, because I kept the grouter_debug(). Only the nullstream class which was actually an ostream and the operator<< function which was wrong C++ and thus correctly rejected by Clang, got removed.

sehraf:
Thanks for this patch! I was always curios whether clang creates faster or smaller binaries 👍

merkeld:
At the moment some subprojects define QMAKE_CXX=g++, so clang is not used for the whole project. I created a branch https://github.com/Nyfor/RetroShare/tree/build in my private repo to address this issue and to activate clang's link time optimizations.

sehraf:
having all compiling options in one place and not repeated in each sub project is a good idea! For debugging i'd like to use -Og but clang doesn't offer this right now.

PR #346, opened 2016-04-10
Move Chat notify button from search bar to tool bar.

PhenomRetroShare:
As search bar is hidden by default.


csoler:
took me a while to see the difference ;-) 

In the same spirit, I thing we should put the search field in the toolbar as well, ans remove the option to show/hide it. There's no reason to have a search field alone (and hidden!) in an empty place.

csoler:
other suggestions:
- the "message size" that shown in the tooltip is the "html" size. So saying this messages consists in "230" characters when the message is "xx" seems innapropriate. Just remove the tooltip. Only keep a warning tooltip when the message is too big.  
- tool button sizes seem to be absolute, since they appear super small on high DPI screens. Needs fixing too.

PhenomRetroShare:
Done.

PR #348, opened 2016-04-11
Add ntp finder to check Computer Time

PhenomRetroShare:
Some of us are not in time.
Only go to ChatServer FR to see it.

### DIFF
--- a/libretroshare/src/chat/distributedchat.cc
+++ b/libretroshare/src/chat/distributedchat.cc
@@ -128,6 +128,7 @@ bool DistributedChatService::handleRecvChatLobbyMsgItem(RsChatMsgItem *ci)
         std::cerr << "Message item is:" << std::endl;
         cli->print(std::cerr) ;
         std::cerr << std::endl;
+        manageTimeReceivedError();
         return false ;
     }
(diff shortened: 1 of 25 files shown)


csoler:
I don't think that using a NTP finder is an acceptable solution:
- it calls of an external website (I know we do that for extaddr finder already)
- changing system time is not the role of retroshare
- the NTP only indicates you the actual time, not the time difference with your friend nodes, which is what matters in the first place
  The solution to detecting time discrepancies between RS users is IMHO to use the results of the RTT system.

PhenomRetroShare:
1) Yes it calls external, like extadrr, but only when start and when received time error. And this list is editable.
2) RS don't change System time only informs user.
3) If all users are synchronized with NTP, no one have time discrepancies.

csoler:
I'll test it tonight!

sehraf:
i agree with cyril that the RTT service should be used instead of an external website. additionally this notice should only be used once so the user can fix his system time once and for all (so ideally you don't need to check the time on the next startup again)

csoler:
This whole thing should be based on RTT. The cost of calling yet another clearnet service is not compensated by the gain in feature/security/etc.

PR #359, opened 2016-04-27
Improved build files and bugfix.

merkeld:
1. Improved buildfiles:
   -Some subprojects did override compiler flags and commands, which set g++ as the compiler even if clang was chosen.
   -Added link time optimizations for clang. If your toolchain supports it, you can activate it by setting CONFIG += lto in retroshare.pri
   -fstack-protector-strong was added to the compiler flags, although it is set by default in clang and gcc (at least in my distro). However I wanted to raise awareness for security flags and make it easy to add more in the future.
   Edit: Uncommented -fstack-protector-strong because the Travis CI build can't deal with it.
   -Clang users can add sanitizer flags to activate runtime checks for debugging or increasing security by uncommenting certain lines in retroshare.pri (see http://clang.llvm.org/docs/UsersManual.html#controlling-code-generation)
2. Bugfixes
   I fixed some potential bugs found by clang's static analyzer and some compiler warnings.

### DIFF
--- a/libbitdht/src/libbitdht.pro
+++ b/libbitdht/src/libbitdht.pro
@@ -6,39 +6,7 @@ CONFIG -= qt
 TARGET = bitdht
 DESTDIR = lib
 
-QMAKE_CXXFLAGS *= -Wall -DBE_DEBUG
-
-profiling {
-	QMAKE_CXXFLAGS -= -fomit-frame-pointer
-	QMAKE_CXXFLAGS *= -pg -g -fno-omit-frame-pointer
-}
-
-release {
-	# not much here yet.
-}
-
-#CONFIG += debug
-debug {
-        QMAKE_CXXFLAGS -= -O2 -fomit-frame-pointer
-        QMAKE_CXXFLAGS *= -g -fno-omit-frame-pointer
-}
-
-# treat warnings as error for better removing
-#QMAKE_CFLAGS += -Werror
-#QMAKE_CXXFLAGS += -Werror
-
 ################################# Linux ##########################################
-linux-* {
-	QMAKE_CC = g++
-}
-
-linux-g++ {
-	OBJECTS_DIR = temp/linux-g++/obj
-}
-
-linux-g++-64 {
-	OBJECTS_DIR = temp/linux-g++-64/obj
-}
 
 unix {
 	data_files.path = "$${DATA_DIR}"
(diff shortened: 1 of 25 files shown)


libretroshare/src/pgp/pgphandler.cc
```diff
@@ -1978,7 +1978,7 @@ bool PGPHandler::removeKeysFromPGPKeyring(const std::set<RsPgpId>& keys_to_remov
 	char template_name[_pubring_path.length()+8] ;
 	sprintf(template_name,"%s.XXXXXX",_pubring_path.c_str()) ;
 	
-	if(mktemp(template_name) == NULL)
+    if(mkstemp(template_name) == -1)
```
sehraf: this creates a memory leak since mkstemp "returns an open file descriptor for the file" that should be closed  ( http://man7.org/linux/man-pages/man3/mkstemp.3.html )
also this might not work on windows. I had a patch set for this but i never got merged

merkeld: That's a valid point. I removed this line from my branch. However this function should be fixed because mktemp is insecure (see http://man7.org/linux/man-pages/man3/mktemp.3.html section BUGS) and it creates nasty warings. Maybe we can simply replace it with http://man7.org/linux/man-pages/man3/tmpnam.3.html which does not create a file descriptor.

csoler:
Thx for all this! A few questions I have:
- why do you remove &&(cbStore == cbStore) in cacheTransfer::requestCache() ? I mean, I'm not sure it's a problem, but I cannot see where the idea comes from.
- gxs is always on. The software wouldn't work without it now. I think we can remove the option.
- it seems to me that windows doesn't like -O2, that's why we removed it for releases. Has this recently improved?
- all obj previously went into  temp/linux-g++/obj (or -64). Is it still the case?

merkeld:
&&(cbStore == cbStore) is unnecessary because it evaluates to && true and (a && true) is the same as (a). So I simplified this logic expression to reduce complexity.

If DEFINES *= RS_ENABLE_GXS is no longer needed it can be removed.

At the moment object files which contain native code are stored in obj and files which contain llvm-bitcode in llvm-bc (only applies when compiled with clang and link time optimization). But this can easily be changed by setting OBJECTS_DIR = obj in the retroshare.pri file. However this should be set globally at exactly one point in the code and not in every subproject like it is the case right now.

I haven't tried compilation on windows. However -O2 and even -O3 should be safe if you use a modern compiler unless the program depends on undefined behaviour. If this is the case the undefined behaviour should be fixed and not the compile flag. If you really want to deactivate optimization in windows do it globally in retroshare.pri so that it can easily be changed for all subprojects.

csoler:
I read data.cbStore ;-) so yes of course! I wonder why that was here in the first place.

Yes, the define RS_ENABLE_GXS can be made permanent.

I agree with the obj thing.

For the -O2/-O3, I would like to ask the windows team first.

merkeld:
Compiling without RS_ENABLE_GXS causes compiler errors, because there are still some #ifdefs in the code. A global `#define RS_ENABLE_GXS` is possible but I think it would be better to remove the ifdefs and delete the never taken else branches.

sehraf:
i guess as a quick solution adding RS_ENABLE_GXS permanent is fine. But in the long run all #defines should be removed.
anyway i'm looking forward to have some lto 😃

merkeld:
Any feedback from the Windows team concerning the optimisation? https://github.com/RetroShare/RetroShare/pull/385 fixes some warnings, which I already fixed in this pull request. Hesitating results in double work.

sehraf:
maybe @csoler can cherry-pick the risk-free fixes in the mean time

csoler:
I was waiting for some win dev to review it, but I will have to do it myself in the end. Can you plz fix the conflicts? Sorry to have kept you waiting like this.

sehraf:
any update on this? I'm currently preparing the Arch Linux packet to support clang as a second compiler and this patches would help a lot  ( this one to be clear: 0b4097ce42d6a0ec3cc6a23e6319bcad13de138f )

csoler:
old PR that never found its way through the code.

PR #373, opened 2016-05-06
Fix compilation of LibResAPI for MacOS

PhenomRetroShare:


zeners:
this breaks generating webui on incremental build

PR #386, opened 2016-05-16
Add the ability to change OpMode in command line.

PhenomRetroShare:
So you can change OpMode while running via ssh to get more bandwidth.
Usage: RetroShare06 --opmode [full|noturtle|gaming|minimal]


csoler:
Parameters should be handled in rsinit.cc around line 350. This is better than asking Qt to do it, since it allows rs-nogui to also have the same option.

PhenomRetroShare:
This parameter is sent to running instance.
No-gui don't manage it for the moment.

PhenomRetroShare:
Added to RsInit::InitRetroShare.
I let the GUI part for when only one instance is setted.

PhenomRetroShare:
Updated

PR #388, opened 2016-05-16
Change Open existing file link from Question to Warning.

PhenomRetroShare:


csoler:
thx. It's actually needed. But the way the warning states it makes the user believe that somehow RS has determined that the file actually contains malicious code. I prefer:
"Warning: Retroshare is about to ask your system to open this file. Before you do so, please make sure that this file does not contain malicious executable code."

PR #395, opened 2016-05-27
Don't hardcode g++ in the toolchain settings

hasufell:
Although it's already quite ugly that we overwrite QMAKE_CC like this,
hardcoding 'g++' can break the build where prefixed versions of
the toolchain are expected and where QMAKE_CXX is already properly
defined. So we just use QMAKE_CXX variable here.

This fixes build failures on some distributions, e.g.
https://galileo.mailstation.de/jenkins/job/lipidity/36/console


sehraf:
#359 also fixes this .. which is still not merged for unknown reasons

PR #401, opened 2016-06-01
Fix Compilation with unknown pthread_getname_np if not __USE_GNU defined

PhenomRetroShare:


sehraf:
now @sirjenster has to figure out how to get thread naming working on osx 😄

csoler:
apparently the correct way to handle this is to add #define _GNU_SOURCE, before the #include. See "man pthread_setname_np"

PhenomRetroShare:
#define _GNU_SOURCE don't change something...
No pthread_setname_np defined in:
C:\Qt\Tools\mingw491_32\i686-w64-mingw32\include\pthread.h

PR #403, opened 2016-06-02
Fix most compiler warning in libretroshare + cleanup UBUNTU define

sehraf:
This patch set fixes most compiler warnings (gcc) in libretroshare. 
Some warnings (comparison between (unsigned) ints )are still left because i don't know whether i can fix them with a cast or not.

---

The 4th patch renames a define which has a misleading name

---

The patches are separated into categories so in case you aren't happy with one patch you can still cherry-pick the others.

### DIFF
--- a/libretroshare/src/dbase/cachestrapper.cc
+++ b/libretroshare/src/dbase/cachestrapper.cc
@@ -1159,8 +1159,7 @@ bool CacheTransfer::RequestCache(RsCacheData &data, CacheStore *cbStore)
 			 */
 			if ((data.hash == dit->second.hash) && 
 				(data.path == dit->second.path) && 
-				(data.size == dit->second.size) && 
-				(cbStore == cbStore))
+				(data.size == dit->second.size))
 			{
 				std::cerr << "Re-request duplicate cache... let it continue";
 				std::cerr << std::endl;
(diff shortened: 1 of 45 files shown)


csoler:
thx! I'll check them tonight.

For the unsigned/signed problem, most of the time it's a for() loop that uses int over someArray.size() which is uint32_t. So makign the for() variable a uint32_t fixes it.

Still, be very careful when modulos and signed operations are involved with uint32_t. Most of the time the result is not what you expect and these bugs are hard to find afterwards.

sehraf:
when it was clear (like a counter for a for loop) I've changed the variable type. In the other cases I've added a cast that _should not_ break anything (but you never know 😉 )

sehraf:
also that's why I've separated the patches - i'm fine with excluding the "risky stuff"

TheSentry:
@csoler Can you elaborate more on "Still, be very careful when modulos and signed operations are involved with uint32_t. Most of the time the result is not what you expect and these bugs are hard to find afterwards."? Most stuff I found on the net is about performance issues

csoler:
@sehraf : I fixed some of the problems in a different way than you did (in particular some unreferenced variables should have been used to check consistency. So I added the check). This will cause conflicts in your PR. I'll try to merge it manually afterwards.

The most uncertain change I made was to use uint64_t variables for sizes in EncryptedBinFileInterface, except when for calling decryption routines. So there's a few casts and checks added there.

sehraf:
Thx for sorting this out! I still get some compilation errors - will send a pull request with fixes soon

PR #408, opened 2016-06-04
fix Compilation

PhenomRetroShare:


csoler:
I don't understand the change for QT_VERSION
it works here, and I'm using 4.8

Ah ok, I get it: just remove the unicode stuff. I copied it from the .ui.h

PhenomRetroShare:
Removed PR#406 stuff.

PR #426, opened 2016-06-23
This patch set adds an option to use the DHT stunner to qmake

sehraf:
The DHT stunner is used to obtain RS' external ip addr. when it is natted.

@csoler RS runs fine with the stunner being disabled. Though i can't tell if this affects the UDP-stuff in some way for obvious reasons..


sehraf:
ok the UDP stuff still works 👍

PR #436, opened 2016-07-14
Libresapilocal

G10h4ck:
Libresapi now listen on a local socket bind to LOC06_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/libresapi.sock

This has been developed as a feature enabled by default but it is easy to disable ad compile time ( see https://github.com/RetroShare/RetroShare/compare/master...G10h4ck:libresapilocal?expand=1#diff-44c8ac89945db4edd283359b9dc2742eR1 )

It uses Qt library to do so so it is crossplatform under the hood it uses an unix socket when available and windows named pipes on windows that have comparable functionalities

This PR is the first of a series that will bring to RS more modularity so at the end a "liteRScore" may be compiled for android or other constrained environements where most features are not needed and may cause not useful bloat or overload of the system

Take care of reviewing this too
https://github.com/hunbernd/RetroShare/commit/2be7da63b33454c26dd659e418ed4871c7530d52

Thanks!


csoler:
Could you disable it by default? While this is a great feature, it is not used yet while offering a way to remotely control RS. It's better if it gets enabled using a GUI checkbox (for RS GUI, in the future) or a command line argument (for rs-nogui), like what we do with the http access for the webUI.

sehraf:
out of curiosity:
@G10h4ck did you speak with @electron128 about adding Qt as a dependency? Some time ago electron didn't wanted to use Qt (e.g. for json) and instead be independet

csoler:
indeed, that also is a problem.

G10h4ck:
> Could you disable it by default?

It is disabled at runtime (it is enabled when webui is enabled and disabled if not), my enabled referred at compile time.

> @G10h4ck did you speak with @electron128 about adding Qt as a dependency?
> 
> indeed, that also is a problem.

No I could not speak with him it seems he is off-line from quite some time :(
In any case I'll do a commit to disable by default at compile time this feature so by default libreasi doesn't depends on Qt, and if someone (like us working on a QML app) need it can easly enable it at compile time

Cheers!

hunbernd:
When I was testing an older version, I found a problem that causes linking problems under Windows and the resulting executable will not work. I haven't tested it yet, but it is very likely that the current version will not work on Windows if one enables this functionality.

The error is caused by an undocumented qmake behaviour and both libresapi and libretroshare are affected. It didn't cause problems before because there were no Qt dependency in linked static libs.
I will provide more information and a patch if I can at the weekend.

csoler:
Can you confirm that the current code does not depend on Qt? I mean, is it possible to compile/link retroshare-nogui with webUI, but without Qt?

G10h4ck:
Yes I can confirm it, moreover I can confirm that by default webui get compiled while local socket support is not, I would like to receive more feedback on local socket testing having it in main tree would of course help to have more feedback  :)

Cheers!

csoler:
I'm not sure of that: when I compile using default pro files:

g++ -m64 -o RetroShare06-nogui temp/linux-g++-64/obj/notifytxt.o temp/linux-g++-64/obj/retroshare.o temp/linux-g++-64/obj/TerminalApiClient.o     -rdynamic ../../libretroshare/src/lib/libretroshare.a ../../libbitdht/src/lib/libbitdht.a -lsqlcipher -ldl ../../openpgpsdk/src/lib/libops.a -lbz2 -lgnome-keyring -lglib-2.0 -lssl -pthread -lupnp -lthreadutil -lixml -lcrypto -lz ../../libresapi/src/lib/libresapi.a -L/usr/lib/x86_64-linux-gnu -lmicrohttpd -lQtGui -lQtNetwork -lQtCore -lpthread

So it uses QtCore, QtNetwork, etc. I suspect that this is due to libreapilocalserver being set somewhere, as it triggers the use of Qt in libresapi.pro.

hunbernd:
I could fix the linking problem on Windows: https://github.com/hunbernd/RetroShare/tree/libresapilocal2
It needs to be tested on other platforms.

Currently I get broken pipe errors, but I didn't adapt my test program yet, so it is possible that some changes rendered it incompatible.

G10h4ck:
@csoler I cannot reproduce that behaviour with latest commit and untouched .pro files Qt is not linked to retroshare-nogui

g++  -o RetroShare06-nogui temp/linux-g++/obj/notifytxt.o temp/linux-g++/obj/retroshare.o temp/linux-g++/obj/TerminalApiClient.o     -rdynamic ../../libretroshare/src/lib/libretroshare.a ../../libbitdht/src/lib/libbitdht.a -lsqlcipher -lpthread -ldl ../../openpgpsdk/src/lib/libops.a -lbz2 -lgnome-keyring -lglib-2.0 -lssl -pthread -lupnp -lthreadutil -lixml -lcrypto -lz ../../libresapi/src/lib/libresapi.a -lmicrohttpd

Can you try with a fresh build ?

Cheers!

PR #459, opened 2016-08-06
Add possibility to create GXSId directly on GenCertDialog.

PhenomRetroShare:


csoler:
Great. That's a very useful contribution.

Just a few suggestions before we merge this:
- change "Node" to "Node name" in create new node window
- change "GXS nickname" to something meaningful. Suggestions: chatrooms/forums identity. Add "Can be setup later if you need one" in the place holder text.
- avoid the double PGP passphrase question at start. That's not easy and needs a compromise. Options are:
  1 - keep the two passphrase prompts, but make them more easy to understand: add a proper title to the window (by adding more params to notify) so that the user knows he's not asked for a wrong passphrase. Could also be a text  just before the key name, something that tells why the passphrase is asked: "you need to sign your node's certificate, please enter your passphrase", "you need to sign your forum/chatrooms identity, please enter your passphrase".
  2 - keep the passphrase in cache for some seconds (1 sec is enough), and use a QTimer to reset it. 

After thinking about it, I don't like 2. I really prefer 1.

PhenomRetroShare:
Done.

PR #463, opened 2016-08-07
Fix Image Block on RSTextBrowser.

PhenomRetroShare:


csoler:
Are you querying a http resource from inside RS without the user's consent? If so, that's certainly a problem. I'm not sure were these resources come from. Can you ellaborate a little bit plz? I suggest that you add these png into the local resources instead.

PhenomRetroShare:
It's internal resource.

PR #468, opened 2016-08-09
Add Chat Load Embedded Images Option like forum and message.

PhenomRetroShare:


defnax:
seems now chat bubbles still not more works. now need to enable every a settings to chat with bubbles?

PhenomRetroShare:
Fixed

PR #471, opened 2016-08-15
Add AutoHide to RSImageBlockWidget

PhenomRetroShare:


csoler:
normally, when using a QTimer, you need to make sure eventsLocked() is false. Otherwise you face the possibility of a deadlock if a PGP passphrase is requested. RSProtectedTimer is designed for this.

PhenomRetroShare:
Done
Don't know it.
We may make a doc for all this custom class.
(Only you know all of them ;) )

PR #473, opened 2016-08-15
Fix RsCertificate::toStdString by fixing PGPKeyParser::read_125Size.

PhenomRetroShare:
And added some comments, plus values in hexadecimal for better view.

### DIFF
--- a/libretroshare/src/pgp/pgpkeyutil.cc
+++ b/libretroshare/src/pgp/pgpkeyutil.cc
@@ -248,25 +248,25 @@ uint64_t PGPKeyParser::read_KeyID(unsigned char *& data)
 
 uint32_t PGPKeyParser::write_125Size(unsigned char *data,uint32_t size)
 {
-	if(size < 192)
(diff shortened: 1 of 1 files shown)


csoler:
you changed base 10 values for base 16. Fine. But how does this fix the cert bug??

csoler:
ok ;-) thx

csoler:
actually this was wrong. See the latest commits, where I think I made it right. I also removed the hexa, since RFC4880 uses decimal values, and it's easier to correlate it with the code if using non hexa.

PhenomRetroShare:
But you don't take care of "negative" value...
See my uint8_t(b2 + 0xC0) to get addition in byte.

csoler:
you don't need to. Read RFC4880: the value needs to be shifted down by 192, _then_ converted to bytes.

PhenomRetroShare:
OK, it works fine now.

But why did you break indent char? ;)

PR #477, opened 2016-08-17
Fix Msys2 build when MINGW_PREFIX is not defined.

PhenomRetroShare:
Please look at AppVeyor success before merging.

### DIFF
--- a/retroshare.pri
+++ b/retroshare.pri
@@ -47,19 +47,19 @@ win32 {
 	}
 
 	# Check for msys2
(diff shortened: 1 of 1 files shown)


thunder2:
This doesn't work.
The environment MINGW_PREFIX is not empty but "isEmpty(&&(MINGW_PREFIX))" is true.

PhenomRetroShare:
This one, now, works for my test.
Does it for you?

thunder2:
It works for me too. Thank you.

PR #493, opened 2016-09-08
Fixing compilation for MacOS X

ohhmm:
Simple enabling pkg-config not enough because of wrong pkg-config
headers information provided in case of home-brew provided version of
openssl for the moment of the change. Manual path searching added to
resolve this compilation issue.


G10h4ck:
Seems like you are not using the variable you assigned here, OPENSSL_INC did you forgot to remove the hardcoded path in

https://github.com/RetroShare/RetroShare/pull/493/files#diff-228238cde70ba48ccf7d450a8a552d13R25

Cheers!

ohhmm:
Totally yes, indeed.

PR #518, opened 2016-09-28
Fix InternalFileHierarchyStorage::check, error_string will never be

PhenomRetroShare:
shown as check always return true.


csoler:
actually, I removed the "return false" statements because now the method cleans up the hierarchy if something is inconsistent. Can you make sure that if returned false, the caller won't bother, but just print the error?

PhenomRetroShare:
In /libretroshare/src/file_sharing/directory_storage.cc:156

```
void DirectoryStorage::locked_check()
{
    std::string error ;
    if(!mFileHierarchy->check(error))
        std::cerr << "Check error: " << error << std::endl;
}
```

PR #532, opened 2016-10-12
Fixed a bunch of warnings in safe ways

G10h4ck:


csoler:
your change in p3filelists.cc is wrong ;-) I suppose the indent was misleading. The rest seems ok.

sehraf:
are your sure it's wrong?
it was 

```
    if (ref == NULL)
        if(flags & RS_FILE_HINTS_LOCAL)
        {
            if(row != 0)
                return false ;

            convertEntryIndexToPointer(0,0,result);

            return true ;
        }
        else if((uint32_t)row < mRemoteDirectories.size())
        {
            convertEntryIndexToPointer(mRemoteDirectories[row]->root(),row+1,result);
            return true;
        }
    else
        return false;
```

it is 

```
    if (ref == NULL)
    {
        if(flags & RS_FILE_HINTS_LOCAL)
        {
            if(row != 0)
                return false ;

            convertEntryIndexToPointer(0,0,result);

            return true ;
        }
        else if((uint32_t)row < mRemoteDirectories.size())
        {
            convertEntryIndexToPointer(mRemoteDirectories[row]->root(),row+1,result);
            return true;
        }
    } else return false;
```

looks fine to me.

csoler:
yes I am sure. Look at the code: that would make no sense.

sehraf:
ahh i see what you mean 😉 
you are right
That's why indention and brackets are important 😸

sehraf:
mhh it is still wrong, isn't it?
should be
```
    if (ref == NULL)
    {
        if(flags & RS_FILE_HINTS_LOCAL)
        {
            if(row != 0)
                return false ;

            convertEntryIndexToPointer(0,0,result);

            return true ;
        }
    } else if((uint32_t)row < mRemoteDirectories.size()) {
            convertEntryIndexToPointer(mRemoteDirectories[row]->root(),row+1,result);
            return true;
    } else return false;
```
right?

PR #535, opened 2016-10-16
V0.6 rtt4

csoler:


csoler:
This was an experimental PR that is no longer compatible with master.

PR #544, opened 2016-10-22
Fix Some gcc warnings 383 to 208 lines

PhenomRetroShare:
Left explicit #warning and deprecated.
Some lines are wrong indented and produce mal formed code.

### DIFF
--- a/libretroshare/src/chat/rschatitems.cc
+++ b/libretroshare/src/chat/rschatitems.cc
@@ -297,13 +297,13 @@ uint32_t RsChatLobbyConnectChallengeItem::serial_size()
 }
 uint32_t RsChatLobbyBouncingObject::serialized_size(bool include_signature)
 {
-    uint32_t s = 0 ;              // no header!
-    s += 8 ;                      // lobby_id
-    s += 8 ;                      // msg_id
-    s += GetTlvStringSize(nick) ; // nick
+	uint32_t s = 0;              // no header!
+	s += 8;                      // lobby_id
+	s += 8;                      // msg_id
+	s += GetTlvStringSize(nick); // nick
 
-    if(include_signature)
-        s += signature.TlvSize() ;    // signature
+	if (include_signature)
+		s += signature.TlvSize();  // signature
 
 	return s ;
 }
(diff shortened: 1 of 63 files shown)


csoler:
... by adding/removing spaces in all these places???

sehraf:
i guess @PhenomRetroShare want to fix the numerous -Wmisleading-indentation warnings. Looking at #532 i suggest to merge this 😉

PhenomRetroShare:
@sehraf : You're right.
I don't understand why @csoler don't merge it.

csoler:
Mainly because I think that adding {} around a single statement in a if just complicates the code. See for instance rsgxsnetservice.cc:3295. When you have an if into an if, that makes sense because it kills a reading ambiguity (which is not a real ambiguity, but some people dont like it, and I understand why), when in single ifs, it's up-to-me unnecessary.

PhenomRetroShare:
@csoler : it's updated, tell me if you have some remarks.

csoler:
That actually happens in many places.  I just cannot list all of them. I don't see the point in adding all these {} everywhere.

I can see you still fixed a bug in the VideoProcessor.cpp file. Was not easy to find if not thx to the warnings.

PhenomRetroShare:
Sorry, I don't find more single if with added {}.
All are with else or inside other branch.
The warnings were:
warning: this ‘else’ clause does not guard... [-Wmisleading-indentation]

I've updated this branch an fix some new warning too...

csoler:
here's some examples of unneeded {} that you added:
grouteritems.cc:264
rsgxsnetservice.cc:1645
p3cfgmgr.cc:513+515
IdDetailsDialog.cpp:167+169
[etc]

PhenomRetroShare:
grouteritems.cc:264: if have its ones, more readable if else too.
rsgxsnetservice.cc:1645: To be sure to not be in default case.
p3cfgmgr.cc:513+515: same as grouteritems
IdDetailsDialog.cpp:167+169: these ones show you how you can make mistake if no {} ;)
For me, "we have to"(I prefer to) have {} for each time the condition group have more than 2 lines.
If you have only one line put it just after if condition like that:
`if (notOK) return;`

PhenomRetroShare:
@csoler : I've updated this branch as you want...

defnax:
hi, it seems it has now conflicts?

PhenomRetroShare:
Updated

PR #547, opened 2016-10-23
Fix: unable to upload files on Windows

hunbernd:
Caused by multiple unneeded / character in file path.


csoler:
Thx for finding this! Actually I am worried about why dir_parent_path already ends with a '/'. Nothing in the code seems to be likely the cause of this. If so, your change might only work on windows. I will test it on linux asap.

csoler:
ok. It seems that the ending of FolderIterator is system-dependent.

PR #550, opened 2016-10-26
Android

G10h4ck:
Initial work on android porting
- Do some slight modifications to libretroshare so it copile on android
- Some modularization in qmake files  
- Added an Android service that run a nogui-like daemon at phone init
- Added a QML app that connect to the service via unix local socket and is able to create locations, login, display trusted nodes with different colours if them are connected or not, and list GXS identities 

All Android related stuff must be enabled explicitly passing CONFIG+=what_you_need to qmake so because compiling for Android is not that straight forward by default people compiling RetroShare for their own machine will not notice any difference


PhenomRetroShare:
Some error when running under Linux QtCreator Gdb:

```
Debugging starts
&"warning: GDB: Failed to set controlling terminal: Ioctl() inappropr\303\251 pour un p\303\251riph\303\251rique\n"
No value following switch l on command line
No value following switch d on command line
No value following switch i on command line
No value following switch p on command line
No value following switch c on command line
Memory still in use at end of program: 0 bytes.
Debugging has finished
```

PhenomRetroShare:
This is due to /retroshare-qml-app/src/retroshare-qml-app.pro:3
QT += qml quick

libretroshare/src/rsserver/rsloginhandler.cc
```diff
@@ -200,6 +200,7 @@ bool RsLoginHandler::tryAutoLogin(const RsPeerId& ssl_id,std::string& ssl_passwd
 	/******************** OSX KeyChain stuff *****************************/
 #else /* UNIX, but not HAS_GNOME_KEYRING or APPLE */
 
+#ifdef TODO_CODE_ROTTEN
```
csoler: ROTTEN means unused?

libretroshare/src/pqi/pqinetwork.cc
```diff
@@ -317,7 +322,15 @@ bool getLocalAddresses(std::list<sockaddr_storage> & addrs)
 		}
 	}
 	free(adapter_addresses);
-#else // WINDOWS_SYS
+#elif defined(__ANDROID__)
+	foreach(QHostAddress qAddr, QNetworkInterface::allAddresses())
```
csoler, line 326: The policy of not using Qt in libretroshare should be enforced. Can you please find a way to fix this?

libretroshare/src/rsserver/rsloginhandler.cc
G10h4ck: Not just unused, that part of the code doesn't compile anymore if enabled, probably the code around has changed, and that part has remained the same because was not used (this kind of stuff is usually said code rotting :p)

csoler: Then you can suppress it. There's already too many ifdefs here ;-)

libretroshare/src/pqi/pqinetwork.cc
G10h4ck: We need Qt anyway to deploy on Android, btw I have tried alternative approaches (as an example https://github.com/morristech/android-ifaddrs ) before using the one provided by Qt but I have experienced crashes, moreover we already use Qt in libretroshare if libresapi local socket server is enabled, in any case this doesn't depend on Qt by default (only if you need to deploy on android or if you need libreaspi local socket)

libretroshare/src/rsserver/rsloginhandler.cc
G10h4ck: Done ;)

PR #551, opened 2016-10-27
Add I2P BOB support to RetroShare. 

sehraf:
**ready to merge**

This enabled RS to automatically manage I2P.

This also includes a new alternative to pqisslproxy since BOB doesn't provide SOCKS tunnels.

This was tested with i2p 0.9.27 and i2pd 2.10. (Note for i2pd: it may take a while until the tunnel is established (can take up to minutes) even after BOB said that the tunnel is active)

### DIFF
--- a/libresapi/src/api/RsControlModule.cpp
+++ b/libresapi/src/api/RsControlModule.cpp
@@ -414,8 +414,10 @@ void RsControlModule::handleCreateLocation(Request &req, Response &resp)
         }
     }
 
-    if(hidden_port)
-        RsInit::SetHiddenLocation(hidden_address, hidden_port);
+	if(hidden_port) {
+		/// TODO add bob to webui
+		RsInit::SetHiddenLocation(hidden_address, hidden_port, false);
+	}
 
     std::string ssl_password = RSRandom::random_alphaNumericString(RsInit::getSslPwdLen()) ;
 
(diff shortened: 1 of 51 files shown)


sehraf:
mhh i don't have any compiling problems here locally - i'll check that

ASmith-:
Is this a user selected option or forcing all I2P current proxy and hidden node Retroshare operators to use this? As a user selected additional I2P connection option, sure why not. As a non-selected option forcing all current I2P socks5 proxied Retroshare Regular and I2P Hidden Node episite operators to use I2P Bob or dump I2P Retroshare, that is a very bad move which I certainly would not endorse nor encourage.

sehraf:
It's an option. You can still stick to SOCKS

majestrate:
this kind of patch would make i2p far far easier to configure on RS

shantorn:
I would be interested in trying this to further test this branch. I am new to i2p but how better to learn?

sehraf:
I'm appreciate every tester! Please report your results and opinion here or in RS 😃

majestrate:
compiled this branch and tested it out. it's a turnkey setup minus 1 gotcha where you have to enter the ip address of the bob interface, IMO it should default to 127.0.0.1 but be configurable. otherwise it's by far a lot easier than it was previously.

majestrate:
rather, given that you have bob enabled it is turnkey.
for i2pd you just need an entry in `i2pd.conf` like so and it works without any messing around:

    [bob]
    enabled=true

sehraf:
Thx for testing!

fun fact: if you leave the ip field empty it will use 127.0.0.1 😉 
though that might be a good idea to enter 127.0.0.1 as default (since a hidden node can only use 127.0.0.1 anyway)

sehraf:
replaced by #836

PR #555, opened 2016-10-31
Fixed emoticon order

hunbernd:
The emoticon selector widget used to display the emoticons in random order.


sehraf:
Thx! This is/was really annoying

PR #557, opened 2016-11-01
V0.6 ft encryption

mr-alice:
Hi!

I created this pull request so that developers and testers can try this code.  The major changes are:
* FT can be end-to-end encrypted using chacha20 (See my post on internal dev forum for details)
* only when encryption is enforced does ftController request encrypted tunnels. ftServer always accepts encrypted tunnels, but refuses clear tunnels in encryption is enforced too. This should work with swarming.
* GUI for shared files have been improved: icons instead of 'flags' in directory list, merged the two 'browsable' flags into one, added a flag for 'search' which allows to deactivate searching files, hence allowing to hide the real hash of shared files in needed. Reworked ShareManager into a simpler and more interactive class (no additional window needed mostly).

I suppose that the encryption will not be used immediately, but its important to make it already available for when the software switches to it by default.

Please test and report any bugs below.
Mr.Alice.


sehraf (changes requested):
please fix the spaces in libretroshare/src/ft/ftserver.cc. It's a huge diff with hardly any changes except spaces.

mr-alice:
Apparently qtcreator decided to convert all leading tabs into spaces upon saving. So I applied vim-style indentation to the whole file, which should make it right.

PR #570, opened 2016-11-16
Improved total file size and a small bug with a comma in TB string of…

felisucoibi:
… file size 2 CLEANED CODE


felisucoibi:
It breaks with qt5 when sorting for file size, but it does also without the commit, so it seems a qt5 problem with the whole rs.

sehraf:
right, sorting kills retroshare since years 🙃

PR #575, opened 2016-11-18
Fix Crash when checking a big-big file storage.

PhenomRetroShare:
After check lot of file with errors, error_string become too big and RS
crash.


csoler:
I suppose that the check method was implemented as a safeguard, but having so many errors probably means that there is a bug elsewhere. Any clue what causes it?

PhenomRetroShare:
I don't know, but I see a stack trace reporting that a memory error occurs when attempting to allocate memory and error_string was very big...
We cannot clean all user drive ;).

csoler:
What I'm saying is that this error tells us that an other bug causes inconsistencies in the hierarchy.

PR #591, opened 2016-12-03
Made autologin optional at compile time

G10h4ck:
Autologin is disabled by default at compile time, and a warning to
discourage it's usage is printed if it is enabled.
This will make default RetroShare build safer and reduce dependencies as
example we don't depends anymore on gnome keyring is not needed in default
build for linux anymore.

### DIFF
--- a/libretroshare/src/libretroshare.pro
+++ b/libretroshare/src/libretroshare.pro
@@ -151,7 +151,7 @@ HEADERS += $$PUBLIC_HEADERS
 
 ################################# Linux ##########################################
 linux-* {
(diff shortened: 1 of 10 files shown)


csoler:
I think that the appropriate warning should read like this: "Warning: enabling auto-login on some linux distributions may allow someone having access to your session to steal the SSL keys of your node location and therefore compromise your security". It's a bit technical, but it's more accurate and since we dont show it to the average user, it should be fine.

G10h4ck:
Should be ok now ;)

PR #593, opened 2016-12-07
Expose libresapi for distant chat

G10h4ck:
Added macro to deprecate symbols usage in a crossplatform way.
Deprecated Request::mMethod and related stuff that make implementation
 more complex without advantage.
Added /chat/{initiate_distant_chat, distant_chat_status,
 close_distant_chat} to libresapi.
Solved subtle bug in ChatId::ChatId(std::string str) that caused zeroed
 DistantChatPeerId being created.


csoler:
p3msgs.cc:96 is quite scary !

PR #602, opened 2016-12-14
set gxs default storage period to 372 days

dimqua:
... for compatibility with old clients.


csoler:
It's more complicated then that. Indeed, in the GUI for a start, the expected number is 365 days not 372. So you'd need to take care of making sure that it will display correctly as one year. 

I suppose that instead of numbers we should use some abstract constants like "GXS_STORAGE_TIME_ONE_YEAR" everywhere in the code.

What is the effect/gain in switching from 365 to 372 btw?

dimqua:
I agree. With 365, I can't recive some of the oldest posts (dated 11/12/2015, for example).

PR #621, opened 2017-01-02
Cleanup rstokenservice defines

G10h4ck:
Removed unused defines
Marked for future cleanup the ones the needed ones to refactor as inner
enums in future


csoler:
are you saying that these defines are not used??

G10h4ck:
I have removed the unused ones, deduplicated the duplicated ones, and commented some for further cleanup

PR #627, opened 2017-01-05
Updated top bar buttons....

felisucoibi:
This is my proposal for top bar buttons...

HOME
CHAT
FILES
CHANNELS
FRIENDS
PEOPLE
MAIL
FORUMS
LINKS
LOG
OPTIONS
QUIT

I renamed almost all of them and order them too.
Also I removed Add Friend button, because now with home is enough, maybe the actual content of Add Button can be placed in Options.
About is also removed, and i will try to make another PR with About inside Options.
Also I think for the future the main top bar and the mini bar with options and quit should be together not in two blocks.
Also removed the feedplugin from the main code, should be built (i think) outside main RS code as a real plugin, so by default there is no FeedPlugin button when you install RS, should be a way to see somewhere possible plugins for RS, or maybe not activate them by default.
Well that's what i did.

Code needs review.


csoler:
could you please not import changes from other branches to be merged? I would not like to review the code twice, not to mention the possible side effects of mixing up the changes from different branches.

PR #630, opened 2017-01-07
small gui fixes in lobbies id menu and stats window

RetroPooh:
small gui tweaks:
 chatlobby id contextmenu fix icon on ban option, add vote up option;
 chatlobbies - id tooltip - add votes counter;
 statistics dht page moved from first place - it freezes whole client


csoler:
Nice. Could you please fix these?
1 - lobbies participant list -> right click. "show author in people tab" is disabled for your own ID. It should not.
2 - in the forum author tooltip, make the image fit the number of lines (there's one more line now, with the votes)
3 - in forum author tooltip, use "Votes: +2" instead of just "+2"
4 - when last post TS is 0, replace the date by "Never" or "Unknown", instead of Jan 1, 1970.  
Thx for your work!

RetroPooh:
1,3,4 - done.
2 - string count is different for anon and signed idents, so leave it.

csoler:
I would like to merge this. Could you plz take a look at the Travis compilation problems? Thx.

PR #637, opened 2017-01-10
Fix some new warnings and comment some other.

PhenomRetroShare:


csoler:
I'm currently doing lots of changes in p3gxsreputations and p3idservice. I would prefer to hold this PR until the big one I'm working on now is merged. It's likely that most of your modifications will not be needed anymore.

PR #648, opened 2017-01-17
0.6 New Buttons Names

felisucoibi:
Hope can be merged, i cleaned everything just bottons ready and renamed. One day if somebody has time, could code a way to order them "as you wish" and "rename them as you wish" in Options > Apearence, but by default should be something and this is my proposal.

HOME
CHATS
FILES
CHANNELS
FORUMS
MAIL
LINKS
PEOPLE
FRIENDS
LOG 
OTIONS
ABOUT
QUIT

### DIFF
--- a/retroshare-gui/src/gui/ChatLobbyWidget.cpp
+++ b/retroshare-gui/src/gui/ChatLobbyWidget.cpp
@@ -171,8 +171,8 @@ ChatLobbyWidget::ChatLobbyWidget(QWidget *parent, Qt::WindowFlags flags)
 
     int S = QFontMetricsF(font()).height();
     QString help_str = tr("\
-                          <h1><img width=\"%1\" src=\":/icons/help_64.png\">&nbsp;&nbsp;Chat Lobbies</h1>                              \
-	                      <p>Chat lobbies are distributed chat rooms, and work pretty much like IRC.                                      \
+                          <h1><img width=\"%1\" src=\":/icons/help_64.png\">&nbsp;&nbsp;Chat Rooms</h1>                              \
+	                      <p>Chat rooms work pretty much like IRC.                                      \
 	                      They allow you to talk anonymously with tons of people without the need to make friends.</p>                    \
 	                      <p>A chat lobby can be public (your friends see it) or private (your friends can't see it, unless you           \
                           invite them with <img src=\":/images/add_24x24.png\" width=%2/>). \
@@ -181,10 +181,10 @@ ChatLobbyWidget::ChatLobbyWidget(QWidget *parent, Qt::WindowFlags flags)
 	                      <p>The list at left shows                                                                                     \
 	                      chat lobbies your friends are participating in. You can either                                 \
 		  <ul>                                                                                                            \
-			  <li>Right click to create a new chat lobby</li>                                                              \
+			  <li>Right click to create a new chat room</li>                                                              \
 		     <li>Double click a chat lobby to enter, chat, and show it to your friends</li>                      \
 	                      </ul> \
-	                      Note: For the chat lobbies to work properly, your computer needs be on time.  So check your system clock!\
+	                      Note: For the chat rooms to work properly, your computer needs be on time.  So check your system clock!\
 	                      </p>                                      \
 	                      "
                           ).arg(QString::number(2*S)).arg(QString::number(S)) ;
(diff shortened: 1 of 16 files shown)


csoler:
I tried it. What about this:
- change all GUI usages of "chat lobbies" for "chat rooms" ?
- take the "recommend friends to each others" button that was in "Add friend" and put it in the drop menu of the Home tab?
- "friends" and "people" look too similar and that is certainly confusing. I suggest to keep "Network"+"People". Also move back "Network" between "Home" and "Chats".
- what about 'Log"=> "News" ?

felisucoibi:
**change all GUI usages of "chat lobbies" for "chat rooms" ?**
AGREE, the only one place i would like to keep Chats is in the button. The rest for me OK to rename all.
**take the "recommend friends to each others" button that was in "Add friend" and put it in the drop menu of the Home tab?**
AGREE
**>>>>"friends" and "people" look too similar and that is certainly confusing. I suggest to keep**
My opinion, People is people, gxids, and Friends are friends, direct friends, in RS is complicated but you realize soon about the IMPORTANT  differences, like different private chats, different behavior sharing files, a broadcast chat for friends, Network (that btw in source code name is Friends not Network) does not mean what it really is, that is your NETOWRK OF FRIENDS" and in short FRIENDS, if you have NETWORK you think a first commer thinks it's about network config, tor, i2p. udp, tcpi ports etc.

**>>>>"Network"+"People". Also move back "Network" between "Home" and "Chats".**
I put them togehter and in the last place because of at the end you dont' use People or Friends as much as forum, chats, channels, I putted the most used stuff first, almost all of the time we can use only forums, chats, file sharing and channels.

**>>>what about 'Log"=> "News" ?**
News.... well ... OK, I AGREE if you agree with me in the rest :D BUT still i think it's a log, is not only news because of the conected user's log, it could be NEWS if only forums updated, channels updated etc, agree it's news, BUT it's also a log of connected people, unknown connections etc...

**FEED READER PLUGIN**
I would also remove feed plugin as default installed, or maybe disabled by default.... and you have to enable it after first installation in Options.

**Default size of buttons:**
I think almost everybody uses now big resolution screens, by default i would set up them in 64x64, now it's 32x32, or maybe if the resolution is very small 32x32 and the rest 64x64 by default, after that you can change it in Options

**THE REDUNDANT BAR**
I think you said that, there is a redundant bar that repeats the name of the sections, we could remove it also, only leave in a section if necessary, the button says in wich section you are because is marked, not necessary to have a whole bat sayin Chat Lobbies, Forums, Network, using some space, i think is there because of messenger design of 2005...

csoler:
so what about "Friends" -> "Connections" ? I would like a word that insists on the mesh structure of the network. Friends is not appropriate because it's not people we're talking about, but connected computers, on top of which you talk to "People". Another two word solutions would be "Friend nodes". It's longer, but conveys the idea.

felisucoibi:
Maybe F2F NET? F2F Network, for me still is more Friends than anything else, with two words is easier to describe but the idea was to give a single name.... Connections again for a new commer does not describe enough, f2f net is short but also not really easy, Friend Nodes is descriptive enough like Friends Network, but too long both, that's why i changed to Friends, is more one word friends than anything.... is a pitty to not have one easy word.... 

My vote one word Friends
two words Friend Nodes

But for design and remember is easier Friends, RS is a friend to friend network, and my betatesters easy realized about the diference about People and Friends (they did not understand Network name because of it is for direct talk with friends, broadcast messages to friends, and see th friends there etc)

csoler:
It seems that you forgot to add the "recommend friends" button (which originally was in "Add -> recommend friends") into the Home->drop menu button. Except for that, we're good to go! thx

felisucoibi:
I did not forgot, don't know how to do it, my C++ + qt skills are negative. Can you do this? are you going to merge? Bests.
ps: you want me to send a pr wit this to your branch for gui?

csoler:
I already merged it in my branch. So that we can globally improve the GUI while keeping the trunk reasonably consistent. Dont worry about the missing button. I'll add it.

felisucoibi:
Nice,! if you can add also this to your branch, is to hide by default status bar so a newcommer will concentrate in add a friend.
https://github.com/RetroShare/RetroShare/pull/650

csoler:
This has been manually merged to branch v0.6-ImprovedGUI in main repository. So I'm closing it.

PR #649, opened 2017-01-18
Attempt to reduce BW bursts

jolavillette:


csoler:
@jolavillette: this has been tested for a while right? I suppose we can integrate it. Any comments?

PR #650, opened 2017-01-18
Add StatusBar and its widgets hide options.

PhenomRetroShare:
The StatusBar is hidden by default.

### DIFF
--- a/retroshare-gui/src/gui/MainWindow.cpp
+++ b/retroshare-gui/src/gui/MainWindow.cpp
@@ -19,14 +19,15 @@
  *  Boston, MA  02110-1301, USA.
  ****************************************************************/
 
-#include <QMessageBox>
-#include <QString>
-#include <QtDebug>
-#include <QIcon>
-#include <QPixmap>
 #include <QColorDialog>
 #include <QDesktopServices>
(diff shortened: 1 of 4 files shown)


felisucoibi:
love it, thanks!

defnax:
i dont agree it, statusbar needs to be visible by default, for experts maybe they neeed settings.
and now retrohare goes overfloated with lots settings.
i dont know why work on some code, to hide some stuff, instead of help on importend things.

felisucoibi:
I tested it, IT'S PERFECT! i already added to my fork!

csoler:
It's very nice to have. I just would like it more consistent:
- make all the "hide" checkboxes to be a "show" (the parent checkbox "status bar" acts a a disabling item, so the buttons below should be to *show*, not to *hide*).
- check them by default
Then, in the first time start, we may uncheck them automatically if needed, but they should appear by default when running the softare.

thx!

PhenomRetroShare:
Done

RetroPooh:
Hi. natstatus and dhtstatus should be hidden permanently when running hidden node.
discovery exchange display is currently disabled for hidden nodes - is it bug or as intended?

PR #653, opened 2017-01-20
chatlobby option to disable sending typing notifications; friend connect wizard minor tweaks

RetroPooh:


csoler:
how is that useful? I mean,the typing notification is itself very useful to help people not leave while you're typing. I don't see a situation where this would not be wanted.

RetroPooh:
1. Why not?
2. https://www.bustle.com/articles/39042-8-times-the-typing-awareness-indicator-is-the-worst
;)

csoler:
why not? Simply because we have tons of checkboxes and options everywhere. I just want to keep the useful ones only.

RetroPooh:
Many other messaging software have that option, so i'm not the first one who invents it and not first who needs it too.
If you worry about scarying newbs with tons of options, i'll leave it for advanced mode only.

felisucoibi:
I like to know when somebody is writing, it happens in fb chats, telegram and other messaging stuff people is used to it.... but for advanced options if somebody needs to hide it... ok, btw wich other messages software allow to disable when writing just curious? Bests.

sehraf:
> wich other messages software allow to disable when writing just curious?

Conversations (normal options) and psi+ (advanced options) allow this 😉

PR #657, opened 2017-01-21
Add no cpp #warning Config flag to better show other ones.

PhenomRetroShare:
Use CONFIG+=rs_nocppwarning to qmake command line.


csoler:
I agree with this, but it should be the opposite: the default should be to *show* the warnings, and then with an option, to hide them. We'll use that option in the compilation scripts for packaging, for instance, but it's important for the devs to see the warnings, without the need to add an extra option.

PR #658, opened 2017-01-21
Fix Compilation

PhenomRetroShare:


csoler:
I think that changing service_type to uint16_t is wrong. The value passed is a uin32_t that is (service_type << 8) + [something]. So if you chunk it into 16 bits, you lose the info.

PR #659, opened 2017-01-21
Fix slow emoticon loading

hunbernd:
by using single regexp for :emoji_name: and (emoji_name)

It is currently containing some debug output.
Run it with `./RetroShare06.exe 2>&1 | grep "\-\-\-"` and send me the result if something is broken.


hunbernd:
I've found some unexplainable bug `somerandomtext ::` and text will be missing after posting
I have to fix it.

csoler:
ok. I let you fix that bug before we can merge this PR.

PR #682, opened 2017-02-10
V0.6 edit posts

csoler:
To be tested.

New version fo post do not propagate to old peers, because of an existing bug in message signature validation, that is also fixed in this PR.


csoler:
I'm not merging this before 0.6.2, since it needs to be tested and does not represent a critical fix either.
For testing:
- right click on a forum message and select "Edit post". This only is available if your own identity posted that message. 
- messages with multiple versions have a combo box with the date, that allows to switch between all existing previous versions. 
Warning: post edits will not propagate to old peers, because of a bug, that has also been corrected in this PR.

PR #689, opened 2017-02-18
fixing compilation with openssl-1.1.0 (part 1)

csoler:
I fixed the compilation with openssl-1.1.0. But I need someone to test the code for me. Here's a roadmap:

You'll need to have openssl1.1.0 installed on your system, as a normal package. Don't try to install it in /usr/local/openssl, or whatever in the same system than an existing openssl-1.0.0 version (I tried: it's a mess mainly because sqlcipher is compiled with 1.0.1 in his case).

- checkout PR/689, or github.com/csoler/Retroshare/ branch v0.6-SSL110Fix
- compile

Then check these:

1 - launch an existing node. Does it run? [this tests signature validation in your own certificate and PGP decryption of your pem file]

2 - does it connect to other nodes? Try both TCP-IN and TCP-OUT [this tests signature validation with PGP over SSL cert for friend's certs]

3 - try UDP. This takes more time, as TCP is used in priority. You can force listenning on udp only with -u, but you still will do TCP-out. [this tests tcp on udp code]

4 - try distant chat. Can you open it, send/recv messages? [this tests DH key echange]

5 - create a new node (node key + location). Can it connect to other nodes? [ this tests signature of own certificate and creation of PGP cert]

6 - create a second new node. Try to connect to the other new one [tests signature validation over new code for new signature]

7 - does options files load ok? If not you should lose your config when restarting. Easy to see. [tests ssl encryption/decryption code]

8 - create a signed GXS identity. Use it to chat. Does it work? [tests GXS sgnature by PGP key]

9 - try file transfer: does e2e encrypted tunnels work ok? Try in forced mode to make sure you get them only. [tests chacha20 code and HMAC authentication]

Many thx to anyone who can do that!

### DIFF
--- a/libretroshare/src/crypto/chacha20.cpp
+++ b/libretroshare/src/crypto/chacha20.cpp
@@ -548,7 +548,7 @@ bool AEAD_chacha20_sha256(uint8_t key[32], uint8_t nonce[12],uint8_t *data,uint3
 #else
        HMAC_CTX *hmac_ctx = HMAC_CTX_new();
 
-       HMAC_Init_ex(hmac_ctx,key,32,EVP_sha256()) ;
+       HMAC_Init_ex(hmac_ctx,key,32,EVP_sha256(),NULL) ;
        HMAC_Update(hmac_ctx,aad,aad_size) ;
        HMAC_Update(hmac_ctx,data,data_size) ;
        HMAC_Final(hmac_ctx,computed_tag,&md_size) ;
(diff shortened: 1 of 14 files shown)


PhenomRetroShare:
Have to add CONFIG+=no_sqlcipher to get it running.
1: OK but:
  News Feed: receive channel, forum and post list but no data (field left empty)
  Network: Friend List OK with connection, Broadcast OK, Keyring OK
  People: Empty and always Loading... Create new Id don't trig error. but no more in list Get a lot of
    RetroDb::execSQL_bind(): Error preparing statement
    Error code: file is encrypted or is not a database
    RsDataService::storeGroup() sqlInsert Failed
  File sharing: Download OK, Search OK, Friend File OK, My files OK
  Chat Lobbies: Get the list
    Unsigned can receive but not send all id don't have name.
    Signed can enter but nothing happens. Have to create a new Id at each time.
  Messages: Old messages are here. Can send and receive by node, people is empty.
  Channels, Forums, Posted : All is empty
  Options are restored and saved.
2: OK TCP-in and TCP-out
3: Not tested yet
4: No distant chat as no identity.
I'll test the rest tomorrow...

csoler:
I fixed two important bugs. It seems that GXS data now syncs ok (be careful when testing since sync is limited both in upload and DL by the sync period, which is one month by default).
Only thing I havn't tested is UDP on openssl-110 side. The rest seems to all work after the last two commits.

PR #710, opened 2017-02-28
(gui) add friend wizard tweaks; chatlobby splitter saving; QuickStartWizard to Home; GenCertDialog minor fixes; gxs sync 372 thing

RetroPooh:


csoler:
Thx for all this. I cherry-picked commits 241b160, f775026, 68313c6 , 5d836d6, ebc8299, 09d7455
	

Some notes:
	- peer details email is not used anymore. That's why it is hidden. It's an
	  old compatibility feature with RFC4880.

	- do we need that quick start wizard anymore? I dont really like it.

	- I cannot cherry=pick df754df because of .ui conflicts.

RetroPooh:
we maybe don't need start wizard, but newbs do. there are some good options to configure.
home page was edited in more than one commit, you need to take them all.

> hide the frame showing own cert since it is not needed anymore

 It was unhidden only for advanced user mode. Why:
the original idea was to leave whole Home page to newbs, everybody was fine using button on the mainbar for years... in the end i think we could hide home page entirely for advancedmode enabled, and restore add friend button on Network tab, and there we will need to have own cert functions.
btw, one option from old dialog is missing - add friend certificate from the file

PhenomRetroShare:
You may make a branch and a PR for each.
Track and pull+rebase them.

RetroPooh:
ok, clean version here https://github.com/RetroShare/RetroShare/pull/711

PR #711, opened 2017-03-01
homepage: add webhelp and move quickstartwizard from settings general

RetroPooh:


csoler:
Quick start wizard has many problems now, which is part of the reasons why I wanted to get rid of it. But if you manage to fix them, I'm ok with adding it back. Here's a list:
- the window has a fixed size (probably with explicit numbers that do not account for font/DPI size) so it's unusable on high DPI screens. Use QFontMetricsF(font())/14.0 as a multiplication factor for all sizes.
- some panels are irrelevant up to me, because the default values are fine and easy to change in settings: toolbar/list view. That's a joke. Remove it. Start minimized/do not show message when closing: same: remove the whole page.
- the shared file editing is not consistent with the normal shared file editor. Two options: either remove it (I vote for that), or use the actual SharedManager widget in that file sharing tab. Do not copy the code. Just instantiate the ShareManager widget into it.

Thx!

csoler:
I would like to keep he HomePage as clean/simple as possible. I dont think we need the "Add certificate file" button. It adds unnecessary complexity (duplicates the add friend functionality so will raise unnecessary questions). So I'll merge your PR, but I would like to keep this button hidden.
thx!

PR #716, opened 2017-03-05
Renamed settings name sections now they are the same as buttons

felisucoibi:
Renamed settings name sections now they are the same as buttons


felisucoibi:
Please cyril have a look, is just nothing and makes better the settings for the new release.

PR #717, opened 2017-03-05
Fix Setting list width.

PhenomRetroShare:


csoler:
If you set an absolute maximum size (here 50) to anything, it is bound to fail on high DPI screens.

PhenomRetroShare:
First 50 is to get the minimum, could be 0, but no sense.

PR #721, opened 2017-03-06
chat names coloring made based on id hash

RetroPooh:


csoler:
This is good, but if you dont mind, can you keep the GxsId type as long as possible, instead of casting it to a string? That will avoid unwanted color change the day we decide to e.g. print RsGxsId using upper case letters rather than lower case.

RetroPooh:
sorry, i have no idea what you talking about :)

csoler:
I'm saying: dont convert RsGxsId to std::string. Just pass it as is, and then read its bytes using id.toBytesArray()[i]

RetroPooh:
updated version here https://github.com/RetroShare/RetroShare/pull/741

PR #722, opened 2017-03-07
Tweaks based on requests in Transifex

mestaritonttu:


csoler:
You cannot manually change this file! It is generated using Qt lupdate tool. If you do, it will be erased at next update. The way to go is to modify the strings in the code.

mestaritonttu:
Damn.. I'll have to look into it then.

PR #742, opened 2017-03-07
chat nicks colors based on hash; add option to disable custom font color

RetroPooh:


defnax:
nice, why not merged? we had in the colors for nicks

PR #744, opened 2017-03-10
transfers - add peer names for turtle routed sources

RetroPooh:
ok, it works but please check one thing
p3turtle::getInfo
uses some RsStackMutex thing - i have no idea what is it and if it is needed in the new function, but since it uses same data it probably must be added.


csoler:
it's fine except for the change in tunnel names. E2EE is really not obvious ;-). Instead of # to replace 0x, you can simply remove the 0x. Thx!

RetroPooh:
meh. just merge it and fix as you like.
i was trying to make text most compact but still informative. plus equal in size for better quickreading :)

PR #748, opened 2017-03-14
Fix Clang Warnings

PhenomRetroShare:
warning: 'bdnet_inet_ntoa' has C-linkage specified, but returns user-
defined type 'std::string' (aka 'basic_string<char>') which is
incompatible with C [-Wreturn-type-c-linkage]
std::string bdnet_inet_ntoa(struct in_addr in);
            ^


csoler:
You're returning a pointer to a temporary memory location. This is bound to fail. I dont understand why you wanted to replace the return of std::string by a char* in the first place.

PhenomRetroShare:
You're right, see that with cpp checker. I'll change it by using it by argument with strcpy.
I've change it because clang warning (look at comment) std::string is not in C.
But in the form do you agreed by fixing warning with multi commit like that?

csoler:
yes I agree. It's much easier for me to proof read them. thx.

PhenomRetroShare:
Is it better now?

PhenomRetroShare:
@csoler : All done.
Let only these:
```
/libresapi/src/api/RsControlModule.h:35: warning: 'override' keyword is a C++11 extension [-Wc++11-extensions]
        virtual bool askForPassword(const std::string &title, const std::string& key_details, bool prev_is_bad , std::string& password,bool& canceled) override;
```

Waiting c++11.

csoler:
Hi phenom. I'm commenting about the ones I dont merge:

6a6779c: actually, the problem with bdnet_inet_ntoa would have been solved in a much easier and elegant way if simply moving that function out of the "C" linkage brackets. The problem with your current solution is that it does not check (it cannot!) the size of the data array passed in parameter to write into and therefore can corrupt memory.

7139e53: some of the unused variables are actually used in defines, that can be enabled by the devs. Some others are really not used. I would prefer to not merge this commit to avoid losing track of these variables.

d68ea15: asc2bin() should return an array of uin8_t, not char. I think that's where the error is.

901d953: I'm not sure drbob didn't want to use this later on. I prefer to keep it.

26b4a28: I dont understand this one. Is sizeHint() not virtual in the parent class?

a15a302: this is actually done on purpose. Numerical Recipes code is always using arrays starting at 1, so you need to call them at index -1 to work ;-) It's less dangerous than changing their code that is very hard to follow.

4df1365: you can't just remove the fabsf! Replace it with abs. Otherwise the expression computes an inconsistent value.

624fac6: let's wait for c++11

ac88456: I'm not sure about this one.

All other comits are merged. Thx!

PhenomRetroShare:
Hello,
6a6779c C-linkage: OK in PR #776
7139e53 unused variables: No, The ones used I add #ifdef. Look at libretroshare/src/pqi/pqiservice.cc:30. And other stay in comments so it's easy to reactivate them. And some use find(), that use cpu for nothing.
d68ea15 implicit conversion: Right in PR #776
901d953 private field m_UPnPControlPoint not used: I can add define too?
26b4a28 RSGraphWidget::sizeHint: It's defined like this in QFrame:
    QSize sizeHint() const Q_DECL_OVERRIDE;
And like this in QWidget:
    virtual QSize sizeHint() const;
a15a302 array index -1: That's why I clean there code too. It's only to move bullets...
4df1365 using fabsf: lst.back().first is qint64 and ms is qint64 too. Why get absolute value of integer? The mistake is to multiply (qint64)_update_period_msecs by 1.2.
624fac6 run override: Yes
ac88456 getPeerStatus: All of them are never called. The warning is generated because of argument type.

PR #760, opened 2017-03-26
Added Windows support to retroshare-android-service

kdebiec:
Small changes that helps interact with libresapi and retroshare-android-service.

### DIFF
--- a/libresapi/src/api/ChatHandler.cpp
+++ b/libresapi/src/api/ChatHandler.cpp
@@ -159,10 +159,6 @@ ChatHandler::ChatHandler(StateTokenServer *sts, RsNotify *notify, RsMsgs *msgs, 
 	addResourceHandler("initiate_distant_chat", this, &ChatHandler::handleInitiateDistantChatConnexion);
 	addResourceHandler("distant_chat_status", this, &ChatHandler::handleDistantChatStatus);
 	addResourceHandler("close_distant_chat", this, &ChatHandler::handleCloseDistantChatConnexion);
-
-	addResourceHandler("private_lobbies", this, &ChatHandler::handlePrivateLobbies);
-	addResourceHandler("subscribed_public_lobbies", this, &ChatHandler::handleSubscribedPublicLobbies);
-	addResourceHandler("unsubscribed_public_lobbies", this, &ChatHandler::handleUnsubscribedPublicLobbies);
 }
 
 ChatHandler::~ChatHandler()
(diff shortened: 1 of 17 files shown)


libresapi/src/api/ChatHandler.cpp
```diff
@@ -1197,4 +1204,88 @@ void ChatHandler::handleCloseDistantChatConnexion(Request& req, Response& resp)
 	else resp.setFail("Failed to close distant chat");
 }
 
+void ChatHandler::handlePrivateLobbies(Request &req, Response &resp)
```
electron128: Wouldn't it be easier to just fetch all lobbies, and filter the list on the frontend side?
`let private_lobbies = all_lobbies.filter((lobby)=>lobby.is_private); // replaces 17 lines of backend code`
The lobbies list is not too big (not like a files list with 10e6 entries), so i can't see an advantage of server-side filtering.
But i see the disadvantage of having a new function for every combination of "subscribed" and "is_private".

libresapi/src/api/PeersHandler.cpp
```diff
@@ -121,6 +177,79 @@ static bool have_avatar(RsMsgs* msgs, const RsPeerId& id)
     return size != 0;
 }
 
+void PeersHandler::handleGetStateString(Request& req, Response& resp)
+{
+	{
+		RS_STACK_MUTEX(mMtx);
```
electron128, line 204: no state token for own state string?

libresapi/src/api/PeersHandler.cpp
```diff
@@ -121,6 +177,79 @@ static bool have_avatar(RsMsgs* msgs, const RsPeerId& id)
...
+		state_string = "undefined";
+
+	resp.mDataStream << makeKeyValueReference("state_string", state_string);
+	resp.setOk();
+}
+
+void PeersHandler::handleSetStateString(Request& req, Response& resp)
+{
+	{
+		RS_STACK_MUTEX(mMtx);
```
electron128: no state token?

libresapi/src/api/RsControlModule.cpp
```diff
@@ -456,6 +458,123 @@ void RsControlModule::handleCreateLocation(Request &req, Response &resp)
...
+
+		// Not found. Store in the queue and emit a signal.
+		//
+		std::cerr << "NotifyTxt:: deferred signature event requeted. Pushing into queue" << std::endl;
+
+		SignatureEventData *edta = new SignatureEventData(data,len,*signlen, reason);
+
+		_deferred_signature_queue[chksum.toStdString()] = edta;
+	}
+	handleSignatureEvent();
```
electron128: do we need all the "SignatureEventData" and "_deferred_signature_queue" stuff, if we perform the call in the same thread anyway?
Maybe only the QT GUI needs this stuff, because NotifyQt::askForPassword creates a dialog box synchronously.

libresapi/src/libresapi.pro
```diff
@@ -3,13 +3,48 @@
 TEMPLATE = lib
 CONFIG += staticlib
 CONFIG += create_prl
-CONFIG -= qt
+CONFIG += qt
```
electron128: I don't want libresapi to depend on Qt. Can you make SettingsHandler optional?

electron128 (commented):
Hi, thanks for your patches. I added some comments to the diff.

retroshare-android-service/src/service.cpp
```diff
@@ -39,11 +39,19 @@ int main(int argc, char *argv[])
 #endif
 
 	QCoreApplication a(argc, argv);
+
+	RsInit::InitRsConfig();
```
electron128: is there a special reason to call RsInit::InitRsConfig()?
IMHO it should not be necessary, because RsControlModule already calls it.

kdebiec: Unfortunately, without this line it won't even compile (Have I forgotten about something?).

electron128: is it a linker error?
If yes, then libretroshare has to come after libresapi on the linker command line.

PR #767, opened 2017-04-08
Fix comment view when they have line return.

PhenomRetroShare:
Before, they grow row height but only show 1 line if it is elided.
Now, height stay at one line, but tooltip show the correct message.


felisucoibi:
+1 this kinf of things makes channels, links comments unusable in real world.

felisucoibi:
tested, would be nice a textarea instead of a line for each comment...

PR #773, opened 2017-04-15
V0.6 identity

csoler:
This PR adds the ability to copy/paste identities into links that contains all public parts of the identity. The technique that is used can easily be extended to share groups of any kind.


dimqua:
>to share groups of any kind.

Do you mean 'Circles', in particular?

csoler:
it can be extended to share circles. The problem would be that because circles need to a connected set of friends, you may well be requesting membership in a circle that you imported using a link, but nobody will be able to grant your membership.

PR #774, opened 2017-04-16
Fix Lobby Topic not send on invite.

PhenomRetroShare:


csoler:
You did right, but this change is not backward compatible: old peers will not be able to deserialize the new invites and the reverse is also true. Among potential things that can be done:
- change the old item name to *_deprecated and still deserialize and handle it (don't try to put an if in the item deserialisation code to handle both items in one place. This is not going to work with the new serialization) 
- add a new item type, with a new sub_item number, and send this one only. 

=> new peers will still receive invites from old peers (with missing topic)

If you want old peers to receive invites from new peers, you'd need to send invites in both formats. I dont think that this is necessary. The gain in compatibility is not worth the increase in bw.

This is rather important because the invite system is what makes the lobbies connected. If you break it, it's likely that many people will end up alone, or in two separate sub-lobbies, one for each version of RS.

PhenomRetroShare:
@csoler: Updated with new serializer.

Issue #772, opened 2017-04-13
RetroShare links should use % econding for + sign


G10h4ck:
I have been struggling with URI.js that apparently got crazy parsing retroshare certificate links, it was apparently putting magic spaces in the cert, but it wasn't URI.js fault is that retroshare isn't escaping URL correctly when we generates links

https://en.wikipedia.org/wiki/Percent-encoding#Percent-encoding_reserved_characters

This should be fixed ASAP

G10h4ck:
fixed by https://github.com/RetroShare/RetroShare/pull/777

PR #777, opened 2017-04-18
Encode Radix in RSLinks.

PhenomRetroShare:
Because it can contains special characters like '+', '/'.
But this is not backward compatible.
Fix Issue #772.


G10h4ck:
Thank you @PhenomRetroShare, IMHO this should be merged ASAP @csoler

PR #781, opened 2017-04-20
Added few requests handling in libresapi

kdebiec:


libresapi/src/api/IdentityHandler.cpp
```diff
@@ -282,6 +324,201 @@ void IdentityHandler::handleOwnIdsRequest(Request & /*req*/, Response &resp)
...
+
+	RsGxsIdGroup data;
+	std::vector<RsGxsIdGroup> datavector;
+	if (!mRsIdentity->getGroupData(token, datavector))
+	{
+		resp.setFail();
+		return;
+	}
+
+	data = datavector[0];
```
electron128, line 397: please check for !datavector.empty() before accessing the first element.

libresapi/src/api/IdentityHandler.cpp
```diff
@@ -282,6 +324,201 @@ void IdentityHandler::handleOwnIdsRequest(Request & /*req*/, Response &resp)
...
+	{
+		if (data.mMeta.mGroupFlags & RSGXSID_GROUPFLAG_REALID_kept_for_compatibility)
+			pgp_name = "[Unknown node]";
+		else
+			pgp_name = "Anonymous Id";
+	}
+	resp.mDataStream << makeKeyValue("pgp_name", pgp_name);
+
+	time_t now = time(NULL);
+	resp.mDataStream << makeKeyValue("last_usage", difftime(now, data.mLastUsageTS));
```
electron128: i would prefer to return absolute timestamps. Or do you have any objections against absolute timestamps?

libresapi/src/api/IdentityHandler.cpp
```diff
@@ -282,6 +324,201 @@ void IdentityHandler::handleOwnIdsRequest(Request & /*req*/, Response &resp)
...
+	RsIdentityDetails details;
+	mRsIdentity->getIdDetails(RsGxsId(data.mMeta.mGroupId), details);
+	StreamBase& usagesStream = resp.mDataStream.getStreamToMember("usages");
+	usagesStream.getStreamToMember();
+
+	for(std::map<RsIdentityUsage,time_t>::const_iterator it(details.mUseCases.begin()); it != details.mUseCases.end(); ++it)
+	{
+		usagesStream.getStreamToMember() << makeKeyValue("usage_time", difftime(now, data.mLastUsageTS));
+		usagesStream.getStreamToMember() << makeKeyValue("usage_service", (int)(it->first.mServiceId));
+		usagesStream.getStreamToMember() << makeKeyValue("usage_case", (int)(it->first.mUsageCode));
```
electron128: every call to getStreamToMember() will make a new object. It would be more logical to have one object in the array for each use case. Stream operator << returns a reference to the stream, so you can just chain the values.

    usagesStream.getStreamToMember()
        << makeKeyValue("usage_time", difftime(now, data.mLastUsageTS))
        << makeKeyValue("usage_service", (int)(it->first.mServiceId))
        << makeKeyValue("usage_case", (int)(it->first.mUsageCode))
    ;

libresapi/src/api/PeersHandler.cpp
```diff
@@ -106,6 +108,84 @@ bool peerInfoToStream(StreamBase& stream, RsPeerDetails& details, RsPeers* peers
...
+		return "Connected";
+	} else if (peerState & RS_PEER_STATE_UNREACHABLE) {
+		return "Unreachable";
+	} else if (peerState & RS_PEER_STATE_ONLINE) {
+		return "Available";
+	} else if (peerState & RS_PEER_STATE_FRIEND) {
+		return "Offline";
+	}
+
+	return "Neighbor";
```
electron128, line 122: i'm not happy with sending user interface strings in json. But i don't have a better solution at the moment, so i have to accept it the way it is.
What we would need, is some kind of enum.

G10h4ck:
Be careful we do not overlap code

https://github.com/G10h4ck/RetroShare/commit/5ee517b64fef49a2c06c28ba9a6af29ff372bda3

ATM it seems we are not re-implementing the wheel each other ;)

PR #785, opened 2017-04-24
Add warning about relays bandwidth.

PhenomRetroShare:


csoler:
Sorry, but this complicates things more than it should. Relays is an obscure feature that should be kept in the relay tab (if possible, ti should be moved to the Network tabs).

csoler:
I suggest that instead of adding this information in server, just add it in relays in the form of "this bandwidth adds up to the max bandwidth". So all relay stuff is kept in relays. I'm not sure about keeping relays. They are not used, and somehow leak IP addresses, so we might not keep them forever.

PhenomRetroShare:
Done

PR #795, opened 2017-05-02
Exposing rsnotify from libretroshare to API

kdebiec:
This is very poor notification system that works as the rest of libresapi via polling. It expose rsnotify from libretroshare to API.

### DIFF
--- a/libresapi/src/api/ApiServer.cpp
+++ b/libresapi/src/api/ApiServer.cpp
@@ -233,6 +233,7 @@ public:
     ApiServerMainModules(ResourceRouter& router, StateTokenServer* sts, const RsPlugInInterfaces &ifaces):
         mPeersHandler(sts, ifaces.mNotify, ifaces.mPeers, ifaces.mMsgs),
         mIdentityHandler(sts, ifaces.mNotify, ifaces.mIdentity),
+	    mNotificationHandler(sts, ifaces.mNotify),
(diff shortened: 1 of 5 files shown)


PhenomRetroShare:
Warning, you've got error:
https://travis-ci.org/RetroShare/RetroShare/jobs/228075987#L2848

electron128:
This is what i don't like about this addition:
- it only supports one client. So it is totally unusable for the webinterface.
- it needs explicit event handling on the client side. The idea of the state token system was: state change handling on the client is the same for every single bit. So only a single function is required to handle every possible state change. The client has a map<resource_url, token>. If a token is invalid, the client would just download the resource again and redraw. With different notify events for different resources, one has to figure out which event is required for which resource. And if you forget to subscribe to an event, you have a bug. Now multiply this with 20 events.

In my opinion, life is too short to think about "when do i have to update what" and "oops i forgot to update the widget". Making GUIs is much more fun with automatic data binding. QML uses automatic data binding, the webinterface uses automatic data binding, so the API should also provide automatic data binding.

G10h4ck:
After Thunder fix, now we have working C++11 builds on windows too, so there is no more reasons to stick to older C++, C++11 will be required anyway for GxsTrans that is candidate for merge soon

https://github.com/RetroShare/RetroShare/pull/698

PR #796, opened 2017-05-02
V0.6 serialization

csoler:
This branch totally rewrites the high level serialization code. See the documentation in the new file serialiser/rsserializer.h

The gain is three-fold:
- 12,000 lines of code have been removed
- adding a new serializer is now trivial, and automatically consistent (no possible errors caused by inconsistent size()/serialise()/deserialize() functions.
- all error handling, printing and serialization is handled as the same place.

All changes here are 100% backward compatible. This will be heavily tested before merge.


sehraf:
> 12,000 lines of code have been removed

great work :+1:

PR #816, opened 2017-05-10
Fix Start page password focus.

PhenomRetroShare:
And readd "log in" button.


csoler:
For the record, I dont think that this login button is necessary at all.

PR #818, opened 2017-05-13
Changed: Initial message is added only when retroshare_gui is defined

kdebiec:
This kind of messages should not be added to chat history in libretroshare. However, if they have to be, their author name should be a "system" or "application".


G10h4ck:
It is ok for me too to remove this, in the pex branch I have putted a boolean parameter on that method to enable or disable that extra message

G10h4ck:
This is superseeded by PR https://github.com/RetroShare/RetroShare/pull/874

PR #872, opened 2017-06-05
Smaller rs login logo for better design.

felisucoibi:
The actual logo is too big this one is smaller and fits better width the width of login input areas.
This is how look like the change.
logonew


csoler:
changing the actual image size is not a correct solution: it will look ok on your screen but not on all screens. What you need to do it to resize the image within RS when adding it to the login screen as a fraction of the font size using QFontMetricsF(font()) as a factor. When you resize, use QImage::scaled() with options to scale smoothly.

felisucoibi:
Well, i'm not a programmer, that's all i can do, i close the pr, if somebody has the skills...

PR #892, opened 2017-06-21
Move Relay setting page to network one.

PhenomRetroShare:


felisucoibi:
+1

sehraf:
While i like the idea of having everything (network related) in one place, i must oppose that ServerPage.cpp is getting larger and larger (almost hit 2k lines of code). 
Maybe we can spit it into multiple files (e.g. one for general stuff, one for hidden stuff (maybe even one just for BOB) and so on)

csoler:
you have a point sehraf. Still, this move reduces the visibility of relays, which is too complicated as compared to how often people use it. Let's try this new setting for a while. It will not be difficult to revert it if it happens to be worse.

I agree with the splitting of ServerPage.cpp into multiple files. Could be serverpage_bob.cpp, server_page_relays.cpp, etc

PR #907, opened 2017-06-26
Add General Direct Download setting.

PhenomRetroShare:
Yes, No, Per User

### DIFF
--- a/libretroshare/src/ft/ftcontroller.cc
+++ b/libretroshare/src/ft/ftcontroller.cc
@@ -99,24 +99,27 @@ ftFileControl::ftFileControl(std::string fname,
 }
 
 ftController::ftController(ftDataMultiplex *dm, p3ServiceControl *sc, uint32_t ftServiceId)
-    : p3Config(),
-	last_save_time(0),
-	last_clean_time(0),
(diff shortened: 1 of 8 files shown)


csoler:
Nice work! I just have a few suggestions to improve it a little bit: 
- I suggest to rename mFilePermDirectDL into mFilePermDirectDLPolicy which is more informative
- in loadConfigMap, it seems that the default value of mFilePermDirectDL is not set, so if the file is corrupted or the value is not in the file (which happens at first load) the value is undefined. Maybe it is already set in the class constructor?
- in ftController::setFilePermDirectDL(), add a test before calling IndicateConfigChanged, to make sure the value is different from what is already stored, so as to not call IndicateConfigChanged for nothing (the UI is quite good at calling these methods multiple times...).
- the consts in rsfiles.h (RS_FILE_DIRECT_DL_YES,...) have values 1,2,4, which let the user think they are flags. I would prefer 0x01, 0x02, 0x03 (not 0x04).

PhenomRetroShare:
@csoler: Done
mFilePermDirectDL is initialized in constructor.
I've added other uninitialized values.

PR #915, opened 2017-07-03
Always allow Friend Time Offset Feed item.

PhenomRetroShare:
Even if "Friend Connected" is uncheck in Notify setting page.


sehraf:
could you add a comment instead of just commenting out a single line?
In a few month/years nobody will know why it is commented out and might even thing it's a bug.

PhenomRetroShare:
Done.

PR #924, opened 2017-07-05
Change RetroShare06 to lower case

PhenomRetroShare:
This is important to comply with linux standards.


sehraf:
while you are right, this might break anything that is relying on the executable's name.

PhenomRetroShare:
It's on Cyril todo ;)

sehraf:
That comment was not meant to stop the PR - i'd just like to point out that it could lead to people complaining 😉

csoler:
If you get the time, could you also remove the 06 suffix? It was meant to allow to use both 05 and 06 on the same computer, but that is not needed anymore.

PR #925, opened 2017-07-05
Fix channel comments line wrap

PhenomRetroShare:


PhenomRetroShare:
@csoler  QT4 compilation fixed.

PR #929, opened 2017-07-06
Fixed: Libresapi was throwing exception when requested to add new cer…

kdebiec:
…t which was empty


G10h4ck:
It is ok for me to merge this, just take in account for the future that

void PeersHandler::handleWildcard(Request &req, Response &resp)

do actually too many things depending on which params it get, we should split up all those different functionalities in different API calls, without forgetting about retrocompatibility

PR #937, opened 2017-07-09
V0.6 improved no gui

csoler:
This pull request fixes the terminal API client for several bugs:
- inhability to handle more than 10 locations
- asking for the same location twice after password is entered
- bad echo output on password
- messy text on the terminal making it hard to select the location to load

However, I cannot test the effects of the changes on the hability to log from the web interface. The login from the webinterface does not work in trunk and does not work in the PR either.

Does it cause a problem to the android/QML apps ? Please test and tell me.

### DIFF
--- a/libretroshare/src/rsserver/rsinit.cc
+++ b/libretroshare/src/rsserver/rsinit.cc
@@ -348,18 +348,21 @@ int RsInit::InitRetroShare(int argc, char **argv, bool /* strictCheck */)
 		argstream as(argc,argv) ;
 
 
-		as >> option('a',"auto-login"       ,rsInitConfig->autoLogin      ,"AutoLogin (Windows Only) + StartMinimised")
+		as
+#ifdef RS_AUTOLOGIN
+		        >> option('a',"auto-login"       ,rsInitConfig->autoLogin      ,"AutoLogin (Windows Only) + StartMinimised")
+#endif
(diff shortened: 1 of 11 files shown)


libretroshare/src/rsserver/rsinit.cc
```diff
@@ -270,304 +270,234 @@ bool doPortRestrictions = false;
...
-                dupline[j] = '\0';
-                argv[i++] = &(dupline[j+1]);
-        }
-  }
-  argc = i;
+	/******************************** WINDOWS/UNIX SPECIFIC PART ******************/
+#ifdef WINDOWS_SYS
+		/* THIS IS A HACK TO ALLOW WINDOWS TO ACCEPT COMMANDLINE ARGUMENTS */
+
+		int argc;
```
kdebiec, line 286: shadows a parameter

libretroshare/src/rsserver/rsinit.cc
```diff
@@ -270,304 +270,234 @@ bool doPortRestrictions = false;
...
+		int i;
+		const int MAX_ARGS = 32;
+		int j;
+		char *argv[MAX_ARGS];
```
kdebiec, line 290: shadows a parameter

G10h4ck:
Tested the API both on Android and on Desktop compatibility is not broken, things seems worked more smoothly but I haven't tested so deep to be sure of the improvement.

Issue #933, opened 2017-07-07
Select picture doesn't show pictures


WombatPete:
When I'm sending a message to a connection (or a broadcast), there's a button to insert a picture. Oddly, the pop-up allows you to see a list of files to choose from, or a detailed list... but not PICTURES (thumbnails).

That can't be right, right?

cavebeat:
you are talking about https://retroshare.readthedocs.io/en/latest/user-guide/interface/#broadcast
right? 

The Broadcast does not sends pictures. I'd say it sends them as a file (hashes it and sends a Download Link with the hash). 

so no, no pics in Broadcast. in Chatrooms and messaging it works. though there are some restrictions on size in chatrooms, to reduce the picture spamming.

WombatPete:
No, I meant "broadcast" as in what you see under networks, when you're messaging a group of contacts simultaneously. It's what you're talking about in messaging. There's a button at the bottom to add a picture. It works, but the views available are all text views - the function to choose a picture displays the files only as texts, so you can't see the picture you want. Does that make sense?

cavebeat:
i have to correct, the broadcast send pictures, as you said. my wrong, i just did not use it for years.
the link directs to "broadcast". 

i did have no problem in sending a picture with the attach button right now. 

there are a few buttons: Smiley, attach picture, attache file, settings and search. 

the attach picture selects only images, the attach file "all files (*)"

i am not able to reproduce your issue

WombatPete:
I'm not saying it doesn't work - it does WORK!
When you open a directory on your computer using a file manager, you can generally choose what to see - a list of files; a list with details, like the date created or file type; or thumbnails. When a program has a button to select specifically a picture, you'd expect the view to be "thumbnails." Here, when you press the Attach Picture button, you get to choose a file - but the window that comes up not only doesn't DEFAULT to a view with thumbnails of your image files, it won't even LET YOU view them that way. If I don't know the pictures attached to a file name, I have to open them in another program to know what to choose. It's not fatal, but it seems like it should be easy to do it right. Does that make more sense?

cavebeat:
sry, now i get it. ya, right, thats oddly. 

ping @PhenomRetroShare  any idea for thumbnails in the file select popup. 
similar to GIMP

hunbernd:
It can be fixed by using the native file dialogs.
Just remove this piece of code from everywhere: QFileDialog::DontUseNativeDialog

WombatPete:
[...]

PR #938, opened 2017-07-10
removed QFileDialog::DontUseNativeDialog from src, to enable thumbnai…

cavebeat:
…ls from native load picture dialog

regarding https://github.com/RetroShare/RetroShare/issues/933

before dontUserNative
https://www.pictshare.net/rbtigq2gaj.png

after with native 
https://www.pictshare.net/jejxwcs0p9.png

but i am not sure if QFileDialog::DontUseNativeDialog has been in use for a reason

### DIFF
--- a/retroshare-gui/src/gui/ShareManager.cpp
+++ b/retroshare-gui/src/gui/ShareManager.cpp
@@ -124,7 +124,7 @@ void ShareManager::doubleClickedCell(int row,int column)
 {
     if(column == COLUMN_PATH)
     {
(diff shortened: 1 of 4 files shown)


csoler:
Yes it's used for good reasons: it makes Qt crash 50% of the time. That's also why we implemented a RS version of the QFileDialog method to call QFileDialog with the correct flag always.

cavebeat:
so the solution is _WontFix_ and this PR can be dismissed.

PR #941, opened 2017-07-14
Add [Unsigned] before name in chat for unsigned (anonymous) GxsID.

PhenomRetroShare:


dtschmitz:
```
6:06 PM DTS
anybody here? 
6:08 PM [Unsigned]jennifer



6:08 PM [Unsigned]jennifer
i'm here! 
6:08 PM [Unsigned]jennifer
what am i? chopped liver? 
```
Please add a space character between "Unsigned" and Alias to result in:
[Unsigned] jennifer
                 ^space
Also, double-click on alias in top window no longer inserts into chat when Unsigned shows.

BabushkaRetroShare:
Maybe it would've been better to simply add [Signed] to the ones who are signed or do something similar to what was done with the signed chat rooms? Most of the IDs in chat rooms are anonymous IDs anyway. Also, this feels similar to the way Jews were forced to wear the yellow badge as to be recognized and seen by everyone. Do we really need this?

csoler:
I vote for moving this Signed/Unsigned information into the tooltip. There's already a tooltip in place, so it shouldn't be difficult to do.

sehraf:
As an alternative we could also introduce a small icon indicating whether an ID is signed or anonymous

PhenomRetroShare:
Fixed in PR #949

PR #942, opened 2017-07-15
Fix warnings

PhenomRetroShare:


csoler:
I dont know about 506dd3fe99101255ff3ad9054365abb5e382d7c9: first I dont see why the functions prototypes are different, then if they *are* different, it's a bug and they should be made the same so that overloading works.

Same for c494b66d2c347fa6cf4540aa4d003fbc18d078e3

csoler:
About 5daaf6d71a1849e64b698b54b8aff54c12d483f1, I prefer to have classes than structs. It's an old debate, but we tend to use classes for services and keep structs for small data items.

csoler:
I cherry picked all commits except a few (See comments above). Can you take a look at these overloading problems? It could be serious. thx

PhenomRetroShare:
About getPeerStatus:
This is not used like some other code So I disable it.
But in three commits.

About sizeHint:
Original is: QSize QFrame::sizeHint() const;

PhenomRetroShare:
@csoler 
I've reverted struct/class commit and redone them by changing struct as class.

csoler:
about dae98062bca946bb415a4cf40808fada4d99242d, I'd like to keep it even if it's unused, because it tells you what 0 meant.

csoler:
I dont think that p3DhtMgr is used anymore (73677d05c07033a64a1e43f13d4cc0e025048f4c). I have replaced sockaddr_in by sockaddr_storage, so as to remove the earning without corrupting the service. I'm not sure what we're goign to do with this.

PR #946, opened 2017-07-16
Fix GUI AutoLogin

PhenomRetroShare:


csoler:
I prefer to only use autologin when -U is specified. That allows the user to start without auto-login, by changing the location.

PhenomRetroShare:
@csoler : What's missing for merging?
Need it for GUI.

PR #952, opened 2017-07-19
Fix Lobby auto-subscribe.

PhenomRetroShare:
Save GxsId for each.
Disallow to auto-subscribe to signed lobby with unsigned id.

### DIFF
--- a/libretroshare/src/chat/distributedchat.cc
+++ b/libretroshare/src/chat/distributedchat.cc
@@ -23,8 +23,9 @@
  *
  */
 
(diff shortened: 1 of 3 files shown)


csoler:
line 2007 of distributeschat.cc, you need to first verify that kit->value can be correctly interpreted as a GxsId, and then if yes, put it in the lobby map. Otherwise, you create an invalid entry, which would need to test everywhere else.

PhenomRetroShare:
Done

PR #953, opened 2017-07-19
Fix Ask Password in new window if wrong in StartDialog.

PhenomRetroShare:
And Fix LineEditClear to show placeholder text even if it has focus
(Like Qt5).

### DIFF
--- a/libretroshare/src/pqi/p3notify.cc
+++ b/libretroshare/src/pqi/p3notify.cc
@@ -267,6 +267,12 @@ bool p3Notify::clearPgpPassphrase()
     return true ;
 }
 
+bool p3Notify::setDisableAskPassword(const bool bValue)
+{
+	_disableAskPassword = bValue;
+	return true;
+}
+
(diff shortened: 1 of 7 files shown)


csoler:
I'm not sure it works:
- I entered a wrong passphrase for account A. It does not display anything like "sorry, wrong passphrase"
- then I changed accounts and entered the right passphrase -> login fails. Something's probably wrong.
Also, the idea to highjack p3notify with a "disable passphrase asking" variable is not very elegant ;-). It should be sufficient to do that:
- use getSSLPassword() as you do, displaying an error msg when passphrase is wrong
- once the passphrase is right, put it in cache, and launch the login.
This shouldn't need to disable it from inside p3notify, sinply because notify will grab the passphrase from the cache.

PhenomRetroShare:
@csoler: 
1. Changed to LineEditClear and fix its placeholder text.
2. This is due to getSSLPassword(). I don't know why but new AccountId is not updated...
3. I move disableAskPassword in p3Notify like cache for password.

PhenomRetroShare:
@csoler : I've fixed conflicts.

PR #958, opened 2017-07-21
Fix cpp checker warnings

PhenomRetroShare:


csoler:
- why did you add all these "explicit" ?
- In the copy constructor of AvatarInfo, you need to also delete the memory before assiging it to a new memory chunk, otherwise you cause a memory leak.

PhenomRetroShare:
1. As CppCheck say:
Cppcheck(noExplicitConstructor): Class 'XXX' has a constructor with 1 argument that is not explicit.
And this avert bad assignation.
2. Done in Fix commit ccfe8167ce1adbcbc3135f496cb4ac22fa1fdfb9

PhenomRetroShare:
@csoler : Finished
Only let warnings on
/retroshare-gui/src/util/log.h:111: warning: Cppcheck(copyCtorPointerCopying): Value of pointer 'stream', which points to allocated memory, is copied in copy constructor instead of allocating new memory.
/retroshare-gui/src/util/log.h:96: warning: Cppcheck(copyCtorAndEqOperator): The class 'LogMessage' has 'copy constructor' but lack of 'operator='.
But this file should be removed as not used.

csoler (commented):
When adding a reference, you need to add a const

PhenomRetroShare:
Done.

libretroshare/src/util/rsdir.cc
```diff
@@ -840,7 +840,7 @@ int RsDirUtil::createLockFile(const std::string& lock_file_path, rs_lock_handle_
 	/******************************** WINDOWS/UNIX SPECIFIC PART ******************/
 }
 
-void RsDirUtil::releaseLockFile(rs_lock_handle_t lockHandle)
+void RsDirUtil::releaseLockFile(rs_lock_handle_t &lockHandle)
```
csoler, line 895: missing const

libretroshare/src/serialiser/rstypeserializer.cc
```diff
@@ -255,13 +255,15 @@ template<> bool RsTypeSerializer::deserialize(const uint8_t data[],uint32_t size
 
     r.first = (uint8_t*)rs_malloc(r.second) ;
 
-    ok = ok && (NULL != r.first);
```
csoler: The idea is that because C++ has lazy evaluation, the second condition is not tested if ok==false. This is the reason it's written like this. Besides, ok is a bool, so && is the appropriate operator, not &

libretroshare/src/util/rsdir.cc
PhenomRetroShare: I can't as lockHandle is assigned a value after...

libretroshare/src/serialiser/rstypeserializer.cc
PhenomRetroShare: Done

libretroshare/src/util/rsdir.cc
csoler: well, then dont change it at all! Without the const, it would modified the variable passed as parameter. This particular case is a perfect example of why the const is useful when doing such changes!

libretroshare/src/serialiser/rsserial.cc
```diff
@@ -64,7 +64,7 @@ RsItem::RsItem(uint32_t t)
 class Counter
 {
 	public: 
-		Counter(int i): _i(i) {}
+		static Counter(int i): _i(i) {}
```
csoler: static? It should be explicit. I'm curious whether that actually compiles, and if so, what it means!

libretroshare/src/serialiser/rstlvbase.cc
```diff
@@ -596,7 +596,7 @@ bool GetTlvString(const void *data, uint32_t size, uint32_t *offset,
 
     char *strdata = (char *) right_shift_void_pointer(tlvstart, TLV_HEADER_SIZE);
     uint32_t strsize = tlvsize - TLV_HEADER_SIZE; /* remove the header */
-    if (strsize <= 0) {
+    if (strsize == 0) {
```
csoler: that test is wrong. We should test tlvsize<TLV_HEADER_SIZE. Otherwise if the result is < 0, then the uint32_t strsize will be large  and then the code will allocate some memory it shouldn't . It's probably a security issue!

libretroshare/src/serialiser/rstlvbinary.cc
```diff
@@ -58,10 +58,12 @@ RsTlvBinaryData::~RsTlvBinaryData()
 	TlvClear();
 }
 
-void RsTlvBinaryData::operator =(const RsTlvBinaryData& b){
+RsTlvBinaryData& RsTlvBinaryData::operator =(const RsTlvBinaryData& b){
+	TlvClear();
```
csoler: setBinData() does the clear already.

libretroshare/src/tcponudp/tou.cc
```diff
@@ -265,7 +265,7 @@ int 	tou_connect(int sockfd, const struct sockaddr *serv_addr,
 		return -1;
 	}
 #else
-	UdpPeerReceiver *upr = (UdpPeerReceiver *) (tous->udpsr);
+	UdpPeerReceiver *upr = reinterpret_cast<UdpPeerReceiver *>(tous->udpsr);
```
csoler: reinterpret_cast is extreme. If you can, use static_cast<>.

libretroshare/src/util/extaddrfinder.cc
```diff
@@ -145,7 +145,7 @@ void* doExtAddrSearch(void *p)
 	
 	std::vector<std::string> res ;
 
-	ExtAddrFinder *af = (ExtAddrFinder*)p ;
+	ExtAddrFinder *af = reinterpret_cast<ExtAddrFinder*>(p);
```
csoler: use static cast for void*->type* (https://stackoverflow.com/questions/310451/should-i-use-static-cast-or-reinterpret-cast-when-casting-a-void-to-whatever)

retroshare-gui/src/gui/mainpagestack.cpp
```diff
@@ -72,7 +72,7 @@ MainPageStack::setCurrentPage(MainPage *page)
 void
 MainPageStack::setCurrentIndex(int index)
 {
-  setCurrentPage((MainPage *)widget(index));
+  setCurrentPage(reinterpret_cast<MainPage *>(widget(index)) );
```
csoler: use static_cast

libretroshare/src/serialiser/rstlvbase.cc
PhenomRetroShare: Done

libretroshare/src/serialiser/rstlvbinary.cc
PhenomRetroShare: Done

libretroshare/src/tcponudp/tou.cc
PhenomRetroShare: Done

libretroshare/src/util/extaddrfinder.cc
PhenomRetroShare: Done

retroshare-gui/src/gui/mainpagestack.cpp
PhenomRetroShare: Done and all other

libretroshare/src/serialiser/rstlvbase.cc
PhenomRetroShare: ReDone

PhenomRetroShare: ReDone

dtschmitz:
++ Great Work!

G10h4ck:
This PR has conflicts and stagnated, a discussion on how to make easier to review and merge warning fixing PRs has happened here https://github.com/RetroShare/RetroShare/pull/1382

PR #961, opened 2017-07-25
V0.6.3

csoler:
Campain against memory leaks


tests/unittests/libretroshare/gxs/common/data_support.cc
```diff
@@ -209,10 +213,11 @@ RsSerialType* init_item(RsNxsSyncGrpItem& rsgl)
     rsgl.publishTs = rand()%23;
     init_random(rsgl.grpId) ;
 
-    return new RsNxsSerialiser(RS_SERVICE_TYPE_PLUGIN_SIMPLE_FORUM);
+    if(ser)
+    *ser = new RsNxsSerialiser(RS_SERVICE_TYPE_PLUGIN_SIMPLE_FORUM);
 }
 
-RsSerialType* init_item(RsNxsSyncMsgItem& rsgml)
+void init_item(RsNxsSyncMsgItem& rsgml,RsSerialType **ser)
```
G10h4ck, line 220: What is the purpose of the interface change?

tests/unittests/libretroshare/gxs/data_service/rsdataservice_test.cc
```diff
@@ -35,55 +36,47 @@ void test_groupStoreAndRetrieve(){
...
 
         // check if it exists
         if(gR.find(grpId) == gR.end()) {
             grpMatch = false;
             break;
         }
 
-        RsNxsGrp *l = mit->first,
-        *r = gR[grpId];
+        RsNxsGrp *l = (*mit);
```
G10h4ck: why the parentheses?

tests/unittests/libretroshare/gxs/common/data_support.cc
csoler: two reasons: there was a leak in the unittest due to some code parts not using the return value of these, and also for consistency reasons: the function  that inits an item should not create memory at the same time. I understand that this had been done because there was no way in the calling template functions to know  the type of the serialiser. But I also changed it so it's not needed anymore.

G10h4ck: After reviewing the whole MR I agree that it make completely sense!

tests/unittests/libretroshare/gxs/data_service/rsdataservice_test.cc
csoler: just a side effect of changing from (*mit).something, to (*mit). The () are not needed of course.

G10h4ck: would you get rid of them before merging?

tests/unittests/libretroshare/gxs/common/data_support.cc
csoler: yes. Most of the time we go for small changes (which is good) but at some point these changes may form an inconsistent group of changes, so it needs to be re-written...

PR #994, opened 2017-08-13
Add max data rate option for when idle.

PhenomRetroShare:


PhenomRetroShare:
@csoler ;)

csoler:
How does this work? I understand that you added a IsIdle state to libretroshare so that it can use specific upload/DL speeds limits. But I do not see code in rs-gui to detect that the computer is idle. Does the user have to manually set RS to idle mode?

PhenomRetroShare:
The first lines in diff:
https://github.com/RetroShare/RetroShare/pull/994/files#diff-a1c5e77897bc213df569811153bf10bc00d1ed0d1c2d57394b753abf22fdc746R1442

PhenomRetroShare:
@G10h4ck @csoler 
Updated with last libretroshare master.

PR #1001, opened 2017-08-20
Add Only Hidden Node configure flag

PhenomRetroShare:
To have only hidden node generation append the following assignation to qmake command line "CONFIG+=rs_onlyhiddennode"


felisucoibi:
+1

PR #1004, opened 2017-08-24
Orange skin webUI proposed

digitigrafo:
hi, 
I added the file orange.scss

Instruction to see running: 
https://github.com/digitigrafo/RetroShare/wiki

 I worked as possible, without modify the js/html, but there is inside a lot of style hardcoded. 
Nextstep: Can I throw off the style by the html content?

### DIFF
--- a/libresapi/src/webui-src/app/orange.scss
+++ b/libresapi/src/webui-src/app/orange.scss
@@ -1,13 +1,14 @@
 body {
-    background-color: #666;
-    color: orange;
-    font-family: times;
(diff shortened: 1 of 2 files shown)


G10h4ck:
Really nice!
To me it's a good idea to split the styling from the app too, so green light on that too IMHO

felisucoibi:
say no to hardcoded css, go on moving it to the stylesheet!

csoler:
That's a very nice start! Please dont stop ;-)

PR #1008, opened 2017-09-02
Add a filter to DownLoad list

PhenomRetroShare:


felisucoibi:
Thanks for this, by now the most important thing to be fixed is somehow check for files in download queue in friend file lists and make direct connections if possible, when you download from channels, search results or forums.... maybe this is not the apropiate place but just let you know, and thanks for your work.

csoler:
works perfectly! It's just too bad to lose the vertical space for the search. Can we integrate it in the "File Transfers" line? To be tested.

PR #1009, opened 2017-09-03
V0.6 ft

csoler:
This should fix the bug of available direct file sources not added during transfer. To be tested.


felisucoibi:
it works!

PR #1010, opened 2017-09-04
Fix WebUI API doesn't update IP and Port

myfingerhurt:
Fix WebUI API doesn't update node address and port when adding the certificate, especially for hidden note once address changed, will never be able to get connected again.

Add a check about don't add own gpg_id.

Add an inactive friends list at last seen Unix timestamp API.

Add an remove a specific peer API.

### DIFF
--- a/libresapi/src/api/PeersHandler.cpp
+++ b/libresapi/src/api/PeersHandler.cpp
@@ -32,11 +32,6 @@
 #include "Operators.h"
 #include "ApiTypes.h"
 
-//#include "util/rsdebug.h"
-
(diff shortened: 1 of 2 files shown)


csoler:
Thx for the contribution. However, I cannot review this PR since the diff is over the entire files. Please re-do your PR by limiting your changes to the actual changes in the code. Try not to re-indent the whole files in order to limit the scope of the changes.

myfingerhurt:
Sorry, I have tried, as you can see, I copied the original version and replace my fixes, then based on it, editing line by line in the GitHub web editor, somehow the GitHub still thinking I have changed the whole file......

myfingerhurt:
I will close it and try again.

PR #1019, opened 2017-09-06
Regroup Upload Items by file in TransfersDialog.

PhenomRetroShare:


csoler:
this seems to work fine. It just needs a few things: 
1 - in the parent lines, do not draw anything for the progress bar (For now in looks like an empty bar. Just remove the bar)
2 - in the parent lines, display the total upload speed among children
3 - the "Transferred" value seems to be always wrong. Sometimes it's normal to not have the full value (it's requested asynchronously to the client) but in your PR, it's never displayed correctly (only modulo one megabyte). Maybe it's simply not updated after you received it the first time?
thx!

PhenomRetroShare:
Done

PR #1022, opened 2017-09-07
Move SharedFilesDialog to /gui/FileTransfer/

PhenomRetroShare:
And remove old DEFINE because very olds files are deleted.


csoler:
I'm surprised that the blog UI files have previously been removed. I totally missed that point. Where they empty files?

PhenomRetroShare:
Seems been removed since 2013-09-07 by drbob:
https://github.com/RetroShare/RetroShare/commit/683e75aeefd2e998606a4def80183d8a4b207b3b#diff-d14145bd0e223f12686775bb61ce0103

PR #1026, opened 2017-09-10
V0.6 ft

csoler:


felisucoibi:
Testing...
I just paused hasing, and after a few seconds, the hashing info dissapeared, i cant resume anymore because the play button, and the info is gone. Btw is a good aproach this pause in the bottom bar, but i miss there the hasing speed.

Also the hasing should be a tab in Files tabs, so i can see the progress there with kind of completing bar, the actual file being hased, and a big play/stop button. why? first, the hashing stuff in bottom bar is hidden by default, and can be hided, should be a place where to see ti more explicit and big, instead of use a popup just a tab in Files.


About hide files in Files > Config, should be a default content like .* files, and kind of explanation somehwere to understand hot it works and what can it do.

Thanks for your work.

csoler:
I cannot reproduce the bug. I should probably check that everything is committed. 
Anyway, the pause does not pause in the middle of a file. If you're hashing the last file and you're pausing, the hash will therefore finish and the hash info will disappear.

felisucoibi:
tested again and maybe was that, it works!

PR #1027, opened 2017-09-11
Fix Uninitilized Peer Max Rates.

PhenomRetroShare:
When called by /libretroshare/src/pqi/pqihandler.cc:140


csoler:
well, normally when the function returns false, the variables are not meant to be used. So it should't create an uninitialized situation.

PhenomRetroShare:
Done.
All other calls take care of value = 0. For ui and saving configuration.

PR #1033, opened 2017-09-14
Add Expand/Collapse All in Upload List

PhenomRetroShare:
And don't make child item if only one peer.


felisucoibi:
Everything looks perfect but a clumn is missing net to file name column, user. so with one user downloading we can see the user in question.
COLUMNS NOW;
NAME SPEED SIZE TRANSFERED ETC
COLUMNS AFTER:
NAME USER SPEED SIZE TRANSFERED ETC

felisucoibi:
Hi please, add the column USER and use it with multiline and single line, it don't makes sense this, there is lot of space and i don't know who is downloading, yes i can put the mouse over and wait one second to know, so if y have 14 downlaods i have to wait 30 secs to know who is downloading... makes no sense:
up

PR #1037, opened 2017-09-15
Searchspy

RetroPooh:
display search request strings in router statistics
also fixed search name for advanced search

### DIFF
--- a/retroshare-gui/src/gui/statistics/TurtleRouterDialog.cpp
+++ b/retroshare-gui/src/gui/statistics/TurtleRouterDialog.cpp
@@ -162,7 +162,7 @@ void TurtleRouterDialog::updateTunnelRequests(	const std::vector<std::vector<std
 
 	for(uint i=0;i<search_reqs_info.size();++i)
 	{
-		QString str = tr("Request id: %1\t from [%2]\t %3 secs ago\t\t %4").arg(search_reqs_info[i].request_id,0,16).arg(getPeerName(search_reqs_info[i].source_peer_id)).arg(search_reqs_info[i].age).arg(QString::fromUtf8(search_reqs_info[i].keywords.c_str()));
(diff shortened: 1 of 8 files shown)


retroshare-gui/src/gui/statistics/TurtleRouterDialog.cpp
```diff
@@ -162,8 +162,8 @@ void TurtleRouterDialog::updateTunnelRequests(	const std::vector<std::vector<std
 
 	for(uint i=0;i<search_reqs_info.size();++i)
 	{
-		QString str = tr("Request id: %1\t from [%2]\t %3 secs ago").arg(search_reqs_info[i].request_id,0,16).arg(getPeerName(search_reqs_info[i].source_peer_id)).arg(search_reqs_info[i].age);
-
+		QString str = tr("Request id: %1\t from [%2]\t %3 secs ago\t\t %4").arg(search_reqs_info[i].request_id,0,16).arg(getPeerName(search_reqs_info[i].source_peer_id)).arg(search_reqs_info[i].age).arg(QString::fromUtf8(search_reqs_info[i].keywords.c_str()));
```
G10h4ck: Please pass the string length too to QString::fromUtf8(...) otherwise it will rely on the string being properly terminated with \0 to determine where the array of bytes ends, thus potentially introducing an out of bound read if the string has no ending \0 (it may happen by chance or by malicious intent)

G10h4ck (changes requested):

retroshare-gui/src/gui/statistics/TurtleRouterDialog.cpp
RetroPooh: done, i think

felisucoibi:
+1 small details

csoler:
The display for regular expression is incomplete. We should display something more accurate. 
There's probably a way to derive a toStdString() method in regular expression that calls the child expressions recursively, producing strings like  "(size > 3000) or (hash starts with "0394...")"

I can write this part if you don't feel comfortable with it.

[Edit] ok, I'll do it. You'll just need to call the toStdString() method.

[Edit2] done. Can you update your PR plz? Also, when you call RsRegularExpression::LinearizedExpression::toExpr(expr)->toStdString(), do not forget to delete the result of toExpr(), otherwise there is goign to be a memory leak.

RetroPooh:
Done.
I left search name with old style, since it is more compact and fits better in the narrow window tab

PR #1038, opened 2017-09-15
Add Peer Column in Upload list

PhenomRetroShare:


felisucoibi:
Perfect, there is a small detail for perfection  that i will try to fix myself if you or cyril agree, the name of a anonimous connection is tooo long:
"Anonymous end-to-end- encrypted tunnel 0x2a32fv3123123. "
I don't know wich word use to simplify this text....
The "username" of anonimous tunnel is longer than filename being uploaded....

Maybe "Anonymous E2E 0x2a32fv3123123" ?
"Anon E2E Tunnel 0x2a32fv3123123"?
Anon 0x2a32fv3123123 <- in the future everything is going to be E2E so maybe we can remove the E2E....
Anonymous <- the msot simple

csoler:
I imagine that the only situation where this is useful is when the number of sources is 1. 
For the other cases, the peers where already listed in the Name column. Now it takes far too much space.  
In the 1-peer case, we could still add one child with the peer name. So it would mean always add children, and if there's only one, also copy the progress/transfered values of that children into the parent.

RetroPooh:
felisucoibi
tunnel names may be upgraded with this https://github.com/RetroShare/RetroShare/pull/744
and later id may be omitted completely and all text changed into the icon, so tunnels will be like icon  -  f2f(human icon)/anon(tor icon)/encrypted(lock sign) plus peer name

felisucoibi:
Hi, this PR code is perfect, not more columns needed. With one sight to PEERS colum to see who is downloading. But anon user name is too long in one line or multiline (now users is not in name column is in multiline with this PR), that's why I think we should Use just Anonymous, or maybe what random suggest can be beautiful too with small icons, these is how it looks like in a very width screen the e2e tunnel.

Also add next to the number of anon users downloading the word peer, or user, or tunnel to not whos only the lonley number. Already talked with phenom and included in this PR too.

Also so often appears 0 tunnels uploads, the one in the photo is Gohack testing, i think there is a important bug with anon uploads, after a restart they never start again.... gohack is trying to download the file that exist in my hD, i receive the try to connect that is markedwith the 0 tuunel , but he can't continue downloading.... Also in this PR, thanks phenom.

up

PR #1040, opened 2017-09-15
allow multiple filenames display for a single hash in search

RetroPooh:


sehraf:
Thx! I always wanted this

csoler:
I'm wondering if it would be better to not display the name for the children nodes. This way it will be very clear that the children represent alternate names for the same files.
If you do that, then you need to always climb up to parent when trying to find the hash/size to create links etc.

RetroPooh:
you mean duplicated string entries like hash, size? i think it is possible to make them invisible with font options or something like... don't know how exactly, for now.

felisucoibi:
I liked a lot the tunnel stuff about where it goes, and this feature is nice too, emule have it and it's nice. Also the search spy is nice to see it.

PR #1055, opened 2017-09-28
Changed prefixes and sufixes for starting with and ending with …

felisucoibi:
On internet people use "starting with" and "ending with" instead of prefixes and sufixes, in literature is ok even on internet sometimes, but when you search for files or filter files you use the filter starting with and ending with, is not prefixes is wrong but less understandable in internet language talking about files.

I tested with 2 rs users and they understood how it works with this text.


sehraf:
maybe also remove the space between the last word and the colon?

felisucoibi:
I tested with the space, without it, and with "..." and looked better with the sapce... but for me is ok to remove it if makes more sense.

PR #1077, opened 2017-10-23
Make it possible to send images over chat lobby

hunbernd:
by reducing colors and resolution to fit into message size limit


csoler:
this code does not compile on qt4 because of QtMath. Can you fix this plz before I merge? thx!

felisucoibi:
This PR has a bug that make images huge in private chats.... in public chats works or at least don't break anything but in private messages or chats the images gets huge. Please fix it.

PR #1079, opened 2017-10-24
V0.6 keyring

csoler:
work from sss about creating an item model to make keyring more efficient


csoler:
for some reason the table looks very sparse on my screen. Is that something we can fix?

PR #1083, opened 2017-10-27
V0.6.3 networking

csoler:
This PR supposedly offers the possibility to tie RS to a specific net interface. The option is in preferences->network.

This is for testing only. I have no clue what the effect will be on windows.

Still to be done:
- the net interface is not saved. Restarting RS will set it to "auto" again
- testing. The expected behavior is that when selected to auto, RS uses the first net interface available (except lo), otherwise it will only use the selected one. If not available, RS will stay offline.


felisucoibi:
+1

felisucoibi:
This kind of things with just a simple VirtualBox you can test it.

G10h4ck:
TL;DR IMHO we should not merge this neither do more efforts in this direction.

Long version:
At current merge request state this constrains apply only to incoming connections not outgoing. While recurring to more networking dirty trickery it may be possible to at some extent constraint outgoing connection too, this would not be 100% reliable and the actual behaviour would be influenced by many details specific of the platform, the socket type, the network topology and even the host configuration, thus introducing some way unreliable mechanism making abstraction more and more difficult is a NO-NO for me. 

Below I paste the conversation we had in the chat lobbies with more arguments about do/not do this kind of things:

22:04:03 G10h4ck: i believe touching that part of code is like opening the pandora box...
22:04:39 Cyril: I sggest you make a PR. Because we'll have to fix a few other things to get this to work
22:04:25 G10h4ck: It is a pile of awfully organized #ifdef
22:05:05 Cyril: yeah
22:04:40 anmo: I am working on it
22:05:24 Cyril: also, retroshare/rspeers.h should be split into a rsnetwork and a rspeers
22:05:28 Cyril: thx anapel
22:04:56 sehraf: thx anmo
22:04:58 G10h4ck: moreover windows networking API is crazy... much worse than any unix...
22:05:30 Cyril: anmo
22:05:35 Cyril: bloody completion￼
22:07:57 Cyril: @ghibli: 1 and 2 in PR 1084
22:08:12 Lobby management: anmo joined the room.
22:09:34 G10h4ck: moreover i am not yet 100% sure that this approach really limit the networking to one interface
22:10:09 G10h4ck: I agree that selecting a specific network interface addess for listening connections will accept only connections from that interface
22:10:20 G10h4ck: but what happen for the outgoing connection?
22:10:34 Lobby management: Eugene Tooms has left the room.
22:10:37 Lobby management: Eugene Tooms joined the room.
22:11:13 Cyril: same
22:10:49 G10h4ck: As far as I rembember we don't specify a source address when we open sockets
22:11:31 Cyril: arg
22:11:18 G10h4ck: and also if we do that in the end the routing inside the kernel can do whathever it wants
22:11:21 G10h4ck: arg?
22:12:19 Cyril: it means "if you're right, I've wasted my time"
22:12:29 G10h4ck: I tried to explain this some month ago, but probably my argumenting was not well expressed enough at that time
22:14:31 Cyril: maybe. But still, that surprises me. At least we should be able to disable some net interfaces in outgoing sockets.
22:16:40 G10h4ck: In the end it always will depends on the kernel choice also if we do all in our power
22:17:13 G10h4ck: I am investigating the pqissl code to confirm that we don't specify source address
22:17:18 G10h4ck: it is the usual practice
22:17:44 G10h4ck: I am not even sure one can specify source address in a simple way for outgoing connections
22:17:58 G10h4ck: it is usually the kernel that do such coiches
22:18:55 G10h4ck: seems the code i am looking for is in pqissl::Initiate_Connection()
22:19:24 G10h4ck: int osock = unix_socket(PF_INET, SOCK_STREAM, 0);
22:20:24 G10h4ck: of course tons of #ifdef WINDOWS_SYS


22:20:48 G10h4ck: if (0 != (err = unix_connect(osock, (struct sockaddr *) &addr, sizeof(addr))))
22:20:58 G10h4ck: no source address specified yet
22:21:53 Cyril: no, but you specify the sockaddr, which is tied to the specific net interface
22:22:13 Cyril: that's probably how you choose
22:23:18 G10h4ck: mmmm not sure
22:25:02 Cyril: when you request the net interfaces, you get them sorted by addr. Then each has a char* as a name, but the original entry point seems to be the addr
22:25:09 G10h4ck: if we are doing this by default and not letting the kernel do it i can understood many problem we get some times when attempting connection
22:26:20 Cyril: sure. f the interface is not up, or not available, it will fail. But it's the user's choice
22:26:14 G10h4ck: 
struct sockaddr {
    unsigned short    sa_family;    // address family, AF_xxx
    char              sa_data[14];  // 14 bytes of protocol address
};


// IPv4 AF_INET sockets:

struct sockaddr_in {
    short            sin_family;   // e.g. AF_INET, AF_INET6
    unsigned short   sin_port;     // e.g. htons(3490)
    struct in_addr   sin_addr;     // see struct in_addr, below
    char             sin_zero[8];  // zero this if you want to
};

struct in_addr {
    unsigned long s_addr;          // load with inet_pton()
};
22:26:45 G10h4ck: sockaddr as no way to specify outgoing interface as i tought
22:27:04 G10h4ck: so or we do some bind/sockopt magic at some point
22:27:17 G10h4ck: or the kernel always choses source address and interface for us
22:27:38 G10h4ck: that is ok in 99.9% of cases
22:28:15 Cyril: I think it looks at sin_addr and see what it binds to
22:28:32 Cyril: I agree with you that we should seek some doc about this
22:28:52 G10h4ck: so i can confirm no bind is done before connect
22:29:12 ghibli: 22:07 Cyril: @ghibli: 1 and 2 in PR 1084 ￼
22:29:24 G10h4ck: and no sockopt that could alter this bahaviour is set before connect
22:30:08 Cyril: sure
22:29:36 ghibli: hi gio
22:29:41 G10h4ck: so kernel is actually chosing the outgoing interface and source address and port
22:29:44 G10h4ck: hi ghibli
22:30:41 G10h4ck: the other good news is that as i rembembered we are creating the networking sockets on our oen and then we pass them to openssl probably by file descriptor
22:31:17 Cyril: https://stackoverflow.com/questions/3998569/how-to-bind-raw-socket-to-specific-interface
22:30:58 G10h4ck: so we have access to the tcp socket FD
22:31:25 G10h4ck: and we could use that to do more accurate statistics of sent/received data without magic euristics
22:31:30 Lobby management: Stripe has left the room.
22:32:02 G10h4ck: raw socket are not good reference
22:32:32 G10h4ck: they are very platform specific and you have to build all the frame (IPs MAC and so on)
22:33:31 ghibli: i used to code sockets in c in windows but i forgot everything.....
22:33:46 Lobby management: Phenom joined the room.
22:34:48 G10h4ck: moreover seems they are interested in a listening socket altought I thinks listening has complitely different meanings in RAW and TCP sockets
22:35:26 G10h4ck: btw until now do we agree that we are using specific interface for listening only ?
22:35:56 G10h4ck: I mean in the PR code
22:36:47 Cyril: yes
22:36:52 Cyril: that's what's going on now
22:36:49 G10h4ck: ok
22:37:01 Phenom: Hello devs.
I prefer choose physic interface (enpxxx, wlxxxx, ...) than IP one.
22:37:45 Cyril: of course
22:37:39 G10h4ck: I continue to think this is the pandora box...
22:38:15 Cyril: the way we present things in the UI is not necessarily what's implemented below
22:38:36 G10h4ck: I feel that we should consider this kind of corner cases once we have the networking code understandable and with abstraction in the right places
22:38:51 Phenom: GN8 devs
22:39:26 G10h4ck: It seems to me that introducing this kind of changes right now will probably introduce lot of unexpected behaviours very difficult to debug
22:40:10 Cyril: @G10h4ck: if there is no simple solution, I wont try to do it
22:40:00 G10h4ck: I don'T see simple solution ATM
22:40:42 G10h4ck: take in account that also if you specify source address/interface the kernel has always the final decision, because routing is in its full control
22:41:33 G10h4ck: that limitation will be never solvable unless we do very low level things like packet injection, or raw sockets, that we should never do
22:41:56 G10h4ck: but in general i believe kernel tend to respect those kind of decision in normal situation
22:42:16 G10h4ck: so I believe it can make sense at some point
22:43:40 G10h4ck: also we should not forget that if we attempt to do things at lower level of the networking stack we will pay that price pretty soon
22:44:26 Lobby management: Stripe joined the room.
22:46:11 G10h4ck: ideally we should work only at connection level
22:46:45 G10h4ck: when we attempt to do such "non standard" things like constraining the outgoing interface and so on we are woring at level 1 or 2
22:46:57 G10h4ck: while connection level is 4
22:47:51 G10h4ck: doing this kind of crosslevel things will cause our code to be less understandable and the biggest problem to be impossible to abstract in manageably way
22:48:33 G10h4ck: so for example if one is using TOR too, what does it means to constraint to one interface only in that case?
22:49:12 G10h4ck: does it means that we constraint only non TOR connections?
22:49:31 G10h4ck: or should the user know that we cannot constraint TOR connections too ?
22:51:46 G10h4ck: in the "ideal" internet model if the user want to constraint a certaing application to just one intervase she should use iptables, or the firewall of her system
22:51:48 Lobby management: Stripe joined the room.
22:51:59 Lobby management: #derrundetisch has left the room.
22:56:31 Cyril: you're right G10h4ck
22:57:51 Lobby management: Jimbob joined the room.
22:59:52 Lobby management: iwi8305 joined the room.
23:05:17 G10h4ck: do you mind i paste what i wrote in the PR ?
23:06:02 Cyril: go on
23:05:39 G10h4ck: it is to not lost this things
23:05:54 G10h4ck: so next time we think about this we know what happened ;)

hunbernd:
It would be still useful for selecting which interfaces to announce through discovery or DHT.

PR #1085, opened 2017-10-28
network cleanup: split unix/windows code

sehraf:
After the discussion yesterday i decided to make this small clean up patch. 
I'm not sure if it is worth merging. I just want to show how the unix/windows code can be handled without any \#ifdefs (expect one for including the right file 😉 ). 

The idea is to have a ops struct that contains function pointers for every function that is OS specific. These functions pointers get set to the corresponding OS specific function once and the rest of the code just calls the function pointer. This results in a much cleaner code.
Since the new code doesn't use any run time ifs the execution speed should be similar to the old version (+ the extra function call).

(tested on Linux)

### DIFF
--- a/libretroshare/src/libretroshare.pro
+++ b/libretroshare/src/libretroshare.pro
@@ -553,7 +553,7 @@ SOURCES +=	ft/ftchunkmap.cc \
 			ft/ftfilesearch.cc \
 			ft/ftserver.cc \
 			ft/fttransfermodule.cc \
-			ft/ftturtlefiletransferitem.cc 
+                        ft/ftturtlefiletransferitem.cc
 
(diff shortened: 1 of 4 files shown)


G10h4ck:
Can't we do it without the struct pqinetworkOps with function pointers?
I believe we could use a plain class with static methods, or even a namespace or just the functions in a file without recurring to functions pointers that make code even more old C style and with a runtime pointer dereferencing at runtime that i believe it's not needed ;)

But yes i agree that having all those ifdef dispersed around is not a good idea, and your proposal is definitely better

sehraf:
Sorry for the late reply.

Yes, getting rid of the struct is easy.
The simplest (imho) solution is to just include the correct .c file (and use the same name for each  uni/windows functions).

PR #1086, opened 2017-10-29
Fix Speed/Queue Position column sorting

PhenomRetroShare:


felisucoibi:
+1 a good quality product pr, like it

csoler:
I like it too, as it fixes the sorting. Just one thing: because you use these constants (0.1, 0.2, etc) in multiple places, you'd better use a define or a static const float to name them, so that changing the values will not cause a bug that is hard to track.

PhenomRetroShare:
Done.

PR #1094, opened 2017-10-30
Hide zero File Label (size) in channel post.

PhenomRetroShare:


felisucoibi:
+1 i hope is soon merged a good point for channels future.

PR #1103, opened 2017-11-09
Fix gcc warnings

PhenomRetroShare:


libretroshare/src/util/rsdir.cc
```diff
@@ -24,31 +24,31 @@
...
-#include "util/rsdir.h"
-#include "util/rsstring.h"
-#include "util/rsrandom.h"
-#include "util/rstime.h"
-#include "util/rsmemory.h"
-#include "util/folderiterator.h"
-#include "retroshare/rstypes.h"
-#include "retroshare/rsnotify.h"
+#include "folderiterator.h"
+#include "rsdir.h"
```
csoler: why remove the util/ ? it makes the include line more confusing.

libresapi/src/api/FileSharingHandler.cpp
```diff
@@ -84,7 +84,9 @@ void FileSharingHandler::handleGetSharedDir(Request& /*req*/, Response& resp)
 	DirDetails dirDetails;
 	mRsFiles->RequestDirDetails(NULL, dirDetails, RS_FILE_HINTS_LOCAL);
 
+#pragma GCC diagnostic ignored "-Wstrict-aliasing"
 	resp.mDataStream << makeKeyValue("parent_reference", *reinterpret_cast<int*>(&dirDetails.ref));
```
csoler: The proper way to fix this is not to hide the warning, but to cast using intptr_t:

resp.mDataStream << makeKeyValue("...", (int*)(intptr_t)(dirDetails.ref));

libretroshare/src/gxstrans/p3gxstransitems.cc
```diff
@@ -48,7 +48,9 @@ void OutgoingRecord_deprecated::serial_process(RsGenericSerializer::SerializeJob
 	RS_REGISTER_SERIAL_MEMBER(recipient);
 	RS_REGISTER_SERIAL_MEMBER(mailItem);
 	RS_REGISTER_SERIAL_MEMBER(mailData);
+#pragma GCC diagnostic ignored "-Wstrict-aliasing"
 	RS_REGISTER_SERIAL_MEMBER_TYPED(clientService, uint16_t);
```
csoler: This warning tells something concerning. We should not hide it, but rather fix the wrong cast below.

libretroshare/src/util/rsdir.cc
PhenomRetroShare: Done

libresapi/src/api/FileSharingHandler.cpp
PhenomRetroShare: Done but use:
*reinterpret_cast<int*>(reinterpret_cast<intptr_t>(&dirDetails.ref))
Yours failed to compiles.

libretroshare/src/gxstrans/p3gxstransitems.cc
PhenomRetroShare: Done, only remove them, now no warnings???

G10h4ck: I would not concern much about those casts as PR #1155 introduce a new system that doesn't suffer of that kind of problems

PhenomRetroShare: Do you says that all of these code could be removed?

G10h4ck: no absolutely!! I say that those warning disappears when https://github.com/RetroShare/RetroShare/pull/1155 will be merged as i have found a better way to handle enums in serialization there

G10h4ck:
This PR has conflicts and stagnated, a discussion on how to make easier to review and merge warning fixing PRs has happened here https://github.com/RetroShare/RetroShare/pull/1382

Issue #191, opened 2015-12-07
In channels if you donwload a .rscollection "open" button won't open it, but open the folder where it is.


felisucoibi:
Updated: 01/2016

This happens in all versions, windows and linux.
 If you download from a channel a .rscollection when you click to Open, it opens the folder, in windows and linux, different versions, ONLY OPEN IT TO EXPLORE FILES INSIDE .rscollection IF YOU UPLOADED IT TO YOUR OWN CHANNEL.

At first I thought it was only in windows because I downloaded my .rscollection on linux, and on windows  was donwloaded by my friends, but not, if a friend with windows creates a .rscollection I can't see it, only opens the folder where the .rscollection is.

It works if you create the .rscollection and upload it to your channel, after that you can click on it and a popup appears, but not open it with others .rcollection files in channels

And another problem is that there is not a way in download queue to add a .rscollection, you have to go to your downloaded files and choose a directory and choose to open a .rscollection to add to this directory. SHOULD BE A WAY TO LOAD A .rscollection to download queue, in downloads queue window.

Thanks for your time.

felisucoibi:
I updated the bug, it fails also on linux if you don't create the .rscollection file, no popup offering you to download but opens the folder where the .rscolection is.

PR #1107, opened 2017-11-12
fixed bug causing RsCollection to fail on remote files. Now using RsC…

csoler:
…ollectionDialog when downloading directories from friend shared files


felisucoibi:
wow bug fixed! this bug is here since a few years ago.
https://github.com/RetroShare/RetroShare/issues/191

PR #1115, opened 2017-11-18
Add Copy action for Gxs Comments.

PhenomRetroShare:


felisucoibi:
+1

csoler:
The goal is right, but it seems that Qt should allow you to select some text and press ^C on it for copy, right? Isn't there a simple flag to change in the comment TreeView to allow that?

PhenomRetroShare:
That's because comments are only painted.
No way to select text.

PR #1140, opened 2017-12-27
Added to libresapi a few handy functions

kdebiec:


libresapi/src/api/FileSharingHandler.cpp
```diff
@@ -40,7 +42,30 @@ FileSharingHandler::FileSharingHandler(StateTokenServer *sts, RsFiles *files):
...
+FileSharingHandler::~FileSharingHandler()
+{
+	mNotify->unregisterNotifyClient(this);
+}
+
+void FileSharingHandler::notifyListChange(int list, int /* type */)
+{
+	if(list == NOTIFY_LIST_DIRLIST_LOCAL)
+	{
+		RsStackMutex stack(mMtx); /********** STACK LOCKED MTX ******/
```
G10h4ck: You should use RS_STACK_MUTEX(mMtx) macro instead, also the comment is unnecessary

libresapi/src/api/FileSharingHandler.cpp
```diff
@@ -40,7 +42,30 @@ FileSharingHandler::FileSharingHandler(StateTokenServer *sts, RsFiles *files):
...
+		mStateTokenServer->discardToken(mLocalDirStateToken);
+		mLocalDirStateToken = mStateTokenServer->getNewToken();
+	}
+	else if(list == NOTIFY_LIST_DIRLIST_FRIENDS)
+	{
+		RsStackMutex stack(mMtx); /********** STACK LOCKED MTX ******/
```
G10h4ck: You should use RS_STACK_MUTEX(mMtx) macro instead, also the comment is unnecessary

libresapi/src/api/FileSharingHandler.h
```diff
@@ -48,10 +55,13 @@ class FileSharingHandler: public ResourceRouter
 
 	void handleDownload(Request& req, Response& resp);
 
-	StateToken mStateToken;
+	StateToken mLocalDirStateToken;
+	StateToken mRemoteDirStateToken;
 	StateTokenServer* mStateTokenServer;
+	RsMutex mMtx;
```
G10h4ck: Here comments explaining the usage of each token and the mutex usage would be helful

libresapi/src/api/FileSharingHandler.h
```diff
@@ -48,10 +55,13 @@ class FileSharingHandler: public ResourceRouter
...
 
 	RsFiles* mRsFiles;
+	RsNotify* mNotify;
```
G10h4ck: It is preferable to use a reference instead of a pointer if possible (aka. if it is always valid and doesn't change between initialization and destruction)

libresapi/src/api/TransfersHandler.cpp
```diff
@@ -6,20 +6,33 @@
...
 
 TransfersHandler::~TransfersHandler()
 {
     mStateTokenServer->unregisterTickClient(this);
+	mNotify->unregisterNotifyClient(this);
+}
+
+void TransfersHandler::notifyListChange(int list, int /* type */)
+{
+	//RsStackMutex stack(mMtx); /********** STACK LOCKED MTX ******/
```
G10h4ck: If the mutex locking is not needed this line of potential deadcode should be removed before merging

libresapi/src/api/TransfersHandler.h
```diff
@@ -26,6 +28,7 @@ class TransfersHandler: public ResourceRouter, Tickable
     StateTokenServer* mStateTokenServer;
     RsFiles* mFiles;
 	RsPeers* mRsPeers;
+	RsNotify* mNotify;
```
G10h4ck: It is preferable to use a reference instead of a pointer if possible (aka. if it is always valid and doesn't change between initialization and destruction)

G10h4ck (changes requested):

libresapi/src/api/TransfersHandler.h
kdebiec: Pointer is internally used to access RsNotify object, hence there is no need to dereference it adding just another layer of complexity without any reasonable advantage

G10h4ck: I don't suggest to dereference it, I suggest to use a reference like +RsNotify& mNotify;+ instead of a pointer, it doesn't add complexity nor copies but has better guaranties (it can't be NULL or change in it's lifetime)

libresapi/src/api/FileSharingHandler.h
```diff
@@ -22,14 +22,21 @@
...
 
-class FileSharingHandler: public ResourceRouter
+class FileSharingHandler: public ResourceRouter, NotifyClient
 {
 public:
-	FileSharingHandler(StateTokenServer* sts, RsFiles* files);
+	FileSharingHandler(StateTokenServer* sts, RsFiles* files, RsNotify* notify);
+	~FileSharingHandler();
+
+	// from NotifyClient
```
G10h4ck: Please use Doxigen style comments for headers documentation like https://github.com/RetroShare/RetroShare/blob/master/libretroshare/src/gxstrans/p3gxstrans.h#L210

libresapi/src/api/FileSharingHandler.h
```diff
@@ -48,10 +55,13 @@ class FileSharingHandler: public ResourceRouter
 
 	void handleDownload(Request& req, Response& resp);
 
-	StateToken mStateToken;
+	StateToken mLocalDirStateToken; // Token indicating change in local shared files
```
G10h4ck: Please use Doxigen style comments for headers documentation like https://github.com/RetroShare/RetroShare/blob/master/libretroshare/src/gxstrans/p3gxstrans.h#L210

libresapi/src/api/TransfersHandler.h
```diff
@@ -5,16 +5,20 @@
...
 
 namespace resource_api
 {
 
-class TransfersHandler: public ResourceRouter, Tickable
+class TransfersHandler: public ResourceRouter, Tickable, NotifyClient
 {
 public:
-	TransfersHandler(StateTokenServer* sts, RsFiles* files, RsPeers *peers);
+	TransfersHandler(StateTokenServer* sts, RsFiles* files, RsPeers *peers, RsNotify* notify);
```
G10h4ck: Every time ti is possible it is better to use a reference instead of a pointer, in this case i would prefer if you use a reference at least for notify +RsNotify& notify+, it seems possible to use references for the other parameters too but use a reference instead of pointers when possible at least on new code

libresapi/src/api/TransfersHandler.h
```diff
@@ -5,16 +5,20 @@
...
     virtual ~TransfersHandler();
 
+	// from NotifyClient
```
G10h4ck: Please use Doxigen style comments for headers documentation like https://github.com/RetroShare/RetroShare/blob/master/libretroshare/src/gxstrans/p3gxstrans.h#L210

libresapi/src/api/TransfersHandler.h
```diff
@@ -26,7 +30,9 @@ class TransfersHandler: public ResourceRouter, Tickable
     StateTokenServer* mStateTokenServer;
     RsFiles* mFiles;
 	RsPeers* mRsPeers;
+	RsNotify* mNotify;
 
+	RsMutex mMtx;
```
G10h4ck: Please add some Doxygen documentation to explain the purpose of the mutex

G10h4ck (changes requested):

G10h4ck (approved):
Thanks you for great contribution!

PR #1144, opened 2018-01-05
Disable Yes button on chat room invitation if no Id selected.

PhenomRetroShare:


csoler:
well done, except for 1 bug: if you receive an invitation but you already have IDs, one will be selected but the button Yes is still Off until you select another ID.

PhenomRetroShare:
I don't have this bug...
And this is done by /retroshare-gui/src/gui/ChatLobbyWidget.cpp:1123

PhenomRetroShare:
But I cannot connect to room because:
(EE) Cannot subscribe a non visible chat lobby!!
/libretroshare/src/chat/distributedchat.cc:1410

PhenomRetroShare:
OK, Found and fixed.

PR #1147, opened 2018-01-08
gxsforums: not set hidden thread_read icon in thread widget

chelovechishko:


csoler:
I dont understand the goal of this PR: if the read status column is hidden, the icon still needs to be updated, so that when the column is displayed again, the icons will still be consistent with the read status. Did I miss something?

chelovechishko:
Oh, i thought that they are always recalculated [if they are needed again]. I just always keep them hidden.

PR #1153, opened 2018-01-17
V0.6 tor only

csoler:
This PR adds an option to RS to build a Tor version of RS which runs and configures Tor automatically at start. The running RS will only accept/create hidden nodes. GUI has been modified to not run/display unnecessary components such as DHT, IP filters, udp, etc.

To compile:
> qmake CONFIG="retrotor debug"
> make

To run:
- make sure you have a Tor executable in $path
- run RS normally.

Bridges are not supported yet, so if your ISP forbids Tor usage, RS-Tor will not start.


retroshare.pri
```diff
@@ -3,6 +3,12 @@
 CONFIG *= retroshare_gui
 no_retroshare_gui:CONFIG -= retroshare_gui
 
+# To build the RetroTor executable, just uncomment the following option.
+# RetroTor is a version of RS that automatically configures Tor for its own usage
+# using only hidden nodes. It will not start if Tor is not working.
+
+# CONFIG *= retrotor
```
G10h4ck, line 10: Could you reformat this in the same way as other compilation options?
The benefit it the other format (like for example retroshare_gui) is that independently from the default value of the option one can either force enable or force disable that option from the qmake invocation without editing the files

retroshare.pri
```diff
@@ -230,6 +236,10 @@ rs_autologin {
     warning("You have enabled RetroShare auto-login, this is discouraged. The usage of auto-login on some linux distributions may allow someone having access to your session to steal the SSL keys of your node location and therefore compromise your security")
 }
 
+retrotor {
+    DEFINES *= RS_ONLYHIDDENNODE
```
G10h4ck: It may be better to just use CONFIG *= rs_onlyhiddennode at this point

retroshare-gui/src/TorControl/TorControl.cpp
```diff
@@ -0,0 +1,786 @@
...
+#include "PendingOperation.h"
+#include <QHostAddress>
+#include <QDir>
+#include <QNetworkProxy>
+//#include <QQmlEngine>
+#include <QTimer>
+#include <QSaveFile>
+#include <QRegularExpression>
+#include <QDebug>
+
```
hunbernd, line 54: #include <time.h>

G10h4ck (changes requested):

PR #1157, opened 2018-01-20
Fix icon cache in files defs

PhenomRetroShare:


felisucoibi:
is possible that after this pr rs gets slow with big filelists, my rs is getting super slow takes like 20 minutes to starts from one week ago. I have to delete local_dir_ etc file in filelists to be able to lauch rs fast....

csoler:
it's entirely possible. At start, RS performs a "search" because the UI is refreshed. This is not really needed, but simply a side effect of the way it's currently implemented. It could be the cause of the slow down. Still, 20 mins hang up sounds pretty crazy.

felisucoibi:
files is hanging more than usual, and yes 20 minutes if i don't delete the /filelists/local_dir_hicheracy file... and is hanging more than ever the gui, I have 1000 files in download queue and takes 2 3 minutes to escape from the hang....when i click FILES button, i will try to revert this pr in my fork to test

felisucoibi:
Testing, maybe is not this PR but the filelists changes you did, testing the other pr.

PhenomRetroShare:
Hello, did you test to revert this PR?

felisucoibi:
Yes, is ok the PR, rs was hanging because i had 1000 files in download queue and made everything hangs continously, not because of this PR.

PR #1159, opened 2018-01-23
V0.6 file lists

csoler:
For testing purposes. Seems to work much more efficiently than the previous implementation, but needs to be tested with large file lists.

### DIFF
--- a/libretroshare/src/file_sharing/p3filelists.cc
+++ b/libretroshare/src/file_sharing/p3filelists.cc
@@ -1155,6 +1155,8 @@ int p3FileDatabase::SearchKeywords(const std::list<std::string>& keywords, std::
     if(flags & RS_FILE_HINTS_LOCAL)
     {
         std::list<EntryIndex> firesults;
+        std::list<void *> pointers;
+
         {
             RS_STACK_MUTEX(mFLSMtx) ;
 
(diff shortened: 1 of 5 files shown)


csoler:
There's a bug however that I cannot manage to get rid of:when the search is empty, RS expands all items. This takes a long time. This should be avoided, and my code is  normally supposed to do that but it doesn't and I dont understand why.

[EDIT] FIXED. Apparently, QString::isNull() is not equivalent to QString()=="". Weird.

felisucoibi:
I ll test it today

felisucoibi:
Tested, works well with huge file lists, in fact it does not hang the gui, and it just takes 5 secs to crawl my whole friends file lists... very good!, 
BUT is not usefull at all, because it show lot of files and directories that have no relation with the search result, it should HIDE all the directories, and all the files of the friend tree of files that are not related to the search result.
Example, a directory with 1000 files and also 500 directories inside, and one of the files is called, "champions of the world" if you search for "champions of the world" you will see a directory with 1000 files 500 directories, and you will not even be able to see which of these files contains "champions of the world", ones does but is impossible to find it, and also it shows other directories inside this 1000 files directory if exists... so it should hide all the directories without "champions of the world" file inside, and all the files in the same directory without "champions of the world" string.

felisucoibi:
Thanks for all this work here, super useful.

In friends filelists there is a green icon a triangle, and when you search for something ther eis a red cross, in My files is not this icon, so after searching you cant remove the search clicking in the red icon and you can see your filelist without a search result.


Btw in friend filelists, could be useful to remove the users that a) use old rs versions or b) don't share anything with you, now they are there, and they just use space.

And also could be very useful a column for total files shared, and another for total TB shared, so column shorting can work, now, is mixed in one column, and sorting is not working.

PR #1183, opened 2018-02-19
switched version to 0.6.4

csoler:


elthommy:
So the time has come? 0.6.4 will be officially released soon :-D  ^__^

PR #1184, opened 2018-02-19
[0.6.4] Use link local address only if necessary fix #292

### LINKED ISSUE #292: local ip address detection in latest build 0.6.0 Revision 7ded128b not showing right address
In Windows QT5 RetroShare Version: 0.6.0 Revision 7ded128b the local ip address 
seems to be showing wrong address (169.245.x.x) although my local address is in
the 192.168.x.x address space.

checking ipconfig /all for that 169er network does not yield any adapter using
that address space.

the older version qt4 0.6.0x rev 8551 just works fine showing the right address, will
check the qt4 version of the latest version in a sec.
....
Version: 0.6.0 Revision 7ded128b Windows QT 4.8.7 also shows the 169er network.

and i'm using portable install, if that is of any concern.

G10h4ck:
This should solve APIPA address being always selected as preferred local
address on Windows which causes problems connecting to machines on
the same local network.
Now the link local address is used as fallback only if there is non link
local private address available.

### DIFF
--- a/libretroshare/src/pqi/p3netmgr.cc
+++ b/libretroshare/src/pqi/p3netmgr.cc
@@ -3,7 +3,8 @@
  *
  * 3P/PQI network interface for RetroShare.
  *
- * Copyright 2007-2011 by Robert Fernie.
+ * Copyright (C) 2007-2011  Robert Fernie
+ * Copyright (C) 2015-2018  Gioacchino Mazzurco <gio@eigenlab.org>
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Library General Public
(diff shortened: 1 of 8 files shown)


G10h4ck:
this should fix https://github.com/RetroShare/RetroShare/issues/292

csoler:
Could you please replace the "for(auto&& addrs: something)" by a for(auto it(something.begin();it!=something.end();++it), for backward compatibility.

csoler:
what do you mean by "Having just one local address is a flawed assumption" ? getLocalAddresses() mainly retrieves a network interface, which comes with its local address. And RS only talks to one network interface.

PR #1190, opened 2018-02-24
Add ShowEmpty sub menu in Tree Remote SharedFilesDialog.

PhenomRetroShare:


csoler:
There's a few bugs:
- headers sometimes show "Column -1"
- when unsetting "show empty" some lines become empty but the directories are not packed upward. This is not very nice. I suppose that calling a PostMod() or update function after telling the RemoteDirModel that things have changed, it needed

PhenomRetroShare:
Updated.

felisucoibi:
+1 waiting for this pr

felisucoibi:
I hope this pr is ready for 0.6.4 son also flat mode can be updated to hide older than x days files, when sorting.

PR #1191, opened 2018-02-24
IPv6 support

G10h4ck:
Attempt to implement IPv6 dual stack operation in a retro-compatible non chain reaction manner

### DIFF
--- a/libretroshare/src/dht/p3bitdht_peers.cc
+++ b/libretroshare/src/dht/p3bitdht_peers.cc
@@ -4,6 +4,7 @@
  * BitDht interface for RetroShare.
  *
  * Copyright 2009-2010 by Robert Fernie.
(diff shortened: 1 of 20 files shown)


sehraf:
Great to see that IPv6 support is still alive :smile: 
When this is in an usable state i can help testing (i have native IPv4 and IPv6).

sehraf:
Since 0.6.4 is out, can this be merged now? :smile_cat:

G10h4ck:
For the tests I have made this is ready to go

tangdou1:
Does it supports DHT in ipv6？ When would the 0.6.5 including ipv6 be  released？ Thanks.

G10h4ck:
No libbitdht doesn't support IPv6 and we don't plan to add IPv6 support to libbitdht ATM, we have been considering to switch to an independently maintained DHT library that supports IPv6 already. Next release will include the IPv6 support already included in master branch.

PR #1192, opened 2018-02-24
Add a warning in ToolTip if Max files reach in flat view mode Dir Tree.

PhenomRetroShare:


felisucoibi:
it could be nice in flat mode if you order by modification, to only list modified files not the rest in grey like now that limits how many files you can see in flat mode.

PhenomRetroShare:
I'll do that but need PR #1190 merged.

felisucoibi:
So +1(1+1+1 to this pr

PR #1207, opened 2018-03-04
Add a timer in SharedFilesDialog filter to not so often triggs search in

PhenomRetroShare:
files.


felisucoibi:
+1+1+1 maybe this fix the search problemas and haging? testing

felisucoibi:
after testing still hangs... is like on every key pressed it makes a search but it should only when play button clicked, it should not make a search or a test continously

PhenomRetroShare:
This search is only to know how many results and update play button.
So, you ask to not update it and only know if too much results when click on play?

felisucoibi:
Yes only when play clicked i think is even working after play in background again and again.

felisucoibi:
i cant test because has conflicts with actual branch :(

csoler:
whenever you use a timer, you need to check for RsAutoUpdatePage::eventsLocked() in order to avoid deadlocks when asking for passwords. In the current case it's probably better to use a RsProtectedTimer instead.

PhenomRetroShare:
Updated.

PR #1210, opened 2018-03-04
Add Background to xprogressbar text for more readability.

PhenomRetroShare:


csoler:
I suppose you can improve this a little bit: use a gradient instead of a hard edge. That will also allow to extend the grey background a little bit away from the border of the text. Is that possible to do?

PhenomRetroShare:
Done.

PR #1214, opened 2018-03-15
enabled aggressive re-request of pending slices at end of transfer, t…

csoler:
…hus fixing the long delay to finish files with mixed fast/slow sources. To be tested.


felisucoibi:
seems doing it's job, not more almost unfinished files in my queues.

PR #1227, opened 2018-04-05
bring back open local file on link click

RetroPooh:


csoler:
Why such as large diff? I have difficulties to see what's changed. It shouldn't a lot given the summary.

RetroPooh:
It's because of reordered switch with one case block moved up the queue

PR #1231, opened 2018-04-07
add online filter to NetworkView

RetroPooh:


csoler:
Could you plz remove the "show" from the first 2 menu entries so they become more consistent?

G10h4ck:
@csoler seems @RetroPooh applied requested changes, any more comments or should I merge it right away?

PR #1236, opened 2018-04-22
V0.6 file lists

csoler:
This PR should fix the problems related to searching. Because the new method is much faster, I re-enabled the search while typing. I need some testers to tell me if that is ok or not.


felisucoibi:
the realtime search completely hangs rs, i liked the phenom fix for that, why is necessary realtime search on every key pressed? I mean it's fancy but with lot of users and filelists it hangs....

PR #1244, opened 2018-05-02
spamprotection

RetroPooh:


csoler:
can you plz explain a bit what you did? thx

RetroPooh:
Normal users send join and keepalive packets, spammer usually doesn't - the bad message comes first, our client does not even know he participates. So this way we try to differentiate them.
Sure spamer will still have the way to go past this block, but it will work for some extent

PR #1245, opened 2018-05-02
Add Context Menu for GxsId in lobby textBrowser.

PhenomRetroShare:


csoler:
Why do you need the GxsIdForAction variable?
It's not so nice to do it that way. I just want to make sure there's no cleaner way to do that
Normally you can pass data to an action raised by a menu
See for install how we do context menus in chat rooms when entering a room with a specific id.
One example is ChatLobbyDialog constructor. Look for the votePositiveAct->setData() call. Then the data is retrieved in the voteParticipant() method using act->data(). I think you can use the same system to pass the GxsId(s) from the anchor.

PhenomRetroShare:
Done

csoler:
Well done! thx.

PR #1249, opened 2018-05-04
Add support for libsecret (including KDE keyring) and fix retroshare-nogui autologin

sehraf:
As @PhenomRetroShare  pointed out in 5f12b6076dd29e0353c7c6b9788ea73767f56afb libgnome-keyring is deprecated by the gnome project and its successor is libsecret.

While libgnome-keyring was only designed to interact with gnome keyring, libsecret can use any back-end that implements the "Secret Service" using D-Bus. (These are gnome-keyring and ksecretservice). So this PR effectively adds support for KDEs keyring, too.

The first commit fixes 5f12b6076dd29e0353c7c6b9788ea73767f56afb. retroshare.pri takes care of everything so nothing needs to be set manually in retroshare-nogui.pro except that Qts library system must be enabled.

The second commit adds support for libsecret which is very similar to libgnome-keyring.

The third commit adds an automatic switch that will select libsecret when available and fall back to libgnome-keyring otherwise when rs_autologin is set.

This was tested on Arch Linux.

### DIFF
--- a/libretroshare/src/rsserver/rsloginhandler.cc
+++ b/libretroshare/src/rsserver/rsloginhandler.cc
@@ -109,6 +109,19 @@ GnomeKeyringPasswordSchema my_schema = {
     NULL,
     NULL
 };
+#elif defined(HAS_LIBSECRET)
+#include <libsecret-1/libsecret/secret.h>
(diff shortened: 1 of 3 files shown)


csoler:
very nice!

PR #1254, opened 2018-05-08
fix gcc8

sehraf:
As far as i can tell `std::set<TurtleVirtualPeerId, RsGRouterTransactionChunkItem*>` doesn't make any sense.
Compile and (quickly) run tested on Arch with gcc-Version 8.1.0 (GCC)


G10h4ck:
yeah it make no sense to me either, don't know how it end up there and why didn't complain at first

sehraf:
fixed by 4d287d68bc9725f403dc7d952a927d401c5d6c97

PR #1255, opened 2018-05-08
Add Ignore Text list in SearchDialog

PhenomRetroShare:
Ability to modify the saved list when long click on button "Ignore".
All in tooltip.


felisucoibi:
what is that for?

PhenomRetroShare:
This is to ignore some result if they contains some text like 'porn' or other as you want.

felisucoibi:
good idea

PhenomRetroShare:
@csoler ;)

PR #1272, opened 2018-05-22
Net little fixes

G10h4ck:


libretroshare/src/pqi/p3peermgr.cc
```diff
@@ -1239,11 +1239,11 @@ bool p3PeerMgrIMPL::UpdateOwnAddress( const sockaddr_storage& pLocalAddr,
 	sockaddr_storage_copy(pExtAddr, extAddr);
 	sockaddr_storage_ipv6_to_ipv4(extAddr);
 
-#ifdef PEER_DEBUG
+//#ifdef PEER_DEBUG
```
sehraf, line 1242: i think you want to revert these changes :wink:

G10h4ck: nope, those messages are not printed often, and i need them because some user has recently reported that their RS miss some IP change, with this message enabled the console output will be helpful for that case

PR #1284, opened 2018-06-11
Improve build and Windows Continuos Integration

G10h4ck:


G10h4ck:
@thunder2 with this PR the appveyor CI produce retroshare executable and a zip archive with the needed libraries but it still crash after few seconds of running, would you be so kind to take a look at it so we could have testbuild for all pull requests compiled and testable on windows too?

hunbernd:
Not crashes on Win 10 x64

PR #1288, opened 2018-06-21
changed std::vector into std::set in notification system, hence remov…

csoler:
…ing lots of std::find(std::vector::begin(),std::vector::end(),T), causing quadratic costs in multiple places.

This needs to be tested systematically. In particular:
- check that new messages are properly added to GUI in forums and that notifications work normally.
- nothing unexpected happens (beyond already known bugs)

This PR should overall improve the speed of GXS UI, especially identities and forums.


felisucoibi:
Seems forum loading is very fast now, channels are not as fast.

PR #1289, opened 2018-06-23
Implement automatic JSON API generation

G10h4ck:
qmake file add jsonapi-generator target to compile JSON API generator
qmake files add rs_jsonapi CONFIG option to enable/disable JSON API at compile
  time
RsTypeSerializer pass down same serialization flags when creating new context
  for nested objects serial job
RsGxsChannels expose a few methods through JSON API as example
Derive a few GXS types (RsGxsChannelGroup, RsGxsChannelPost, RsGxsFile,
  RsMsgMetaData) from RsSerializables so they can be used for the JSON API
Create RsGenericSerializer::SERIALIZATION_FLAG_YIELDING so JSON objects that
  miss some fields can be still deserialized, this improve API usability
SerializeContext offer friendly constructor with default paramethers
Add restbed 4.6 library as git submodule as most systems doesn't have it yet
Add a bit of documentation about JSON API into jsonapi-generator/README.adoc
Add JsonApiServer class to expose the JSON API via HTTP protocol


sehraf:
maybe b7f5d4286f166d25dbf812e9ec2826a96ebdf10c should be cherry-picked by @csoler

G10h4ck:
I think we should now considering this for merge

sehraf:
Is the restbed issue solved?
An automatic detection of a system installed version would be nice, too

sehraf:
I'd like to emphasize that this pr reduces (de)serialization of anything to a few lines :astonished:

hunbernd:
Additionally generate OpenAPI/Swagger definitions.
It makes possible to generate client side code for various languages.

jsonapi-generator/README.adoc
```diff
@@ -0,0 +1,287 @@
+RetroShare JSON API
+===================
+
+:Cxx: C&#43;&#43;
+
+== How to use RetroShare JSON API
+
+Look for methods marked with +@jsonapi+ doxygen custom command into
++libretroshare/src/retroshare+. The method path is composed by service instance
```
csoler, line 9: Plz cite one particular method here by pasting its header from libretroshare. Maybe rsgxschannels.h?

jsonapi-generator/README.adoc
```diff
@@ -0,0 +1,287 @@
...
+the API because almost all the information needed is already present in the
+C++ API +libretroshare/src/retroshare+.
+
+So a third step to improve the JSON API would be to remove this source of
+boilerplate code by automatizing the mapping between C++ and JSON API call.
+
+This may result a little tricky as language parsing or other adevanced things
+may be required.
+
+Hope this dive is useful for you +
```
csoler, line 289: you may want to remove these 3 lines ;-)

libretroshare/src/retroshare/rsnotify.h
```diff
@@ -206,44 +206,42 @@ class RsNotify
 		virtual bool setDisableAskPassword (const bool /*bValue*/) { return false ; }
 };
 
-class NotifyClient
+struct NotifyClient
```
csoler: Because this struct as some methods you might want to keep it as a class. As we discussed already, we should keep "struct" for pure data structures without functionality.

libretroshare/src/rsserver/rsinit.cc
```diff
@@ -1929,3 +1934,88 @@ int RsServer::StartupRetroShare()
...
+	if(!rsNotify->cachePgpPassphrase(password)) return RsInit::ERR_UNKOWN;
+	if(!rsNotify->setDisableAskPassword(true)) return RsInit::ERR_UNKOWN;
+	if(!RsAccounts::SelectAccount(account)) return RsInit::ERR_UNKOWN;
+	std::string ignore;
+	RsInit::LoadCertificateStatus ret =
+	        RsInit::LockAndLoadCertificates(false, ignore);
+	rsNotify->setDisableAskPassword(false);
+	if(ret != RsInit::OK) return ret;
+	if(RsControl::instance()->StartupRetroShare() == 1) return RsInit::OK;
+	return RsInit::ERR_UNKOWN;
```
csoler: It seems to me we should clear the cached password here right after LockAndLoadCertificates(), since it was passed as a parameter.

libretroshare/src/serialiser/rsserializer.h
```diff
@@ -226,7 +220,7 @@ struct RsGenericSerializer : RsSerialType
 		/** Allow shared allocator usage to avoid costly JSON deepcopy for
 		 *  nested RsSerializable */
 		SerializeContext(
-		        uint8_t *data, uint32_t size,
+		        uint8_t* data = nullptr, uint32_t size = 0,
```
csoler, line 221: make sure that if only data is passed to SerializeContext, an error is printed out, otherwise you end up with some memory and no size for it.

libretroshare/src/serialiser/rsserializer.h
```diff
@@ -261,6 +255,12 @@ struct RsGenericSerializer : RsSerialType
 	static const SerializationFlags SERIALIZATION_FLAG_SIGNATURE;      // 0x0002
 	static const SerializationFlags SERIALIZATION_FLAG_SKIP_HEADER;    // 0x0004
 
+	/** Used for JSON deserialization in JSON API, it causes the deserialization
+	 * to continue even if some field is missing (or incorrect), this way the
+	 * API is more user friendly as some methods need just part of the structs
+	 * they take as parameters. */
+	static const SerializationFlags SERIALIZATION_FLAG_YIELDING;       // 0x0008
```
csoler, line 258: This should obviously be restricted to JSon only. Using this in the binary serialization is very risky. Can you somewhere check this for consistency?

libretroshare/src/serialiser/rstypeserializer.cc
```diff
@@ -328,6 +335,49 @@ bool RsTypeSerializer::from_JSON( const std::string& memberName,
 	return ret;
 }
 
+template<> /*static*/
+uint32_t RsTypeSerializer::serial_size(const double&)
```
csoler, line 339: why is this needed? Either we try to serialize doubles, then we need this, or we don't, then we dont need this either. Why would JSon need doubles where the binary API doesn't? Also if necessary, serializing doubles is quite possible. I did it for float32 already I think.

libretroshare/src/serialiser/rstypeserializer.h
```diff
@@ -643,28 +732,32 @@ struct RsTypeSerializer
...
 
-			RsGenericSerializer::SerializeContext lCtx(
-			            nullptr, 0, RsGenericSerializer::FORMAT_BINARY,
-			            RsGenericSerializer::SERIALIZATION_FLAG_NONE );
+			RsGenericSerializer::SerializeContext lCtx(nullptr, 0, ctx.mFlags);
 			lCtx.mJson.SetObject() = v; // Beware of move semantic!!
 
 			member.serial_process(j, lCtx);
-			ctx.mOk = ctx.mOk && lCtx.mOk;
+			ctx.mOk &= lCtx.mOk;
```
csoler, line 760: Strictly speaking this is dangerous, unless you make sure by design that ctx.mOk is always 0 or 1, which might depend on the definition of true and false and on the particular implementation and on the value you put into ctx.mOk (some function return integers).  !4 is false, !2 is false as well, but !(4 & 2) is true. Normally you'd need &&= but I'm not sure this exists.

retroshare-gui/src/gui/gxschannels/CreateGxsChannelMsg.cpp
```diff
@@ -449,11 +450,10 @@ void CreateGxsChannelMsg::addAttachment(const std::string &path)
 	}
 
 	FileInfo fInfo;
-	std::string filename;
+	std::string filename = RsDirUtil::getTopDir(path);
```
csoler, line 453: getTopDir() returns the file name??? Either a typo, or the name of the function is really not well chosen.

jsonapi-generator/README.adoc
G10h4ck: Done ;)
Thanks!

jsonapi-generator/README.adoc
G10h4ck: Well that is left there just as historical record on how this idea came out ;)

libretroshare/src/retroshare/rsnotify.h
G10h4ck: Done! Thanks!

libretroshare/src/rsserver/rsinit.cc
G10h4ck: Done! Thanks!

libretroshare/src/serialiser/rsserializer.h
G10h4ck: Done! Thanks!

libretroshare/src/serialiser/rsserializer.h
G10h4ck: Done! Now serialization context check for usage of this flag on binary contexts and print an error and stacktrace if that happens

Thanks!

libretroshare/src/serialiser/rstypeserializer.cc
G10h4ck: We need this for JSON because some parameter type of some method exposed in the JSON API use double, binary serialization is not needed ATM because that kind of type is never exchanged between retroshare instances, I don't know the math behind floating point binary serialization so i haven't implemented it as is not needed ATM, but if you want to contribute that you are welcome to do that ;)

libretroshare/src/serialiser/rstypeserializer.h
G10h4ck: Because ctx.mOk is a member of type bool in C++ (not in C) it is guaranted to be either true or false, so is is safe in that case, but i agree that the && variant is usually safer, and sadly &&= doesn't exists (we would benefit of that in other places especially because && is short circuit while & is not)

retroshare-gui/src/gui/gxschannels/CreateGxsChannelMsg.cpp
G10h4ck: At first glance it seems to me that that function just split the path by / and return the last piece, so in case of a file path it is a file name, in case it is a dir path is the last dir name, but i agree that the function name is at least misleading...

PR #1291, opened 2018-07-02
Share extra locators via certificate

G10h4ck:


dtschmitz:
@csoler   This (RsUrl) implementation is too aggressive; existing cert structure is extensible and unknown fields are simply ignored.  Not 100% tamper-proof but certainly hardened.  The above is an invitation to a vast myriad of url crackers who will certainly exploit this opportunity.  I give this a BIG THUMBS DOWN.

PR #1294, opened 2018-07-05
V0.6 gxs dist sync3 turtle search

csoler:
This PR allows distant synchronization of GXS through tunnels. It comes with major modifications:
- GXS services now have a "searchable" flag that make them respond to distant search and tunnel requests.
- A new service (rsgxsnettunnel.h) provides tunnels that give "virtual peers" to GXS, which sync with these peers as with normal friends, using encrypted communications. 
- GxsId service also transparently uses these tunnels to retrieve/request the ID that are needed to check signatures.
- search use a new capability of turtle to perform generic search by client services (before only file search was supported using an explicit method).
- two search methods are implemented: one that searches group IDs (+some info) from keywords, and one that requests the full group data. The later is significantly larger because of images and keys, so it needs to be more focused. That is why I splitted the search in these two parts. Consequently search results can not directly be subscribed and need a manual user action to "request the group data" for which I'm still hesitating about a proper text to use ("add to my channels" ? "request channel data" ?)
- sync using tunnels is automatic. If no direct friend currently supplied a group, tunnels will be requested for it, etc. 
- GxsGroupFrameDialog now provides a search field in the bottom (see in channels) that allows to start distant search from a single keyword.
 
Privacy: Tunnels are not shared between services nor groups, so that it is not possible to associate virtual peers with sets of groups. Virtual peers have a random identifier that is a hash of the GroupId and a random 20 bytes bias, that is kept across restarts. This random identifier is what the GxsMaps (server/client) will use so it needs to be rather stable in time. Tunnels are opened based on a group hash, which prevents man in the middle spying of GXS sync data from anyone who doesn't know the group ID. The whole sync system is totally anonymous (as much as file transfer is).

See rsgxsnettunnel.h for a technical documentation. Most of the changes work for all GXS services, but some (especially in GUI) need a specific implementation. For the time being, I did it for channels. In the future forums, circles and posted will also optionally sync like this.

The UI part has been implemented for channels only.
in terms of UI, the software allows to
- search for distant channels
- "request channel data"
- subscribe and sync the channel with anonymous distant peers.

What to test?
- test that turtle search/FT still works correctly for files. I'm made a few changes there.
- channels can be searched (see the search field down the channels list). Only channels that are currently not known (i.e. not currently  listed) will be reported by the search. 
- Once a channel is reported in the search, the drop contextual menu allows to "Request data" for the channel. This is done only once per click, so I cannot ensure that it will always work. This might be improved later. I first made it automatic, but it's not so nice either because it causes some significant data transfer so it should be kept minimal.
- when the channel data is obtained, the channel will go to either popular groups or other channels, and you'd be able to subscribe. I dont know what happens to unsubscribed channels. They will probably be removed after some time because no direct friend is supplying them.
- even if no friend supplies  a subscribed channel, it should sync using tunnels. Be careful: GXS only syncs after something new has been posted in groups (something to improve that is totally independent of distant sync), either as a server or a client.

Backward compatibility: search needs *all intermediate peers* to use the newest version. Sync only needs the two endpoints (I'm not 100% sure of this).

Next steps:
- extend search to general RegExp
- apply search to not only groupname but also group data and message data (already implemented by G10Hack)
- extend dist sync to forums+circles.

This is a very complex PR, as it both modifies GXS, turtle, GXS UI, and uses some crypto. I dont expect that it will work out of the box. So if testing, please note everything that goes fancy.


libretroshare/src/gxs/rsgxsnettunnel.cc
```diff
@@ -0,0 +1,1167 @@
...
+
+static const uint32_t RS_GXS_NET_TUNNEL_MAX_ALLOWED_HITS_GROUP_DATA    =   1;
+static const uint32_t RS_GXS_NET_TUNNEL_MAX_ALLOWED_HITS_GROUP_SEARCH  = 100;
+
+RsGxsNetTunnelService::RsGxsNetTunnelService(): mGxsNetTunnelMtx("GxsNetTunnel")
+{
+#warning this is for testing only. In the final version this needs to be initialized with some random content, saved and re-used for a while (e.g. 1 month)
+	mRandomBias.clear();
+
+	mLastKeepAlive = time(NULL) + (lrand48()%20);	// adds some variance in order to avoid doing all this tasks at once across services
```
hunbernd: Replace lrand48() with rand() to make it work on Win!

PR #1296, opened 2018-07-06
fix equal names

chelovechishko:


G10h4ck:
did you get an error without those changes?

chelovechishko:
I got ODR violation 6 months ago, now it is not needed for me anymore.

G10h4ck:
by now you mean after applying this PR, or with the current master branch?

chelovechishko:
With current master.

G10h4ck (approved):

G10h4ck:
I have reviewed your patch, it make sense that the unpatched code endup in ODR violations because we use same name for same thing... also the ZeroedInt could confict easly so good to put it into a namespace

PR #1302, opened 2018-07-16
Add auto detection of installed rapidjson

sehraf:
This small pr adds detection of installed rapidjson (version >= 1.1). 
Also removed an unnecessary include.


G10h4ck (approved):
Nice!

PR #999, opened 2017-08-19
Add CommonMark in ChatLobbyDialog

PhenomRetroShare:
Direct use of https://github.com/commonmark/cmark library.

PR #1304, opened 2018-07-23
WIP: auto detect and use system's cmark

sehraf:
won't compile right now (see #999 )
Whith this patch it compiles and works (i have no idea whether it leaks memory now or not)
https://gist.github.com/sehraf/e8f753844427d7a7551c14a84a84fa7e

### DIFF
--- a/retroshare-gui/src/util/HandleRichText.cpp
+++ b/retroshare-gui/src/util/HandleRichText.cpp
@@ -39,11 +39,14 @@
 #include "util/rstime.h"
 
 #ifdef USE_CMARK
-//Include for CMark
-// This needs to be fixed: use system library if available, etc.
-#include <gui/../../../supportlibs/cmark/src/cmark.h>
-#include <gui/../../../supportlibs/cmark/src/node.h>
-#endif
+  //Include for CMark
+  #ifdef HAS_CMARK
+    #include <cmark.h>
+  #else
+    #include <gui/../../../supportlibs/cmark/src/cmark.h>
+    #include <gui/../../../supportlibs/cmark/src/node.h>
+  #endif // HAS_CMARK
+#endif // USE_CMARK
 
 #include <iostream>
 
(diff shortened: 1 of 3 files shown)


G10h4ck:
I tought that the git submodule was downloaded automatically by qmake like i did with restbed, this could be fixed easly looking at how i madi it into the JSON API PR

retroshare-gui/src/util/HandleRichText.cpp
```diff
@@ -39,11 +39,14 @@
...
-//Include for CMark
-// This needs to be fixed: use system library if available, etc.
-#include <gui/../../../supportlibs/cmark/src/cmark.h>
-#include <gui/../../../supportlibs/cmark/src/node.h>
-#endif
+  //Include for CMark
+  #ifdef HAS_CMARK
+    #include <cmark.h>
+  #else
+    #include <gui/../../../supportlibs/cmark/src/cmark.h>
```
G10h4ck, line 46: This ugly include can be avoided by adding cmark headers path to INCLUDEPATH

G10h4ck:
If cmark is not availabe on the system it can be downloaded automatically in similar way i did with restbed for JSON API  https://github.com/RetroShare/RetroShare/pull/1289/files#diff-3a8abf9b3bc2917a3929a7fa760eceaeR882

retroshare-gui/src/util/HandleRichText.cpp
sehraf: it can also be avoided by not "hardcoding" headers into RS.

sehraf:
cmark is available (even as ghc-version). Only node.h is not supposed to be used externally.

PhenomRetroShare:
Did you found something?

PhenomRetroShare:
@sehraf : Any news on it?

sehraf:
`node.h` is still the problem as it is not an exposed header file.

PhenomRetroShare:
So we could use another exposed header file including it?

PR #1306, opened 2018-07-26
V0.6 tor only

csoler:
Makes TorOnly mode available in standard RS version. In UI, only certificate creation windows has changed. At login, TorManager is started or not depending on the type of node that is selected (Tor only, normal, manually configured Tor/I2P). Not to be merged yet. Only for testing!


felisucoibi:
testing

PR #1313, opened 2018-08-01
Fix RS_AUTOLOGIN Compilation

PhenomRetroShare:


PhenomRetroShare:
Already done by Cyril in same moment.

PR #1316, opened 2018-08-03
added a working procedure to compile on Windows using MSYS2

zapek:
In order to improve the Windows build for everyone, here's an updated documentation that I did by installing and testing all steps from scratch until I got a working executable.

### DIFF
--- a/WindowsMSys2_InstallGuide.md
+++ b/WindowsMSys2_InstallGuide.md
@@ -59,7 +59,7 @@ We will use it to checkout Retroshare and build it:
 Go to the RetroShare directory and configure to your liking, for example:
 	
 	cd RetroShare
-	qmake -r -Wall -spec win32-g++ "CONFIG+=debug" "DEFINES+=RS_AUTOLOGIN"
+	qmake -r -Wall -spec win32-g++ "CONFIG+=debug" "CONFIG+=rs_autologin"
 
 Now we're ready to build Retroshare. Use the '-j' option with the number of
 cores you have for a faster build, for instance if you have 4 cores:
--- a/WindowsMSys2_InstallGuide.md
+++ b/WindowsMSys2_InstallGuide.md
@@ -66,9 +66,6 @@ cores you have for a faster build, for instance if you have 4 cores:
 
 	mingw32-make -j4
 
-Note: as of 2018-08-03 the build might fail once in libres during the HTML generation.
-If you get it, just run mingw32-make again and it will pass.
-
 Make sure your current Retroshare is not running. Then just run:
 
 	retroshare-gui/src/debug/retroshare.exe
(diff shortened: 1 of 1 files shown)


csoler:
Nice work. @G10h4ck : does it fits your AppVeyor script?

WindowsMSys2_InstallGuide.md
```diff
@@ -1,88 +1,76 @@
...
-Restart console  
-
-       pacman -Su  
-
-Install all default programms  
+Download MSYS2 from [MSYS2](http://www.msys2.org/). Get the i686 version
+if you run a 32-bit Windows or the x86_64 if you run a 64-bit Windows.
 
-       pacman -S base-devel git mercurial cvs wget p7zip gcc perl ruby python2  
+Run the installer and install MSYS2.
```
hunbernd: New versions of msys2 works only on NTFS pertitions. 
So do not install it on SSD, the installation will fail with some unrelated error.

zapek: SSD? You surely mean an external USB stick with FAT32 or exFAT? I don't think anybody would use those other than for storing files anyway.

hunbernd: My existing msys2 installation didn't work after update because of missing files.
The installer stopped because it couldn't copy some OpenSSL related file.
It failed because my SSD are formatted as exFat, and typically other's SSDs too.

From the official site:
Installation restrictions:
    MSYS2 can't be installed on FAT* partitions.
    Current MSYS2 can't be installed on Windows XP anymore

hunbernd: I just check my SSDs filesystem and it is NTFS, then I don't have an idea why it is failed on my SSD, but not on my hard drive.

WindowsMSys2_InstallGuide.md
```diff
@@ -1,88 +1,73 @@
...
+We will use it to checkout Retroshare and build it:
 
+	git clone https://github.com/RetroShare/RetroShare.git
 
-Open an MSys2 32|64 shell  
-Move to build_scripts:  
+Go to the RetroShare directory and configure to your liking, for example:
+	
+	cd RetroShare
+	qmake -r -Wall -spec win32-g++ "CONFIG+=debug" "DEFINES+=RS_AUTOLOGIN"
```
G10h4ck: You should change `"DEFINES+=RS_AUTOLOGIN"` into `"CONFIG+=rs_autologin"` as you can see into retroshare.pri file

G10h4ck (changes requested):

G10h4ck:
> Nice work. @G10h4ck : does it fits your AppVeyor script?

Yeah it is more or less what the appveyor script does

G10h4ck (approved):

PR #1318, opened 2018-08-03
fixed webui files generation under MSYS2

zapek:
The reason it failed is because MSYS2 uses unix tools and unix pathnames so I just made it use the  build.sh instead of build.bat.

I changed the previous target from "win32" to "win32-x" so it should keep running on appveyor which does use windows cmd.exe I think. Can someone using appveyor confirm the fix is fine?


G10h4ck:
AFAIK msys2 is used on appveyor too, anyway if some appveyor specific instruction is needed better to use appveyor { something here } instead of win32-x and the we can pass CONFIG+=appveyor to qmake

PR #1319, opened 2018-08-03
remove useless zeros from virtual peer ID

sehraf:


csoler:
Well, actually virtual peer IDs are 160bits long (same as RsPeerId) for code compatibility reasons because they act as RsPeerId in the code, but because they are derived from tunnel IDs which are 32bits long, I construct them by padding with zeroes. This is not a cryptographic problem whatsoever.
So in summary, I would prefer to keep the zeroes, as they are part of the ID.

sehraf:
OK

PR #1320, opened 2018-08-03
Added switch to enable native dialogs at compile time

hunbernd:
Native file dialogs were disabled because they caused errors on some platforms.
This commits adds a compile time switch to enable them again.

CONFIG+=use_native_dialogs

### DIFF
--- a/build_scripts/Windows-msys2/build/build.bat
+++ b/build_scripts/Windows-msys2/build/build.bat
@@ -48,7 +48,7 @@ echo.
 
 title Build - %SourceName%-%RsBuildConfig% [qmake]
 
-set RS_QMAKE_CONFIG="CONFIG+=%RsBuildConfig%"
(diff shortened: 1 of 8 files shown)


csoler:
Could you make RS use the misc:: methods in ShareManager.cpp and ChatWidget.cpp? We probably forgot to convert these. So in the end the only #if USE_NATIVE_DIALOGS will be needed in misc.cpp!
thx!

G10h4ck:
please prefix both CONFIG and DEFINES values with rs_/RS_

defnax:
native dialogs would be better on windows when attaching pictures qt dialogs dont have thumbnail view feature it would be cool that native comes back. when this pr is done?

attaching pictures or videos not easy with qt own dialogs dont see any thumbnail since few years

@hunbernd 

please preview this again @G10h4ck @csoler

PR #1321, opened 2018-08-03
WIP: Attempt to fix crashes caused by plugins on Windows

hunbernd:
Currently it turns libretroshare into a dll on Windows.

More to come...


G10h4ck:
is this to be considered for merging already or it is a work in progress ?

hunbernd:
I want to do some minor changes: fix packaging scripts, enable optimization on Windows.

Also it needs to be tested on Linux before merge.

G10h4ck:
Ok so it is better to put WIP (Work in progress) prefix to the PR title in the meanwhile and remove it when it should be considered for merging ;)

hunbernd:
I would add a label when I created the PR, but it seems like Github does not allow adding labels even if you are the owner of the PR.

PR #1323, opened 2018-08-13
V0.6 forum admin

csoler:
This PR is not yet ready but can already be tested.
- Forums can be edited by owner to set moderators.
- forum owner can pin up posts (missing UI) 
Moderators can edit other user's posts. The edited posts are authored by the moderator but earlier versions keep their original author.
The PR also contains post pinning, but the UI for it is still missing.

When a moderator is removed, his edit are shown as normal posts. I suppose we should rather not show them?

TODO: 
- automatically set the existing moderators when editing the forum properties
- list moderators in "Show details"
- dont show edited posts when the author is not a moderator.
- add the UI for pinning up posts.

Backward compatibility:
- new version can read and edit forum properties, add moderators.
- old version will still be able to load the forum, but not see any change to it and will display edited posts as normal posts.


felisucoibi:
Important thing: To be able to mark a thread or suthread as DELETED, and the gui should ignore this threads in the GUI, if a thread or suthread is DELETED it just dissapears from the list, maybe, a user can mark in FORUM settings in CONFIG to see deleted threads, and doing it, it just ignore the DELETED mark and shows the thread or subthread.

Dissapers the whole thread, not only the deleted message but the whole thread that dependes of this post.
Doing this we can delete completely a troll and all the noise the troll does with people answers.
The possibility to ignore admins deletes is good for decetralization, and people freedom in side rs, but i don't know if can be used as a evbil thing to hide things in deleted threads... or if is good for users to keep also deleted threads... but by now and with the good spirit of rs, is ok to keep deleted threads and be able to see them.

Maybe in the same config of see deleted threads, coud be an option of keep deleted threads, so you can choose to resend or not to your friends deleted threads....

felisucoibi:
By now works, just waiting for the 
TODO:

    automatically set the existing moderators when editing the forum properties
    list moderators in "Show details"
    dont show edited posts when the author is not a moderator.
    add the UI for pinning up posts.
Before asking for more things.

felisucoibi:
also i hope all these things and the to do list arrives to channels, so qwe can add and remove admins easyly now is like nobody knows who is admin who is not

felisucoibi:
My sugested TODO for 0.6.5 by now

TODO:

FORUMS AND CHANNELS : automatically set the existing moderators when editing the forum properties
FORUMS AND CHANNELS : list moderators in "Show details"
FORUMS AND CHANNELS : dont show edited posts when the author is not a moderator.
FORUMS AND CHANNELS : add the UI for pinning up posts. **THIS IS THE MOST IMPORTAT FEATURE BY NOW**

csoler:
I plan to work on these this week. Only problem is this one:

FORUMS AND CHANNELS : dont show edited posts when the author is not a moderator.

Edited posts are posts for which the original post ID is mentionned in the mOrigMsg field. But there's no way to know the author original post if it has disappeared (because it's too old). Consequently if the original post is gone it is not possible to decide if the edited post is editing the same author (in which case it is not a moderation action and we should still show it). So we have to choose between:
1 - if original post cannot be found, do not show edited post, whichever the author of the edited post is (independently of he being a moderator), and
2 - if original post cannot be found, always show edited post whether the author is a moderator or not.
It would be easier to handle if we had a flag that says "this post is moderating another post". I'll see if I find room for that.

felisucoibi:
I think is good to show it, option 2, BUT we need to be able to mark posts and its child messages as deleted, so we can clean forums of unnecesary posts, and also tis can be usefull inc ase a edited post we don't like it for example becase the personw ho edited is not anymore an admin, we can mark it as deleted, and thats all..

So option B is ok + be able to delete the post manually if not interesting at all.

Delete function is the last piece for having REAL FORUMS, i really need this function in forums to delete lot of threads in the forums i admin, in channels can be useful too, but in a channel you edit a old postwith new content and is ok, but in a forum there are more posts than you really need to reuse, so is more complciated if there is no delete function.

felisucoibi:
if i edit a pinned post it loose its category of pinned

felisucoibi:
Still i think pinned post are no showed on top of the forum list by default.
Also if i edit forum options, like description etc, the pinned posts dissapear.
Bests.

PR #1325, opened 2018-08-21
V0.6 content filtering

csoler:
Allows users to flag a few bad files and notify friends about these files. Still unfinished and needs some testing. Not to be merged yet!


csoler:
That can now be tested. It seems to work properly.  I left some debug output for testers.

PR #1331, opened 2018-09-04
JSON API improvements

G10h4ck:
Expose more via the API
Prevent crash calling methods on uninitialized services
Make bind address configurable and enforce it is 127.0.0.1 by default (thanks @sehraf for reporting)

### DIFF
--- a/libretroshare/src/retroshare/rsinit.h
+++ b/libretroshare/src/retroshare/rsinit.h
@@ -51,6 +51,15 @@ struct RsLoginHelper;
  */
 extern RsLoginHelper* rsLoginHelper;
 
+
+class RsAccounts;
+
+/**
+ * Pointer to global instance of RsAccounts needed to expose JSON API
+ * @jsonapi{development}
+ */
+extern RsAccounts* rsAccounts;
+
 /*!
  * Initialisation Class (not publicly disclosed to RsIFace)
  */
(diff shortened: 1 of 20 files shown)


libretroshare/src/rsserver/p3peers.cc
```diff
@@ -1099,6 +1109,72 @@ bool p3Peers::GetPGPBase64StringAndCheckSum(	const RsPgpId& gpg_id,
...
 }
 
+bool p3Peers::acceptInvite( const std::string& invite,
+                            ServicePermissionFlags flags )
+{
+	if(invite.empty()) return false;
+
+	const std::string* radixPtr(&invite);
+
+	RsUrl url(invite);
```
sehraf, line 1119: While passing an URL or cert works, i'm wondering if it just works by accident.
See the following from RsUrl:
When using a certificate (not an URL) `size_t schemeEndI = urlStr.find(schemeSeparator);` will return `string::npos`. This results in `mScheme = urlStr.substr(0, schemeEndI);` being the whole string ("*A value of string::npos indicates all characters until the end of the string.*") while `size_t hostBeginI = schemeEndI + 3;` will be 2.
The rest of the parsing will behave similar; using `string::npos` again `size_t queryBeginI = urlStr.find(querySeparator, hostBeginI);` (`hostBeginI` is still **2**) again for math `string qStr = urlStr.substr(queryBeginI+1, queryEndI-queryBeginI-1);`
So "parsing" a certificate with RsUrl seems to be completly broken (as it is not an url) and only happened to work by accident. It this intentional? If so maybe add a comment.

libretroshare/src/rsserver/p3peers.cc
```diff
@@ -1099,6 +1109,72 @@ bool p3Peers::GetPGPBase64StringAndCheckSum(	const RsPgpId& gpg_id,
...
+	std::map<std::string, std::string> query(url.query());
+
+	if(query.find("radix") != query.end())
+		radixPtr = &query["radix"];
+
+	const std::string& radix(*radixPtr);
```
sehraf, line 1125: Since `radixPtr` is set to be the begin of the string i assume the failure of RsUrl was taken into account. And thus using a certificate works fine.

libretroshare/src/rsserver/p3peers.cc
G10h4ck: RsUrl is a very naive but safe implementation of URL parsing, as URI standard is very complex RsUrl has no way to determine if an URL parsing has been successful or not. The acceptInvite code take in account of RsUrl being so naive so it works by design not by chance, if the parsed stuff has no `radix` in the query part (that is the case when a plain cert is passed) `radixPtr` is untouched.

libretroshare/src/rsserver/p3peers.cc
G10h4ck: Yeah it was like that by design ;)

libretroshare/src/rsserver/p3peers.cc
sehraf: :+1:

libretroshare/src/retroshare/rsinit.h
```diff
@@ -51,6 +51,15 @@ struct RsLoginHelper;
...
 extern RsLoginHelper* rsLoginHelper;
 
+
+class RsAccounts;
+
+/**
+ * Pointer to global instance of RsAccounts needed to expose JSON API
+ * @jsonapi{development}
+ */
+extern RsAccounts* rsAccounts;
```
csoler: You shouldn't need this: RsAccounts only has static members (I made sure of it because I want to simplify the code so I started with making this classs static). So if possible convert the getCurrentAccountId() into a static method. thx!

G10h4ck: It is needed by JSON API because the code is automatically generated and because almost all services uses this "wild pointer" pattern the code generator assumes the pointer exists to call the methods, but I'll make getCurrentAccountId static so it can be used without the pointer in the other parts of the code

PR #1332, opened 2018-09-05
make p3GxsTrans a class

sehraf:
It is also already forward declared as a class.


G10h4ck:
Did you double checked that nothing is broken by this PR? Because all inherited types are now inherited as private.

Edit:
As expected it breaks because inherited types are now privates see
https://travis-ci.org/RetroShare/RetroShare/jobs/424803042#L1117

sehraf:
You are right. As Travis shows it didn't compile. I've rebased and force pushed the branch with an additional change. `p3GxsTrans` now inherits everything as public.
This is compile tested and shortly run tested.

G10h4ck:
the line seems to be more then 80 chars at first glance, please break it if it is teh case ;)

PR #1334, opened 2018-09-06
auto send of own signed GXS ids to friends in discovery

csoler:


libretroshare/src/rsitems/rsdiscovery2items.cc
```diff
@@ -150,139 +151,8 @@ void RsDiscContactItem::serial_process(RsGenericSerializer::SerializeJob j,RsGen
...
+void RsDiscIdentityListItem::serial_process(RsGenericSerializer::SerializeJob j,RsGenericSerializer::SerializeContext& ctx)
 {
-	return;
-}
-
-void 	RsDiscServicesItem::clear()
-{
-	version.clear();
-	mServiceIdMap.TlvClear();
+	RsTypeSerializer::serial_process(j,ctx,ownIdentityList,"identityList");
```
G10h4ck: When possible it is better to use RS_SERIAL_PROCESS(ownIdentityList)

libretroshare/src/rsitems/rsdiscovery2items.h
```diff
@@ -138,6 +139,23 @@ class RsDiscContactItem: public RsDiscItem
...
+	RsDiscIdentityListItem()
+	    :RsDiscItem(RS_PKT_SUBTYPE_DISC_IDENTITY_LIST)
+	{
+		setPriorityLevel(QOS_PRIORITY_RS_DISC_CONTACT);
+	}
+
+    virtual ~RsDiscIdentityListItem() {}
+
+    virtual void clear() { ownIdentityList.clear() ; }
+	virtual void serial_process(RsGenericSerializer::SerializeJob /* j */,RsGenericSerializer::SerializeContext& /* ctx */);
```
G10h4ck: The commented parameters suggests they are not used in the implementation, while i guess they are, if it is the case they shouldn't be commented out

libretroshare/src/services/p3discovery2.cc
```diff
@@ -94,9 +95,8 @@ void DiscPgpInfo::mergeFriendList(const std::set<PGPID> &friends)
 }
 
 
-p3discovery2::p3discovery2(p3PeerMgr *peerMgr, p3LinkMgr *linkMgr, p3NetMgr *netMgr, p3ServiceControl *sc)
-:p3Service(), mPeerMgr(peerMgr), mLinkMgr(linkMgr), mNetMgr(netMgr), mServiceCtrl(sc),
-	mDiscMtx("p3discovery2")
+p3discovery2::p3discovery2(p3PeerMgr *peerMgr, p3LinkMgr *linkMgr, p3NetMgr *netMgr, p3ServiceControl *sc, RsGixs *gixs)
```
G10h4ck, line 98: Better to take a reference `(const?) RsGixs& gixs` instead of a pointer in this case as it must not be null

libretroshare/src/services/p3discovery2.cc
```diff
@@ -262,17 +262,14 @@ int p3discovery2::handleIncoming()
...
-#endif
-
 	int nhandled = 0;
 	// While messages read
 	while(NULL != (item = recvItem()))
 	{
 		RsDiscPgpListItem *pgplist = NULL;
 		RsDiscPgpCertItem *pgpcert = NULL;
 		RsDiscContactItem *contact = NULL;
+		RsDiscIdentityListItem *gxsidlst = NULL;
```
G10h4ck: C++11 suggest the safer `nullptr` to be used instead of `NULL` that is just `#define NULL 0`

libretroshare/src/services/p3discovery2.cc
```diff
@@ -287,14 +284,14 @@ int p3discovery2::handleIncoming()
 				recvOwnContactInfo(item->PeerId(), contact);
             else
                 processContactInfo(item->PeerId(), contact);
-
-			continue;
 		}
-
-		if (NULL != (pgpcert = dynamic_cast<RsDiscPgpCertItem *> (item)))	
-		{
+        else  if (NULL != (gxsidlst = dynamic_cast<RsDiscIdentityListItem *> (item)))
```
G10h4ck: Same here either use `nullptr` or completely remove the comparison as only a null `nullptr` evaluates to false anyway

libretroshare/src/services/p3discovery2.cc
```diff
@@ -287,14 +284,14 @@ int p3discovery2::handleIncoming()
...
+        {
+            recvIdentityList(item->PeerId(),gxsidlst->ownIdentityList) ;
+            delete item;
+        }
+        else  if (NULL != (pgpcert = dynamic_cast<RsDiscPgpCertItem *> (item)))
```
G10h4ck: Same here either use `nullptr` or completely remove the comparison as only a null `nullptr` evaluates to `false` anyway

libretroshare/src/services/p3discovery2.cc
```diff
@@ -364,6 +358,18 @@ void p3discovery2::sendOwnContactInfo(const SSLID &sslid)
...
+
+        rsIdentity->getOwnIds(pkt2->ownIdentityList,true);
+        pkt2->PeerId(sslid) ;
+
+#ifdef P3DISC_DEBUG
+		std::cerr << "p3discovery2::sendOwnContactInfo() sending own signed identity list:" << std::endl;
+        for(auto it(pkt2->ownIdentityList.begin());it!=pkt2->ownIdentityList.end();++it)
+            std::cerr << "  identity: " << (*it).toStdString() << std::endl;
+#endif
+		sendItem(pkt2);
```
G10h4ck, line 372: I suppose the sender take ownership of the item, and delete it at right moment

libretroshare/src/services/p3discovery2.h
```diff
@@ -75,7 +76,7 @@ class p3discovery2: public RsDisc, public p3Service, public pqiServiceMonitor, p
 {
 	public:
 
-	p3discovery2(p3PeerMgr *peerMgr, p3LinkMgr *linkMgr, p3NetMgr *netMgr, p3ServiceControl *sc);
+	p3discovery2(p3PeerMgr *peerMgr, p3LinkMgr *linkMgr, p3NetMgr *netMgr, p3ServiceControl *sc,RsGixs *gixs);
```
G10h4ck, line 79: Better to take a reference `(const?) RsGixs& gixs` instead of a pointer in this case as it must not be null

libretroshare/src/services/p3discovery2.h
```diff
@@ -131,6 +133,7 @@ virtual void setGPGOperation(AuthGPGOperation *operation);
 	p3LinkMgr *mLinkMgr;
 	p3NetMgr  *mNetMgr;
 	p3ServiceControl *mServiceCtrl;
+    RsGixs *mGixs ;
```
G10h4ck, line 136: Better to take a reference (const?) RsGixs& gixs instead of a pointer in this case as it must not be null

libretroshare/src/services/p3discovery2.cc
csoler: I agree on the principle. But all functions use these pointer-based system, so I kept it for consistency.

libretroshare/src/services/p3discovery2.cc
csoler: if I turn it into if(pgpcert = dynamic_cast<>...) then the compiler is going to warn about a possible missed ==, which is also confusing. So I'll use the nullptr trick.

libretroshare/src/services/p3discovery2.cc
csoler: yes

G10h4ck:
On Monday, 10 September 2018 20:00:56 CEST csoler wrote:
> csoler commented on this pull request.
> 
> >  		}
> 
> -
> -		if (NULL != (pgpcert = dynamic_cast<RsDiscPgpCertItem *> (item)))
> -		{
> +        else  if (NULL != (gxsidlst = dynamic_cast<RsDiscIdentityListItem
> *> (item)))
> 
> if I turn it into if(pgpcert = dynamic_cast<>...) then the compiler is going
> to warn about a possible missed ==, which is also confusing. So I'll use
> the nullptr trick.

AFAIR in case of assignation () around should be kept  to evaluate the result 
of assignation so it would become

`if((pgpcert = dynamic_cast<RsDiscPgpCertItem *> (item)))`

BTW it is ok for me with nullptr too ;)

PR #1336, opened 2018-09-11
Clean up the mess with RetroShare version

G10h4ck:
RetroShare version now is settable at compilation time via qmake variables, if it is not set attempt to determine it via git describe, this mean no more 1234567 in the main window bar, and strange macro around to check minimum version and so on


felisucoibi:
+1

PR #1340, opened 2018-09-14
remove obsolete version scripts

sehraf:
G10h4ck said that there are obsolete since #1336 and should not be used anymore.


G10h4ck:
yeah this script should not be used anymore!

PR #1341, opened 2018-09-14
Rs fixes first build mac os issues

xn-zhao:
fixes all build issues on Mac OSx.


RetroShare.pro
```diff
@@ -17,7 +17,9 @@
 ################################################################################
 
 !include("retroshare.pri"): error("Could not include file retroshare.pri")
-
+#atai add this one
+CONFIG += qt gui uic qrc resources uitools debug
+CONFIG+=rs_async_chat
```
csoler: why do you need this?

RetroShare.pro
```diff
@@ -17,7 +17,9 @@
 ################################################################################
 
 !include("retroshare.pri"): error("Could not include file retroshare.pri")
-
+#atai add this one
+CONFIG += qt gui uic qrc resources uitools debug
```
csoler: I dont think this is needed. libretroshare that is below this directory does not need qt. All this should be in retroshare-gui/src/retroshare-gui.pro, and it's probably already there.

libretroshare/src/chat/rschatitems.cc
```diff
@@ -29,7 +29,7 @@
 
 #include "chat/rschatitems.h"
 
-//#define CHAT_DEBUG 1
+#define CHAT_DEBUG 1
```
csoler, line 32: you forgot to leave this commented out.

libretroshare/src/util/cxx11retrocompat.h
```diff
@@ -19,15 +19,15 @@
  * along with this program. If not, see <https://www.gnu.org/licenses/>.       *
  *                                                                             *
  *******************************************************************************/
-#pragma once
+//#pragma once
```
csoler: #pragma once has no reason to be commented out. Why would you need to do that?

Commenting the entire bloc out below seems dangerous to me given that is has been put here to allow to compile on other systems. At least put a proper #ifndef MACOS around.

openpgpsdk/src/openpgpsdk.pro
```diff
@@ -2,7 +2,8 @@
 
 TEMPLATE = lib
 CONFIG += staticlib
-
+## Sep 11 2018 - meiyousixin - fix bug: can not find openssl/bn.h
+INCLUDEPATH += /usr/local/include
```
csoler: that should be set in the .pri I suppose.

retroshare.pri
```diff
@@ -118,9 +118,11 @@ CONFIG *= rs_gxs_trans
 
 # To enable GXS based async chat append the following assignation to qmake
 # command line "CONFIG+=rs_async_chat"
-CONFIG *= no_rs_async_chat
+#CONFIG *= no_rs_async_chat
+CONFIG+=rs_async_chat
```
csoler: rs_async_chat should be left disabled by default. How can adding additional code help compilation anyway?

csoler:
thx for this PR. Your fixes give some clues about what to do to fix compilation, but there's lots of questions to solve before we merge this. Please see the inlined comments.

RetroShare.pro
xn-zhao: Thank you for your comment.
The first one I think we can remove it. I just add for debug build, but we can set it in another place.
The second one is the one that I think it can send msg in the distant chat (in people tab) without establishing the secure tunnel, but it still not work at this time, so we can remove it too. I just try to commit to the one master of one account for first test, but it 'touchs' to all our community. So sorry.

libretroshare/src/chat/rschatitems.cc
xn-zhao: Thanks for this

libretroshare/src/util/cxx11retrocompat.h
xn-zhao: Yes, I think in some places we will put the platform checking, so that it will work for all. This one is the case.

openpgpsdk/src/openpgpsdk.pro
xn-zhao: ok, i will move it to .pri, thanks.

libretroshare/src/util/cxx11retrocompat.h
G10h4ck: `#pragma once` should be supported by any decent compiler out there

csoler:
Sorry guys, but a PR that changes 899 files including global replacement of "retroshare" for "p2punseen" does not make sense. If you really want to contribute, please make small pull requests with sensible changes.

PR #1344, opened 2018-09-15
Adding Mac OSX bdboot.txt fixed

unsenep2pdev:
Currently, you have to manually add bdboot.txt into Mac OSX RS Data directory. So by adding this fix will make configuration load automatically every new Identity on Mac OSX Instance.


libbitdht/src/util/bdthreads.cc
```diff
@@ -124,7 +124,11 @@ void bdThread::join() /* waits for the the mTid thread to stop */
 #if defined(_WIN32) || defined(__MINGW32__)
 	/* Its a struct in Windows compile and the member .p ist checked in the pthreads library */
 #else
-	if(mTid > 0)
+    #if __APPLE__
```
G10h4ck: this fi else should be subistituted with `if(mTid)` tha works both on linux and mac without preprocessing

libresapi/src/api/json.cpp
```diff
@@ -19,7 +19,11 @@
  * along with this program. If not, see <https://www.gnu.org/licenses/>.       *
  *                                                                             *
  *******************************************************************************/
-#define nullptr 0
+#if __APPLE__
```
G10h4ck: this should be moved into `libretroshare/src/util/cxx11retrocompat.h` and then you can add `#include "util/cxx11retrocompat.h"` on this file, anyway RetroShare needs C++11 to compile

libretroshare/src/util/cxx11retrocompat.h
```diff
@@ -21,13 +21,20 @@
...
-#	define GCC_VERSION (__GNUC__*10000+__GNUC_MINOR__*100+__GNUC_PATCHLEVEL__)
-#	if GCC_VERSION < 40700
-#		define override
-#		define final
-#	endif // GCC version < 40700
-#	if GCC_VERSION < 40600
-#		define nullptr NULL
-#	endif // GCC_VERSION < 40600
-#endif //defined GNUC
+    #define GCC_VERSION (__GNUC__*10000+__GNUC_MINOR__*100+__GNUC_PATCHLEVEL__)
```
G10h4ck: `#` must be the first character of the line so the indentation goes after it as it was before, moreover you should use tab instead of space for indentation as described in our [coding style](https://retroshare.readthedocs.io/en/latest/developer/coding/#coding-style)

G10h4ck (changes requested):
Please remove all those changes that just remove tabs and put spaces, that make reviewing the code more difficult, moreover you should avoid unrelated changes on the same pull request (better open two different pull requests for two different topic).
Altering the indentation using space instead of tab is against our [coding style](https://retroshare.readthedocs.io/en/latest/developer/coding/#coding-style) so please use tab in the lines you edits

libbitdht/src/util/bdthreads.cc
unsenep2pdev: This recommendation has been fixed and update.

libresapi/src/api/json.cpp
unsenep2pdev: fixed and update.

libretroshare/src/util/cxx11retrocompat.h
unsenep2pdev: fixed

PhenomRetroShare:
Hello,
You need to make a new branch for each PR.

retroshare.pri
```diff
@@ -88,7 +88,8 @@ no_sqlcipher:CONFIG -= sqlcipher
 # To enable autologin (this is higly discouraged as it may compromise your node
 # security in multiple ways) append the following assignation to qmake command
 # line "CONFIG+=rs_autologin"
-CONFIG *= no_rs_autologin
+#CONFIG *= no_rs_autologin
+CONFIG +=rs_autologin
```
PhenomRetroShare: Don't change this.
Only add config when you call qmake, as explained upper.

PR #1346, opened 2018-09-18
V0.6 improved gui

csoler:
Adds an additional context menu into Home to include all IPs from history or not.


felisucoibi:
Maybe you can fix the HOME icon that is not 4k ready but the others they are. bests.

PR #1348, opened 2018-09-19
fixed compilation on Windows/MSYS2

zapek:
This  prevents the following error:

C:/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/7.3.0/../../../../i686-w64-mingw32/bin/ld.exe: ./temp/obj/GxsGroupDialog.o: in function `ZN21FriendSelectionWidget14setSelectedIdsI17t_RsGenericIdTypeILj16ELb0ELj6EELNS_6IdTypeE4EEEvRKSt3setIT_St4lessIS5_ESaIS5_EEb':
C:\msys64\home\zapek\RetroShare\retroshare-gui\src/./gui/common/FriendSelectionWidget.h:98: multiple definition of `void FriendSelectionWidget::setSelectedIds<t_RsGenericIdType<16u, false, 6u>, (FriendSelectionWidget::IdType)4>(std::set<t_RsGenericIdType<16u, false, 6u>, std::less<t_RsGenericIdType<16u, false, 6u> >, std::allocator<t_RsGenericIdType<16u, false, 6u> > > const&, bool)'; ./temp/obj/FriendSelectionWidget.o:C:\msys64\home\zapek\RetroShare\retroshare-gui\src/gui/common/FriendSelectionWidget.cpp:693: first defined here

### DIFF
--- a/retroshare-gui/src/gui/common/FriendSelectionWidget.cpp
+++ b/retroshare-gui/src/gui/common/FriendSelectionWidget.cpp
@@ -689,7 +689,7 @@ void FriendSelectionWidget::requestGXSIdList()
 	mIdQueue->requestGroupInfo(token, RS_TOKREQ_ANSTYPE_DATA, opts, IDDIALOG_IDLIST);
 }
 
-template<> void FriendSelectionWidget::setSelectedIds<RsGxsId,FriendSelectionWidget::IDTYPE_GXS>(const std::set<RsGxsId>& ids, bool add)
+template<> inline void FriendSelectionWidget::setSelectedIds<RsGxsId,FriendSelectionWidget::IDTYPE_GXS>(const std::set<RsGxsId>& ids, bool add)
 {
     mPreSelectedGxsIds = ids ;
     requestGXSIdList();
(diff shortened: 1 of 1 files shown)


G10h4ck:
Have you tested it on appveyor too?

zapek:
>           Have you tested it on appveyor too?

Nope. No idea how to do that.

G10h4ck:
Anyway I guess you tested this fix, would you be so kind to put some comment in the code explaining a bit the issue and the fix? So when someone stumble on that inline keyword there after some year per  understand why that inline there is important ;)

hunbernd:
>Have you tested it on appveyor too?
>Nope. No idea how to do that.

It should appear here, under the checks section, but I cannot see it. 
Something is wrong with the appveyor configuration.

I can compile RS with this fix.
And appveyor is doing the deploy step now, so the build was OK.

zapek:
Tested and works here, yes. It's because MinGW has a slightly different version of gcc (older I suppose) than on Linux and it considers setSelectedIds() as being the same so the workaround is to inline one.

G10h4ck:
Great thanks!

PR #1349, opened 2018-09-19
Implement JSON API HTTP Basic authentication

G10h4ck:
jsonapi-generator is now capable of generating API for headers outside
  retroshare/ directory
jsonapi-generator do a bit of methods parameter sanity check
JsonApiServer is now integrated in the rsinit hell like other services
Add *::exportGPGKeyPairToString to a bunch of classes in cascade
RsControl is now capable of calling back a function when retroshare is almost
  completely stopped, this is useful when running retroshare toghether with
  externally managed runloop such as QCoreApplication
Expose a bunch of methods through JSON API
retroshare-nogui remove some dead code and fix stopping from the RetroShare API


sehraf:
> +809 −388 

for the future: could you split up your patches? it would make reviewing them easier

retroshare-nogui/src/retroshare.cc
```diff
@@ -244,35 +183,19 @@ int main(int argc, char **argv)
...
-#ifndef WINDOWS_SYS
-			sleep(1);
-#else
-			Sleep(1000);
-#endif
-		}
-
-		rstime::rs_usleep(1000);
-
+		rstime::rs_usleep(10*1000);
```
sehraf, line 197: is this increase intentional?

G10h4ck: If you consider the code I deleted that was executed always it is not an increase, anyway that loop does nothing at all, so as soon as we sleep an amount orders of magnitudes bigger then CPU clock cycle is ok

G10h4ck:
> > +809 −388
> 
> for the future: could you split up your patches? it would make reviewing them easier

Yeah I do that usually but for developing authentication I had to integrate it at same time with RetroShare services functionalities and the result is many edits before it builds without compiler complaining or crash, or unexpected behaviors ;)

sehraf:
Right.... Sleep are seconds, not milliseconds. I always get confused with the units 

EDIT: this comment was meant to be an answer to the usleep thing but i guess github mail response can't handle this.

libretroshare/src/jsonapi/jsonapi.cpp
```diff
@@ -84,13 +92,13 @@ static bool checkRsServicePtrReady(
...
-JsonApiServer::JsonApiServer(
-        uint16_t port, const std::string& bindAddress,
-        const std::function<void(int)> shutdownCallback ) :
-    mPort(port), mBindAddress(bindAddress), mShutdownCallback(shutdownCallback)
+JsonApiServer::JsonApiServer(uint16_t port, const std::string& bindAddress,
+        const std::function<bool(const std::string&)> newAccessRequestCallback ) :
+    mPort(port), mBindAddress(bindAddress),
+    mNewAccessRequestCallback(newAccessRequestCallback),
+    configMutex("JsonApiServer config")
 {
```
sehraf, line 101: you forgot to update `WebuiPage.cpp`
```
../../../RetroShare/retroshare-gui/src/gui/settings/WebuiPage.cpp: In static member function 'static bool WebuiPage::checkStartWebui()':
../../../RetroShare/retroshare-gui/src/gui/settings/WebuiPage.cpp:121:54: error: no matching function for call to 'JsonApiServer::JsonApiServer(int, const char*, WebuiPage::checkStartWebui()::<lambda(int)>)'
              [](int /*ec*/) { std::raise(SIGTERM); } );
                                                      ^
```

libretroshare/src/rsserver/rsinit.cc
```diff
@@ -1277,6 +1248,14 @@ int RsServer::StartupRetroShare()
 	//
 	mPluginsManager->loadPlugins(programatically_inserted_plugins) ;
 
+#ifdef RS_JSONAPI
+	{
+		mConfigMgr->addConfiguration("jsonApi.cfg", jsonApiServer);
+		RsFileHash dummyHash;
+		jsonApiServer->loadConfiguration(dummyHash);
```
sehraf, line 1256: is this explicit loading needed?

sehraf:
is get this when running (after fixing the wrong call to `JsonApiServer`):
```GDB
Thread 1 "retroshare" received signal SIGSEGV, Segmentation fault.
0x00007ffff581e244 in pthread_mutex_lock () from /usr/lib/libpthread.so.0
(gdb) where
#0  0x00007ffff581e244 in pthread_mutex_lock () at /usr/lib/libpthread.so.0
#1  0x0000555555f048f6 in RsMutex::lock() (this=this@entry=0x8) at ../../../RetroShare/libretroshare/src/util/rsthreads.cc:321
#2  0x0000555555e45670 in RsStackMutex::RsStackMutex(RsMutex&) (mtx=..., this=<synthetic pointer>) at ../../../RetroShare/libretroshare/src/util/rsthreads.h:93
#3  0x0000555555e45670 in pqiConfig::setFilename(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (this=0x0, name="/home/michael/.retroshare/LOC06_fa8b0b2f12acb6285f9afb77b31767bd/config/jsonApi.cfg")
    at ../../../RetroShare/libretroshare/src/pqi/p3cfgmgr.cc:593
#4  0x0000555555e461ac in p3ConfigMgr::addConfiguration(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, pqiConfig*) (this=this@entry=0x555557988f50, file="jsonApi.cfg", conf=conf@entry=0x0)
    at ../../../RetroShare/libretroshare/src/pqi/p3cfgmgr.cc:184
#5  0x0000555555e698bc in RsServer::StartupRetroShare() (this=0x555556d27360 <RsControl::instance()::rsicontrol>) at /usr/include/c++/8.2.1/ext/new_allocator.h:79
#6  0x000055555592e73f in main(int, char**) (argc=<optimized out>, argv=<optimized out>) at ../../../RetroShare/retroshare-gui/src/main.cpp:411

```

libretroshare/src/rsserver/rsinit.cc
G10h4ck: Don't know didn't tested if configuration (extra authentication tokens) is loaded without that, I see other services upper in the code do the same, but if you want to test you are welcome

G10h4ck:
> is get this when running (after fixing the wrong call to `JsonApiServer`):
> 
> ```gdb
> Thread 1 "retroshare" received signal SIGSEGV, Segmentation fault.
> 0x00007ffff581e244 in pthread_mutex_lock () from /usr/lib/libpthread.so.0
> (gdb) where
> #0  0x00007ffff581e244 in pthread_mutex_lock () at /usr/lib/libpthread.so.0
> #1  0x0000555555f048f6 in RsMutex::lock() (this=this@entry=0x8) at ../../../RetroShare/libretroshare/src/util/rsthreads.cc:321
> #2  0x0000555555e45670 in RsStackMutex::RsStackMutex(RsMutex&) (mtx=..., this=<synthetic pointer>) at ../../../RetroShare/libretroshare/src/util/rsthreads.h:93
> #3  0x0000555555e45670 in pqiConfig::setFilename(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (this=0x0, name="/home/michael/.retroshare/LOC06_fa8b0b2f12acb6285f9afb77b31767bd/config/jsonApi.cfg")
>     at ../../../RetroShare/libretroshare/src/pqi/p3cfgmgr.cc:593
> #4  0x0000555555e461ac in p3ConfigMgr::addConfiguration(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, pqiConfig*) (this=this@entry=0x555557988f50, file="jsonApi.cfg", conf=conf@entry=0x0)
>     at ../../../RetroShare/libretroshare/src/pqi/p3cfgmgr.cc:184
> #5  0x0000555555e698bc in RsServer::StartupRetroShare() (this=0x555556d27360 <RsControl::instance()::rsicontrol>) at /usr/include/c++/8.2.1/ext/new_allocator.h:79
> #6  0x000055555592e73f in main(int, char**) (argc=<optimized out>, argv=<optimized out>) at ../../../RetroShare/retroshare-gui/src/main.cpp:411
> ```

plase test without the explicit loadinng and see if it still happen, and if extra token are loaded

libretroshare/src/rsserver/rsinit.cc
sehraf: `mConfigMgr->loadConfiguration();`
takes care of loading all configs

sehraf:
still happens. Also it happend before the loading `addConfiguration`

EDIT: the problem is that the jsonapi port is not set when using rs-gui. this results in `jsonApiServer` being `nullptr`. Also the line in question is missing an `if != nullptr`

libretroshare/src/rsserver/rsinit.cc
G10h4ck: at which point? jsonApiServer may need the authentication tokens to be loaded ASAP so the API can be used ASAP

libretroshare/src/jsonapi/jsonapi.cpp
G10h4ck: not used there anymore ;) now JSON API has it's own configuration page

PR #1350, opened 2018-09-20
fixed JSONAPI build for Windows

zapek:
JSONAPI now builds on Windows, some notes:
- I added doxygen and cmake as tool requirements in the docs
- $$system_path() shouldn't be needed anymore as MSYS2 uses unix path
- doxygen generation didn't work when not using shadow builds, now it does
- wsock32 is needed because ASIO generates 2 old winsock functions
- I didn't find a more elegant way to have a "jsonapi-generator" executable name work for both shadow/non-shadow and windows/linux, suggestions or fixes welcome
- I didn't test a linux build because I have to setup a machine/VM for it but I tried to not break anything

### DIFF
--- a/libretroshare/src/libretroshare.pro
+++ b/libretroshare/src/libretroshare.pro
@@ -860,16 +860,25 @@ rs_gxs_trans {
 }
 
 rs_jsonapi {
-    JSONAPI_GENERATOR_SRC=$$system_path($$clean_path($${RS_SRC_PATH}/jsonapi-generator/src/))
-    JSONAPI_GENERATOR_OUT=$$system_path($$clean_path($${RS_BUILD_PATH}/jsonapi-generator/src/))
-    JSONAPI_GENERATOR_EXE=$$system_path($$clean_path($${JSONAPI_GENERATOR_OUT}/jsonapi-generator))
-    DOXIGEN_INPUT_DIRECTORY=$$system_path($$clean_path($${PWD}))
-    DOXIGEN_CONFIG_SRC=$$system_path($$clean_path($${RS_SRC_PATH}/jsonapi-generator/src/jsonapi-generator-doxygen.conf))
-    DOXIGEN_CONFIG_OUT=$$system_path($$clean_path($${JSONAPI_GENERATOR_OUT}/jsonapi-generator-doxygen.conf))
-    WRAPPERS_INCL_FILE=$$system_path($$clean_path($${JSONAPI_GENERATOR_OUT}/jsonapi-includes.inl))
-    WRAPPERS_REG_FILE=$$system_path($$clean_path($${JSONAPI_GENERATOR_OUT}/jsonapi-wrappers.inl))
+    JSONAPI_GENERATOR_SRC=$$clean_path($${RS_SRC_PATH}/jsonapi-generator/src/)
+    JSONAPI_GENERATOR_OUT=$$clean_path($${RS_BUILD_PATH}/jsonapi-generator/src/)
+    win32 {
+        CONFIG(release, debug|release) {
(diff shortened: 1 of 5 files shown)


retroshare-gui/src/retroshare-gui.pro
```diff
@@ -170,7 +170,7 @@ win32-g++ {
 
 	OBJECTS_DIR = temp/obj
 
-    dLib = ws2_32 gdi32 uuid ole32 iphlpapi crypt32 winmm
+    dLib = ws2_32 wsock32 gdi32 uuid ole32 iphlpapi crypt32 winmm
```
G10h4ck: If `wsock32` is needed only when using JSON API it should be added conditionally only when JSON API is used and it probably goes into use-libretroshare.pro not retroshare-gui

libretroshare/src/libretroshare.pro
```diff
@@ -860,16 +860,25 @@ rs_gxs_trans {
...
-    JSONAPI_GENERATOR_OUT=$$system_path($$clean_path($${RS_BUILD_PATH}/jsonapi-generator/src/))
-    JSONAPI_GENERATOR_EXE=$$system_path($$clean_path($${JSONAPI_GENERATOR_OUT}/jsonapi-generator))
-    DOXIGEN_INPUT_DIRECTORY=$$system_path($$clean_path($${PWD}))
-    DOXIGEN_CONFIG_SRC=$$system_path($$clean_path($${RS_SRC_PATH}/jsonapi-generator/src/jsonapi-generator-doxygen.conf))
-    DOXIGEN_CONFIG_OUT=$$system_path($$clean_path($${JSONAPI_GENERATOR_OUT}/jsonapi-generator-doxygen.conf))
-    WRAPPERS_INCL_FILE=$$system_path($$clean_path($${JSONAPI_GENERATOR_OUT}/jsonapi-includes.inl))
-    WRAPPERS_REG_FILE=$$system_path($$clean_path($${JSONAPI_GENERATOR_OUT}/jsonapi-wrappers.inl))
-
-    restbed.target = $$system_path($$clean_path($${RESTBED_BUILD_PATH}/library/librestbed.a))
+    JSONAPI_GENERATOR_SRC=$$clean_path($${RS_SRC_PATH}/jsonapi-generator/src/)
```
G10h4ck, line 863: Is there an advantage removing `$$system_path()` ? Because I think using both functions together is more cross build system proof

zapek: The problem is that $$system_path() gives a path like C:\msys64\home\zapek\RetroShare\jsonapi-generator and that breaks the shell which is bash, not PowerShell or cmd.exe.

retroshare-gui/src/retroshare-gui.pro
zapek: You're right. I pushed a fix.

libretroshare/src/libretroshare.pro
G10h4ck: shouldn't `$$shell_path()` instead of `$$system_path()` used then instead of removing it?

zapek: It depends.
Removing it (ie. nothing) gives: C:/msys64/home/blabla
$$shell_path() gives: /C/msys64/home/blabla

The former works better if it's fed to a Windows tool (it's a valid Windows path). The later is needed for some unix tools that do some path manipulation (ie. doxygen) but might break if it's given to a native tool.

For Linux/OSX it doesn't mater, both path should be the same.

G10h4ck: Thanks for explanation I'll try it on linux to see if it keep working there ;)

PR #1353, opened 2018-09-27
V0.6 file lists

csoler:
Displays extra file list in the hierarchy of "my files".
To be tested. Not fully working yet.


csoler:
Seems to be working fine, except for a bug that prevents flat mode to display anything when the software is started with an non empty extra list. I haven't been able to figure it out yet. It seems not connected to this particular PR.

felisucoibi:
at the end it deletes after restaring the list? because my list was empty with the new menu in my files, btw flat mode is broken since ever needs some work.

PR #1356, opened 2018-09-30
Fix warnings

PhenomRetroShare:


G10h4ck:
yeah those warnings are annoying as this touches SSL parts too I hope @csoler review and merge this soon

csoler:
why are the "# if OPENSSL_VERSION_NUMBER < 0x10100000L" needed?? It seems to compile without warnings on ubuntu 18.04 with 1.1.0g already.

csoler (commented):
I'm shared about using override. I understand that it helps when reading the code. However, old compiler do not like it. 

I don't understand the need for the "# if OPENSSL_VERSION_NUMBER < 0x10100000L". It seems to compile file on ubuntu 18.04 with openssl 1.1.0g already.

G10h4ck:
Do not worry about `override` not being supported by old compiler we have it defined here https://github.com/RetroShare/RetroShare/blob/master/libretroshare/src/util/cxx11retrocompat.h#L27  just for them ;)

PR #1362, opened 2018-10-05
V0.6 improved gui

csoler:


csoler:
Normally this PR shouldn't change anything. But the code to check the config files signature was badly designed, so I tried to fix it. If your config files load correctly with this PR, I can merge it. It works with me.

PR #1363, opened 2018-10-05
New cross-platform target retroshare-service

G10h4ck:
This PR add a retroshare "pure" service target retroshare-service.
retroshare-service doesn't need any interaction of the user at startup, unlike
retroshare-nogui which requires the user to login on the shell this doesn't
even need a TTY.
At startup this just parse command line, read the PGP key-ring, look for
available locations, and start listening for JSON API requests.
Another difference with retroshare-nogui is that this is capable to
generate/import PGP identities, generate locations, and in general anything
possible through the RetroShare API.
retroshare-service is suitable also to run it as a system service, even
in very constrained systems such as Android or a Docker container.
retroshare-service drop support for libresapi so only the new JSON API
is exposed, it will completely obsolete retroshare-android-service once
retroshare-qml-app is ported to the new JSON API.
A package for Android is also provided to run as a system service independent from which apps will take advantage of it, a button in the menu to eventually stop/restart it is provided too (also because AndroidManifest.xml requires at least an activity to be present ). Different apps could use retroshare-service to provide user level services like blogging, chat, messaging, F2F version of memetro.net etc.
As restbed need full C++11 support a fresh android NDK is needed to compile retroshare-service, in the build_scripts/Android folder a script to create a toolchain based on clang and new NDK is provided.


RetroShare.pro
```diff
@@ -36,22 +36,34 @@ libretroshare.depends = openpgpsdk libbitdht
...
 }
 
 retroshare_nogui {
     SUBDIRS += retroshare_nogui
     retroshare_nogui.file = retroshare-nogui/src/retroshare-nogui.pro
-    retroshare_nogui.depends = libretroshare libresapi
     retroshare_nogui.target = retroshare_nogui
+
+    libresapi {
+        retroshare_gui.depends = libresapi
```
sehraf: Typo, this should be nogui

G10h4ck: great catch!

jsonapi-generator/src/jsonapi-generator.cpp
```diff
@@ -45,7 +45,7 @@ int main(int argc, char *argv[])
 	if(argc != 3)
 	{
 		qDebug() << "Usage: jsonapi-generator SOURCE_PATH OUTPUT_PATH";
-		return EINVAL;
+		return -EINVAL;
```
csoler, line 48: I'm surprised by the ngative value here. Is there some kind of hidden protocol to interpret that value elsewhere? If so, it could be documented.

libretroshare/src/jsonapi/jsonapi.cpp
```diff
@@ -41,6 +41,26 @@
 
 /*extern*/ JsonApiServer* jsonApiServer = nullptr;
 
+/*static*/ const std::multimap<std::string, std::string>
+JsonApiServer::corsHeaders =
+{
+    { "Access-Control-Allow-Origin", "*" },
```
csoler, line 47: does this line mean you can access the API from another IP than 127.0.0.1 ? If so, it's terribly insecure

libretroshare/src/retroshare/rsfiles.h
```diff
@@ -182,13 +183,22 @@ class FileTree
...
-    rstime_t ban_time_stamp;
+	std::string mFilename;
+	uint64_t mSize;
+	rstime_t mBanTimeStamp;
+
+	/// @see RsSerializable::serial_process
+	virtual void serial_process(RsGenericSerializer::SerializeJob j,
+	                            RsGenericSerializer::SerializeContext& ctx)
+	{
+		RS_SERIAL_PROCESS(mFilename);
```
csoler, line 198: This notation is very cmpact, but it has a drawback: the variables j and ctx cannot have a different name. Not really a problem, but if we document the serialization that will need to be said somewhere.

libretroshare/src/services/p3gxschannels.cc
```diff
@@ -1036,6 +1042,108 @@ bool p3GxsChannels::getChannelsContent(
 	return getPostData(token, posts, comments);
 }
 
+bool p3GxsChannels::createChannel(RsGxsChannelGroup& channel)
```
csoler, line 1045: The async function to do that is called createGroup(), as for all other GXS services. Naming it createChannel is of course more appropriate but it also hides the fact that the right function to call is the other one. I suggest to name this one createChannel_synchroneous() or somthing like this.

libretroshare/src/services/p3gxschannels.cc
```diff
@@ -1036,6 +1042,108 @@ bool p3GxsChannels::getChannelsContent(
...
+{
+	uint32_t token;
+	time_t beginCreation = time(nullptr);
+	if( !createGroup(token, channel)
+	        || waitToken(token) != RsTokenService::COMPLETE )
+		return false;
+	time_t endCreation = time(nullptr);
+
+	std::list<RsGroupMetaData> channels;
+	if(!getChannelsSummaries(channels)) return false;
```
csoler, line 1055: There is no guaranty that this call will return the newly created channel, since the wait() above only waits until the creation order is put into some queue.

libretroshare/src/services/p3gxschannels.cc
```diff
@@ -1036,6 +1042,108 @@ bool p3GxsChannels::getChannelsContent(
...
+	 * not that bad, moreover seems the last created group tend to end up near
+	 * the beginning of the list so it is fast founding it.
+	 * The shortcoming of this is that if groups with same data are created in
+	 * a burst (more then once in a second) is that the id of another similar
+	 * group can be returned, but this is a pointy case.
+	 * Order of conditions in the `if` matter for performances */
+	bool found = false;
+	for(const RsGroupMetaData& chan : channels)
+	{
+		if( IS_GROUP_ADMIN(chan.mSubscribeFlags)
```
csoler, line 1068: Indeed, that is really ugly. Another way to go would be to add the creation token into RsGroupMetaData, or use some kind of preliminary identifier, returned by createGroup() that can further be used to retrieve the correct channel.

libretroshare/src/services/p3gxschannels.cc
```diff
@@ -1036,6 +1042,108 @@ bool p3GxsChannels::getChannelsContent(
...
+	}
+
+#ifdef RS_DEEP_SEARCH
+	if(found) DeepSearch::indexChannelGroup(channel);
+#endif //  RS_DEEP_SEARCH
+
+	return found;
+}
+
+bool p3GxsChannels::createPost(RsGxsChannelPost& post)
```
csoler, line 1093: Same comments for this function.

libretroshare/src/services/p3gxschannels.cc
csoler: Besides if the channel creation fails for some reason, a function calling getChannelSummary() indefinitly would never return.

retroshare-gui/src/gui/GetStartedDialog.cpp
```diff
@@ -236,7 +236,7 @@ void GetStartedDialog::inviteFriends()
 	{
 		RsAutoUpdatePage::lockAllEvents();
 
-		cert = rsPeers->GetRetroshareInvite(false);
```
csoler, line 239: I put (false,false)  on purpose in the HomePage: to avoid leaking previous IPs unless the user really wants this. I think we should do the same everywhere. If I remember correctly, the false should only be removed on android because it's a better compromise on android to leak as many ips as possible.

retroshare-gui/src/gui/connect/ConnectFriendWizard.cpp
```diff
@@ -127,7 +127,7 @@ ConnectFriendWizard::ConnectFriendWizard(QWidget *parent) :
     body = GetStartedDialog::GetInviteText();
 	
     body += "\n" + GetStartedDialog::GetCutBelowText();
-    body += "\n\n" + QString::fromUtf8(rsPeers->GetRetroshareInvite(false).c_str());
```
csoler, line 130: Same comment here about not leaking all IPs -> (false,false)

retroshare-service/src/retroshare-service.cc
```diff
@@ -0,0 +1,78 @@
...
+	{
+		int argc2 = argc + 2;
+		char* argv2[argc2]; for (int i = 0; i < argc; ++i ) argv2[i] = argv[i];
+		char opt[] = "--jsonApiPort";
+		char val[] = "9092";
+		argv2[argc] = opt;
+		argv2[argc+1] = val;
+		RsInit::InitRetroShare(argc2, argv2, true);
+	}
+	else RsInit::InitRetroShare(argc, argv, true);
```
csoler, line 67: I presume it's not currently possible to run an auto-configured Tor node with this.

csoler (commented):
Congrats for this huge amount of work! I left a few comments.

jsonapi-generator/src/jsonapi-generator.cpp
G10h4ck: It is common practice to return -Enumber in case of error in executable

libretroshare/src/jsonapi/jsonapi.cpp
G10h4ck: No, it doesn't mean that, it means that if you access the API trough a browser it doesn't mind the domain name you have on your urlbar, so for example if an APP based on electron or other web framework uses RS API it could access even if they define their own alias of 127.0.0.1 to be more user friendly or use another port fro their own HTTP server on same machine. RetroShare JSON API bind on 127.0.0.1:9092 if it is not otherwise specified by command-line options, or by RetroShare GUI options.

libretroshare/src/retroshare/rsfiles.h
G10h4ck: If I remember well it is documented on the header where that macro is defined (anyway this has been introduced lot of time ago so it is not strictly related to this PR ;) )

libretroshare/src/services/p3gxschannels.cc
G10h4ck: I don't think one is wrong and the other is right, they offer different behavior, I kept the other with the original name because it have the asynchronous semantic of GXS style API like all other services. The documentation in the header clearly states that that function is blocking, also I would avoid the longer name.

libretroshare/src/services/p3gxschannels.cc
G10h4ck: Nope, I have dig into the GXS code and the token is updated correctly only once the channel is effectively created, published and stored into the database, so we get to this point only if the channel has been created successfully.

libretroshare/src/services/p3gxschannels.cc
G10h4ck: > Indeed, that is really ugly. Another way to go would be to add the creation token into RsGroupMetaData, or use some kind of preliminary identifier, returned by createGroup() that can further be used to retrieve the correct channel.

I have experimented various ways of doing that, the problem is that while we keep track of token and token status in RsGxsDataAccess also for channel creation like we do for data read request, RsGxsDataAccess cannot keep track of channel creation data and request because it is done into RsGenExchange so triyng to keep some data into RsGxsDataAccess about the channel a race condition between RsGxsDataAccess and RsGenExchange raise and the channel is never created (also RS need to be stopped by kill -9 in that case...)

The most elegant working solution I could think of after many experiments is to move the channel creation into RsGxsDataAccess and fully keep track of it there like we do with the other GxsRequests, but indeed attempting to to that caused a refactoring chain reaction, and I have no time for that right now and I think it should eventually done into it's own pull request

> use some kind of preliminary identifier

token is meant for that use but is not usable to retrieve data in this case because of what I explained before

G10h4ck: > Besides if the channel creation fails for some reason, a function calling getChannelSummary() indefinitly would never return.

Nope it would return false in that case quite fast, anyway in this context we would never call that function if channel creation failed because if `waitToken` return something different than `COMPLETE` `createChannel` return false before of getting there.

Also take in account that `waitToken` by default has a timeout of 500ms

retroshare-gui/src/gui/GetStartedDialog.cpp
G10h4ck: The false there was applied to `includeSignatures` as you can see here https://github.com/RetroShare/RetroShare/blob/18eee374f1b56ad41760c8ec17f2e90e6eb4717c/libretroshare/src/retroshare/rspeers.h#L557

In this PR I have unified the two versions of `GetRetroshareInvite` so this change is only to fix compilation error but doesn't change the semantic as thi sother function would be called now
https://github.com/RetroShare/RetroShare/pull/1363/files#diff-0a044481cf0cc5773dc30a3a6aeffee8R537
About not exporting the last recently used 10 IPv# i am not sure about how much it improves the privacy or just make connection more difficult, but I think as that is a semantic change should be part of another PR/discussion ;)

retroshare-service/src/retroshare-service.cc
G10h4ck: Isn't it supported by  `RsInit::InitRetroShare(...)` ?
This snippet of code only add `--jsonApiPort` if it is not already specified because JSON API is the only mean of interaction available on retroshare-service, but won't alter other command line options

PR #1366, opened 2018-10-07
Fix versioning script msys2

hunbernd:
Made the msys2 build scripts work with the new versioning system.

### DIFF
--- a/build_scripts/Windows-msys2/build-tor.bat
+++ b/build_scripts/Windows-msys2/build-tor.bat
@@ -9,7 +9,7 @@ call "%EnvPath%\env.bat"
 if errorlevel 1 goto error_env
 
 %cecho% info "Build %SourceName%"
-call "%~dp0build\build.bat" 32 release tor version autologin plugins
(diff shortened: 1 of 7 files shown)


PhenomRetroShare:
Could you call it in appveyor.yml ?

hunbernd:
Yes, but I think it is a bad idea.
Currently the scripts automatically update msys2.
In my experience msys2 packages are often broken, so the success of the build would depend on if msys2 is broken, not on RS is broken.

G10h4ck:
Anyway the appveyor build should works fine right now, one should be even able do download the compiled executable, @PhenomRetroShare did you tried it?
What is missing is to integrate it in RetroShare main repository pull requests, but i have no permissions to do that @csoler  while I have it activated on my personal fork so i get notifications if one of my PR fails to build on appveyor

G10h4ck:
Seems at same time thunder made similar modifications, can you take a look to see if it is better to solve de conflicts or close the PR ?

hunbernd:
It would be redundant

PR #1372, opened 2018-10-14
Fix warnings

PhenomRetroShare:
All warning listed in commits comments.

### DIFF
--- a/libretroshare/src/rsserver/rsloginhandler.cc
+++ b/libretroshare/src/rsserver/rsloginhandler.cc
@@ -31,7 +31,7 @@ bool RsLoginHandler::checkAndStoreSSLPasswdIntoGPGFile(
 	FILE *sslPassphraseFile = RsDirUtil::rs_fopen(
 	            getSSLPasswdFileName(ssl_id).c_str(), "r");
 
-	if(sslPassphraseFile != NULL)	// already have it.
+	if(sslPassphraseFile)	// already have it.
 	{
 		fclose(sslPassphraseFile);
 		return true ;
@@ -52,7 +52,7 @@ bool RsLoginHandler::getSSLPasswdFromGPGFile(const RsPeerId& ssl_id,std::string&
 	FILE *sslPassphraseFile = RsDirUtil::rs_fopen(
 	            getSSLPasswdFileName(ssl_id).c_str(), "r");
 
-	if (sslPassphraseFile == NULL)
+	if (!sslPassphraseFile)
 	{
 		std::cerr << "No password provided, and no sslPassphraseFile : "
 		          << getSSLPasswdFileName(ssl_id).c_str() << std::endl;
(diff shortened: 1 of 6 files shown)


libretroshare/src/rsserver/rsloginhandler.cc
```diff
@@ -31,7 +31,7 @@ bool RsLoginHandler::checkAndStoreSSLPasswdIntoGPGFile(
 	FILE *sslPassphraseFile = RsDirUtil::rs_fopen(
 	            getSSLPasswdFileName(ssl_id).c_str(), "r");
 
-	if(sslPassphraseFile != NULL)	// already have it.
+	if(sslPassphraseFile != nullptr)	// already have it.
```
G10h4ck: I think it is even leaner if changed into `if(sslPassphraseFile)` same goes for other `!= nullptr`

libretroshare/src/rsserver/rsloginhandler.cc
```diff
@@ -52,7 +52,7 @@ bool RsLoginHandler::getSSLPasswdFromGPGFile(const RsPeerId& ssl_id,std::string&
 	FILE *sslPassphraseFile = RsDirUtil::rs_fopen(
 	            getSSLPasswdFileName(ssl_id).c_str(), "r");
 
-	if (sslPassphraseFile == NULL)
+	if (sslPassphraseFile == nullptr)
```
G10h4ck: I think it is even leaner if changed into `if(!sslPassphraseFile)` same goes for other `== nullptr`

retroshare-gui/src/gui/RetroShareLink.cpp
```diff
@@ -730,10 +733,11 @@ void RetroShareLink::check()
 			if(!checkSSLId(_SSLid))
 				_valid = false;			// no break! We also test file stuff below.
 			/* fallthrough */
+			[[clang::fallthrough]];
```
G10h4ck, line 736: are `[[attributes]]` plainly ignored by older compilers?

retroshare-gui/src/gui/RetroShareLink.cpp
```diff
@@ -730,10 +733,11 @@ void RetroShareLink::check()
...
 		case TYPE_FILE_TREE:
 			/* fallthrough */
 		case TYPE_FILE:
-			if(_size > (((uint64_t)1)<<40))	// 1TB. Who has such large files?
+			if(_size > ((static_cast<uint64_t>(1))<<40))	// 1TB. Who has such large files?
```
G10h4ck, line 740: Reading this code I think RS should not have this kind of limit on file size, as an example think of a big disk image shared trought retroshare, taht could be bigger then 1TB of course, or large databases of LHC experimental data, @csoler what you think of it?

csoler: I like these limits because they detect a lot of problems. You can end up with arbitrary large files when e.g. deserializing some corrupted data. A very large size can also be the result of some kind of attack with a crafted RSLink. It's probably good to keep that kind of check. Of course the limit can be extended a bit if you want, but 1TB seems reasonnable to me.

libretroshare/src/rsserver/rsloginhandler.cc
PhenomRetroShare: Done

libretroshare/src/rsserver/rsloginhandler.cc
PhenomRetroShare: Done

retroshare-gui/src/gui/RetroShareLink.cpp
PhenomRetroShare: I don't know.
At least Travis get it...

PR #1378, opened 2018-10-20
V0.6 improved gui

csoler:


G10h4ck:
Didn't tested it, but seems ok, when deciding for default values in the GUI take in accound that IPv6 addresses are not shared if the third parameter is false ;)

PR #1382, opened 2018-10-24
Fix warnings

PhenomRetroShare:


libretroshare/src/util/rsdir.cc
```diff
@@ -327,14 +327,14 @@ bool RsDirUtil::copyFile(const std::string& source,const std::string& dest)
 #else
 	FILE *in = fopen64(source.c_str(),"rb") ;
 
-	if(in == NULL)
+	if(in == nullptr)
```
G10h4ck: This and similar should be replaced by `if(!in)`

G10h4ck (changes requested):

libretroshare/src/util/rsdir.cc
PhenomRetroShare: Done. Will be more concern now. ;)

libretroshare/src/ft/ftserver.cc
```diff
@@ -223,18 +223,18 @@ void ftServer::StopThreads()
 	mFtExtra->join();
 
 	delete (mFtDataplex);
```
G10h4ck, line 225: WTF are those?? those pointers are just nullptr as it is visible on the initializer list, those couple of lines should disappear and followinf similars too

G10h4ck (changes requested):

libretroshare/src/turtle/p3turtle.cc
```diff
@@ -92,11 +88,11 @@ static const uint32_t TURTLE_SEARCH_RESULT_MAX_HITS_DEFAULT= 100 ;		/// default
 
 static const float depth_peer_probability[7] = { 1.0f,0.99f,0.9f,0.7f,0.6f,0.5,0.4f } ;
 
-static const int TUNNEL_REQUEST_PACKET_SIZE             = 50 ;
+static const float TUNNEL_REQUEST_PACKET_SIZE           = 50.0f ;
```
G10h4ck: This doesn't seems like it should be a float, maybe the error is elsewhere where this constant is used

G10h4ck (changes requested):

libretroshare/src/ft/ftserver.cc
csoler: I'd be surprised these pointers are not set anywhere ;-) without them, there's no FT (especially mFtDataplex).

libretroshare/src/turtle/p3turtle.cc
csoler: agreed: we may compute some float value elsewhere that is compared to this const. Would be better to cast the other thing to int, or the const to float only when it is used.

libretroshare/src/ft/ftserver.cc
PhenomRetroShare: They are initialized in 133:
void ftServer::SetupFtServer()
and used on all code...
I don't understand your comment @G10h4ck.

libretroshare/src/turtle/p3turtle.cc
PhenomRetroShare: This const is used only with float values.
So why cast it to float and not defined it to float at start?

libretroshare/src/ft/ftserver.cc
G10h4ck: Sorry Github diff view fooled me those deletes seemed to be inside the constructor hence my WTF reaction

libretroshare/src/turtle/p3turtle.cc
G10h4ck: Then maybe the name of the constant is misleading? Calling it `*_SIZE` one expect it to be size in bytes or similar thing

csoler: it should be a number of bytes. Phenom's option is best: just cast it to float during comparison.

PhenomRetroShare: Perhaps I've explained badly.
Cast was done before (old style) during comparison. But as this value only used with float operators, why don't define it as float?
Look at [libretroshare/src/turtle/p3turtle.cc:1749](https://github.com/RetroShare/RetroShare/blob/master/libretroshare/src/turtle/p3turtle.cc#L1749)

G10h4ck: In this case I would rename it as `TUNNEL_REQUEST_PACKET_SIZE_APPROX` and keep it as float

PhenomRetroShare: Done.

libretroshare/src/ft/ftextralist.cc
```diff
@@ -111,7 +111,7 @@ void ftExtraList::hashAFile()
 #endif
 
 	/* hash it! */
-	std::string name, hash;
+	//std::string name, hash;
```
G10h4ck: better remove dead-code entirely ;)

libretroshare/src/ft/ftextralist.cc
```diff
@@ -466,16 +466,16 @@ bool    ftExtraList::loadList(std::list<RsItem *>& load)
 
 		/* open file */
 		FILE *fd = RsDirUtil::rs_fopen(fi->file.path.c_str(), "rb");
-		if (fd == NULL)
+		if (fd == nullptr)
```
G10h4ck: better `if(!fd)`

libretroshare/src/ft/ftserver.cc
```diff
@@ -1356,7 +1357,7 @@ bool ftServer::decryptItem(const RsTurtleGenericDataItem *encrypted_item,const R
 	decrypted_item = dynamic_cast<RsTurtleGenericTunnelItem*>(deserialise(data,&data_size)) ;
 	free(data);
 
-	return (decrypted_item != NULL);
+	return (decrypted_item);
```
G10h4ck: Are the parentheses needed?

libretroshare/src/ft/ftserver.cc
```diff
@@ -1556,7 +1559,7 @@ void ftServer::receiveTurtleData(const RsTurtleGenericTunnelItem *i,
 #endif
 			getMultiplexer()->recvData(virtual_peer_id,hash,0,item->chunk_offset,item->chunk_size,item->chunk_data) ;
 
-			const_cast<RsTurtleFileDataItem*>(item)->chunk_data = NULL ;	// this prevents deletion in the destructor of RsFileDataItem, because data will be deleted
```
G10h4ck, line 1559: @csoler Shouldn't the method signature changed removing `const` instead of using the canst in this case? Or it causes a refactor chain reaction?

libretroshare/src/gxs/rsgxsnetservice.cc
```diff
@@ -3801,7 +3809,7 @@ bool RsGxsNetService::processTransactionForDecryption(NxsTransaction *tr)
     GXSNETDEBUG_P_(peerId) << "RsGxsNetService::decryptTransaction()" << std::endl;
 #endif
 
-    std::list<RsNxsItem*> decrypted_items ;
+	//std::list<RsNxsItem*> decrypted_items ;
```
G10h4ck: Better removing dead-code completely

libretroshare/src/gxs/rsgxsnetservice.cc
```diff
@@ -3810,7 +3818,7 @@ bool RsGxsNetService::processTransactionForDecryption(NxsTransaction *tr)
     {
         RsNxsEncryptedDataItem *encrypted_item = dynamic_cast<RsNxsEncryptedDataItem*>(*it) ;
 
-        if(encrypted_item == NULL)
+        if(encrypted_item == nullptr)
```
G10h4ck: better `if(!encrypted_item)`

libretroshare/src/gxs/rsgxsutil.cc
```diff
@@ -147,8 +147,8 @@ bool RsGxsMessageCleanUp::clean()
 
 RsGxsIntegrityCheck::RsGxsIntegrityCheck(
         RsGeneralDataService* const dataService, RsGenExchange* genex,
-        RsSerialType& serializer, RsGixs* gixs ) :
-    mDs(dataService), mGenExchangeClient(genex), mSerializer(serializer),
+        RsSerialType& /*serializer*/, RsGixs* gixs ) :
+    mDs(dataService), mGenExchangeClient(genex),/* mSerializer(serializer),*/
```
G10h4ck: Are you sure this won't break deep search? (it is not compiled by default but will set as default soon)

libretroshare/src/gxs/rsgxsutil.h
```diff
@@ -214,7 +213,7 @@ class RsGxsIntegrityCheck : public RsSingleJobThread
 
 	RsGeneralDataService* const mDs;
 	RsGenExchange *mGenExchangeClient;
-	RsSerialType& mSerializer;
+	//RsSerialType& mSerializer;
```
G10h4ck: This breaks deep search, consider putting it and related changes inside and `#ifdef` instead

libretroshare/src/services/p3gxschannels.h
```diff
@@ -24,11 +24,11 @@
 #define P3_GXSCHANNELS_SERVICE_HEADER
```
G10h4ck, line 24: Editing this file will probably give a bunch of merging conflicts with safer gxschannel API PR

libretroshare/src/services/p3gxscircles.cc
```diff
@@ -963,9 +963,9 @@ bool p3GxsCircles::cache_request_load(const RsGxsCircleId &id)
 	int32_t age = 0;
 	if (RsTickEvent::prev_event_ago(CIRCLE_EVENT_CACHELOAD, age))
 	{
-		if (age < MIN_CIRCLE_LOAD_GAP)
+		if( (age < MIN_CIRCLE_LOAD_GAP) && (age > 0))
```
G10h4ck: @csoler I think that if we get age < 0 at this point at least a warning must be printed

libretroshare/src/tcponudp/tcpstream.cc
```diff
@@ -2648,7 +2667,7 @@ int	setupBinaryCheck(std::string fname)
 /* uses seq number to track position -> ensure no rollover */
 int	checkData(uint8 *data, int size, int idx)
 {
-	if (bc_fd <= 0)
+	if (bc_fd = 0)
```
G10h4ck: This assignment here seems a very dangerous typo!!

libretroshare/src/turtle/p3turtle.cc
```diff
@@ -393,7 +388,7 @@ void p3turtle::manageTunnels()
 
 			static const float grow_speed = 1.0f ;	// speed at which the time increases.
 
-			float tunnel_keeping_factor = (std::max(1.0f,(float)total_speed/(float)(50*1024)) - 1.0f)*grow_speed + 1.0f ;
+			float tunnel_keeping_factor = (std::max(1.0f,static_cast<float>(total_speed)/(50.0f*1024)) - 1.0f)*grow_speed + 1.0f ;
```
G10h4ck: @csoler it's just a coincidence it has same value as `TUNNEL_REQUEST_PACKET_SIZE_APPROX` ?

libretroshare/src/turtle/p3turtle.cc
```diff
@@ -2208,11 +2205,12 @@ static std::string printNumber(uint64_t num,bool hex=false)
 	if(hex)
 	{
 		char tmp[100] ;
+		uint64_t one = 1;
```
G10h4ck: Compiler might figure out anyway but better to specify `constexpr` anyway

retroshare-gui/src/TorControl/TorManager.cpp
```diff
@@ -150,7 +152,7 @@ void TorManager::setHiddenServiceDirectory(const QString &path)
 
 bool TorManager::setupHiddenService()
 {
-	if(d->hiddenService != NULL)
+	if(d->hiddenService != nullptr)
```
G10h4ck: Better `if(d->hiddenService)`

retroshare-gui/src/gui/FileTransfer/TransfersDialog.cpp
```diff
@@ -141,15 +141,15 @@ class RsDownloadListModel : public QAbstractItemModel
...
 #endif
-		return mDownloads[entry].peers.size();
+		return static_cast<int>(mDownloads[entry].peers.size());
 	}
-	int columnCount(const QModelIndex &parent = QModelIndex()) const
+	int columnCount(const QModelIndex &/*parent*/ = QModelIndex()) const
 	{
 		return COLUMN_COUNT ;
 	}
 	bool hasChildren(const QModelIndex &parent = QModelIndex()) const
```
G10h4ck, line 150: It is usually better to expose into header instead of that in the implementation that a function has default parameter value

retroshare-gui/src/gui/FileTransfer/TransfersDialog.cpp
```diff
@@ -180,15 +180,15 @@ class RsDownloadListModel : public QAbstractItemModel
...
 		uint32_t entry = 0;
 		int source_id=0 ;
 
 		if(!parent_ref)	// top level. The entry is that of a transfer
 		{
-			void *ref = NULL ;
+			void *ref = nullptr ;
 
-			if(row >= (int)mDownloads.size() || !convertTabEntryToRefPointer(row,-1,ref))
+			if(row >= static_cast<int>(mDownloads.size()) || !convertTabEntryToRefPointer(static_cast<uint32_t>(row),-1,ref))
```
G10h4ck, line 191: @csoler can you confirm this is ok? It reminds me of the 64/32 bit file node reference depending on architecture

retroshare-gui/src/gui/FileTransfer/TransfersDialog.cpp
```diff
@@ -396,29 +428,28 @@ class RsDownloadListModel : public QAbstractItemModel
...
-					priority = 0;
+					priority = PRIORITY_NULL;
 				else
 					switch (fileInfo.priority)
 					{
 					case SPEED_LOW:     priority = PRIORITY_SLOWER; break;
 					case SPEED_NORMAL:  priority = PRIORITY_AVERAGE; break;
 					case SPEED_HIGH:    priority = PRIORITY_FASTER; break;
-					default:            priority = PRIORITY_AVERAGE; break;
+					//default:            priority = PRIORITY_AVERAGE; break;
```
G10h4ck, line 445: this probably will introduce a warning of missing default case on some compilers

retroshare-gui/src/gui/FileTransfer/TransfersDialog.cpp
```diff
@@ -1109,6 +1114,7 @@ void TransfersDialog::activatePage(TransfersDialog::Page page)
 													break ;
 		case TransfersDialog::RemoteSharedFilesTab: ui.tabWidget->setCurrentWidget(remoteSharedFiles) ;
 													break ;
+		case TransfersDialog::DownloadTab: assert(false);
```
G10h4ck: won't this just crash retroshare in that case? Better printing some information about errors/stacktrace

G10h4ck (changes requested):
Next time avoid doing pull requests so huge, it take too much time reviewing them and keeping high level of attention it's hard! Also when you do warning fixing  try to do one pull request for king of warning, for example a pull request where you only change NULL into nullptr  and then another pull request where you static_cast some integers types, and another where you change the if to check directly the pointers without comparing are much easier to review and less error prone then a pull request with the three things all toghether

libretroshare/src/ft/ftserver.cc
PhenomRetroShare: No, just to see that we don't return object but if pointer is not null.

retroshare-gui/src/gui/FileTransfer/TransfersDialog.cpp
PhenomRetroShare: The class is defined here, no header. ;)

retroshare-gui/src/gui/FileTransfer/TransfersDialog.cpp
PhenomRetroShare: int row
static bool convertTabEntryToRefPointer(uint32_t entry,int source_id,void *& ref)

retroshare-gui/src/gui/FileTransfer/TransfersDialog.cpp
PhenomRetroShare: This is to remove once:
    /retroshare-gui/src/gui/FileTransfer/TransfersDialog.cpp:414: warning:
    default label in switch which covers all enumeration values

G10h4ck: ok!

libretroshare/src/ft/ftextralist.cc
PhenomRetroShare: Done

libretroshare/src/ft/ftextralist.cc
PhenomRetroShare: Done

libretroshare/src/gxs/rsgxsnetservice.cc
PhenomRetroShare: Done

libretroshare/src/gxs/rsgxsnetservice.cc
PhenomRetroShare: Done

libretroshare/src/gxs/rsgxsutil.cc
PhenomRetroShare: Done, I don't understand how I can make it possible. Sorry. ;)

libretroshare/src/gxs/rsgxsutil.h
PhenomRetroShare: Done

libretroshare/src/services/p3gxscircles.cc
PhenomRetroShare: Done

libretroshare/src/tcponudp/tcpstream.cc
PhenomRetroShare: Shame on  me...

libretroshare/src/turtle/p3turtle.cc
PhenomRetroShare: I think it isn't.
Replaced

libretroshare/src/turtle/p3turtle.cc
PhenomRetroShare: Done

retroshare-gui/src/TorControl/TorManager.cpp
PhenomRetroShare: Done

retroshare-gui/src/gui/FileTransfer/TransfersDialog.cpp
PhenomRetroShare: Found better: Show DownloadTab ;)

libretroshare/src/ft/ftextralist.cc
G10h4ck: It still appears as `if (fd == nullptr)` in the changes, is just GH that is outdated?

retroshare-gui/src/gui/FileTransfer/TransfersDialog.cpp
csoler: sorry but I really dont get it. The Default is needed here.I dont see why the compiler would complain. what if priority has another value than the possible values? We always need to make sure that kind of errors are handled.

libretroshare/src/tcponudp/tcpstream.cc
```diff
@@ -2648,7 +2667,7 @@ int	setupBinaryCheck(std::string fname)
 /* uses seq number to track position -> ensure no rollover */
 int	checkData(uint8 *data, int size, int idx)
 {
-	if (bc_fd <= 0)
+	if (bc_fd == 0)
```
G10h4ck: despite the misleading name `bc_fd` is a pointer so check it as a pointer and avoid arithmetic

`if(!bc_fd)`

G10h4ck (changes requested):

libretroshare/src/gxs/rsgxsutil.cc
G10h4ck: It' because you are compiling without `CONFIG+=rs_deep_search`

libretroshare/src/tcponudp/tcpstream.cc
PhenomRetroShare: Done.
But new commits was waiting so they are added before you req.

retroshare-gui/src/gui/FileTransfer/TransfersDialog.cpp
PhenomRetroShare: @csoler you'll get a warning that not all enum are used.
Default value is set in initialization.
double priority = PRIORITY_AVERAGE;

libretroshare/src/gxs/rsgxsutil.cc
PhenomRetroShare: CppCheck checks all this case normally.
And I generally check were removed values are used by searching by text in all trunk. Perhaps I was tired. ;)

libretroshare/src/turtle/p3turtle.cc
G10h4ck: ping @csoler

libretroshare/src/ft/ftserver.cc
```diff
@@ -1356,7 +1357,7 @@ bool ftServer::decryptItem(const RsTurtleGenericDataItem *encrypted_item,const R
 	decrypted_item = dynamic_cast<RsTurtleGenericTunnelItem*>(deserialise(data,&data_size)) ;
 	free(data);
 
-	return (decrypted_item != NULL);
```
csoler, line 1359: This is the limit to what I can accept: the test returns a bool as an explicit calculation. If you return the object itself, it is implicitely converted to a bool, which is much less clear and prone to wrong interpretation (not to mention the syntax with () that are not needed). So I would definitely keep the if(decrypted_item != NULL) here.

libretroshare/src/ft/ftserver.cc
```diff
@@ -1481,16 +1482,18 @@ TurtleSearchRequestId ftServer::turtleSearch(const RsRegularExpression::Lineariz
     return mTurtleRouter->turtleSearch(expr) ;
 }
 
-#warning we should do this here, but for now it is done by turtle router.
+#ifndef RS_NO_WARN_CPP
+#pragma message("csoler 2018-06-10: We should do this here, but for now it is done by turtle router.")
+#endif
 //   // Dont delete the item. The client (p3turtle) is doing it after calling this.
 //   //
-//   void ftServer::receiveSearchResult(RsTurtleSearchResultItem *item)
```
csoler, line 1487: How can the compiler generate warnings in a section that is a comment ???

libretroshare/src/ft/ftserver.cc
csoler: Definitely a chain reaction ;-) I would keep it as it is.

libretroshare/src/ft/ftserver.cc
```diff
@@ -1724,9 +1727,9 @@ int ftServer::handleIncoming()
 	// now File Input.
 	int nhandled = 0 ;
 
-	RsItem *item = NULL ;
+	RsItem *item = nullptr ;
 
-	while(NULL != (item = recvItem()))
+	while( (item = recvItem()) )
```
csoler: Well, that is also reported by my compiler as a possible mistake, and I fully agree with him. Please keep the NULL (or change it to nullptr if you need)

libretroshare/src/ft/ftserver.cc
```diff
@@ -1827,7 +1830,7 @@ int ftServer::handleIncoming()
  **********************************
  *********************************/
 
-void ftServer::receiveSearchResult(RsTurtleFTSearchResultItem *item)
```
csoler, line 1830: why this change of names? This function overloads the method in TurtleClientService. So it should not be changed! If you do, FT will just not work anymore.

libretroshare/src/gxs/rsgxsnetservice.cc
```diff
@@ -714,8 +714,8 @@ void RsGxsNetService::syncWithPeers()
 
             // The last post will be set to TS 0 if the req delay is 0, which means "Indefinitly"
 
-            if(req_delay > 0)
-				msg->createdSinceTS = std::max(0,(int)time(NULL) - req_delay);
```
csoler, line 718: (int)something  *is* a static cast. So why convert all the casts in the source into explicit static<cast>(int)? This makes the whole code much less readable.

libretroshare/src/gxs/rsgxsnetservice.cc
```diff
@@ -747,12 +747,12 @@ void RsGxsNetService::generic_sendItem(RsNxsItem *si)
 
 	RsGxsGroupId tmp_grpId;
 
-	if(mAllowDistSync && mGxsNetTunnel != NULL && mGxsNetTunnel->isDistantPeer( static_cast<RsGxsNetTunnelVirtualPeerId>(si->PeerId()),tmp_grpId))
```
csoler, line 750: when I read this, it reminds me that mGxsNetTunnel is a pointer. That is why it is compared to a pointer. Implicily casting to bool for testing is far worse IMHO.

libretroshare/src/gxs/rsgxsnetservice.cc
```diff
@@ -877,7 +879,7 @@ void RsGxsNetService::syncGrpStatistics()
 			for(uint32_t i=0;i<n;++i)
 				++rit ;
 
-			for(uint32_t i=0;i<std::min(rec.suppliers.ids.size(),(size_t)GROUP_STATS_UPDATE_NB_PEERS);++i)
+			for(size_t i=0;i<std::min(rec.suppliers.ids.size(),GROUP_STATS_UPDATE_NB_PEERS);++i)
```
csoler, line 882: I prefer the previous version: .size() is a size_t, so the int GROUP_STATS... should be cast to size_t. So in summary: keep GROUP_STATS_... to be an int. Keep the size_t in the for()

libretroshare/src/gxs/rsgxsnetservice.cc
```diff
@@ -960,7 +962,7 @@ void RsGxsNetService::handleRecvSyncGrpStatistics(RsNxsSyncGrpStatsItem *grs)
 
 	    RsNxsSyncGrpStatsItem *grs_resp = new RsNxsSyncGrpStatsItem(mServType) ;
 	    grs_resp->request_type = RsNxsSyncGrpStatsItem::GROUP_INFO_TYPE_RESPONSE ;
-	    grs_resp->number_of_posts = vec.size();
+	    grs_resp->number_of_posts = static_cast<uint32_t>(vec.size());
```
csoler, line 965: is that really necessary?

csoler:
ok. I've read 20% of this PR. As much as I appreciate contributions to clean and improve the code, I think that they are generic problems here:
- systematically using explicit static_cast<>() instead of (type)something makes the code much less readable. 
- using implicit cast of pointers to bool in comparisons, and especially in boolean return of functions makes the code much less readable
- fixing "warnings" in parts of the code that are commented out does not make sense to me.
- NULL is valid but outdated. So we use nullptr in new code, but changing all NULLs into nullptr looks like a significant waste of time IMHO.

I really do not understand why you spend so much time making these changes. Their non critical nature does not compare favorably with the effort that they need devs proofread them, and the very high risk of subtle errors they can bring in the code  (not because its you who do them, but rather because of their nature and repetitive occurrence).

libretroshare/src/ft/ftserver.cc
PhenomRetroShare: I'm OK with that.

libretroshare/src/ft/ftserver.cc
PhenomRetroShare: This code i commented and could be uncommented. That's why I change it too.

libretroshare/src/ft/ftserver.cc
PhenomRetroShare: OK

libretroshare/src/ft/ftserver.cc
PhenomRetroShare: Looks at warning:
/libretroshare/src/ft/ftserver.h:101: warning:
'ftServer::receiveSearchResult' hides overloaded virtual function [-
Woverloaded-virtual]
/libretroshare/src/turtle/turtleclientservice.h:122: hidden overloaded
virtual function 'RsTurtleClientService::receiveSearchResult' declared
here: different number of parameters (3 vs 1)

libretroshare/src/gxs/rsgxsnetservice.cc
PhenomRetroShare: /libretroshare/src/gxs/rsgxsnetservice.cc:718: warning: use of old-style
cast
/libretroshare/src/gxs/rsgxsnetservice.cc:718: warning: implicit
conversion changes signedness: 'const int' to 'uint32_t' (aka 'unsigned
int')

libretroshare/src/gxs/rsgxsnetservice.cc
PhenomRetroShare: \<Joke>Could be rename to mpGxsNetTunnel and mbAllowDistSync.\</joke> ;)
And we use it just after.
@G10h4ck Need to be clarified in Coding style.

libretroshare/src/gxs/rsgxsnetservice.cc
PhenomRetroShare: GROUP_STATS_UPDATE_NB_PEERS is only used here.

libretroshare/src/gxs/rsgxsnetservice.cc
PhenomRetroShare: You're right, number_of_posts should be size_t. ;)
/libretroshare/src/gxs/rsgxsnetservice.cc:963: warning: implicit
conversion loses integer precision: 'std::vector::size_type' (aka
'unsigned long') to 'uint32_t' (aka 'unsigned int')
size_t depends on system.

libretroshare/src/ft/ftserver.cc
PhenomRetroShare: This is a relevant warning...

PhenomRetroShare:
>     * systematically using explicit static_cast<>() instead of (type)something makes the code much less readable.
warning: use of old-style cast

>     * using implicit cast of pointers to bool in comparisons, and especially in boolean return of functions makes the code much less readable
I'm agreed with that.

>     * NULL is valid but outdated. So we use nullptr in new code, but changing all NULLs into nullptr looks like a significant waste of time IMHO.
Only changed in modified file.

> I really do not understand why you spend so much time making these changes.

Only to spent less time to see if what I'll code add new warnings.
And to see relevant ones clearly.
More than 60000 in trunk.
scrollbar
For only one file...
That's make a long time we don't make a coverity scan...

libretroshare/src/ft/ftserver.cc
csoler: yes I agree. The very weird thing here is that I'm pretty sure to have fixed this already, but apparently I did not. Gio told me the exact same thing weeks ago, and I thought I fixed it.
So: yes, change the same to ftServerReceiveSearchResults() and call this function in p3turtle.cc:1223  instead of             client->receiveSearchResult(ftsr);

G10h4ck:
I think cleaning up the code make sense, but it must be done in little PR so we can manage them, this PR is too huge to keep track of reviewing... @PhenomRetroShare to avoid your work to get lost, would you like to split this in little PRs that we can more reasonably review and agree on?

csoler:
If you do so, please:
- don't change all NULL into nullptr. It's a #define, so there's no point in doing that
- don't implicitly cast pointers to bool in return of functions. 
- don't replace c-style static casts into c++ static_cast<>. It's the same, and it makes the code much harder to read.

G10h4ck:
> don't replace c-style static casts into c++ static_cast<>. It's the same, and it makes the code much harder to read.

This is problematic because C style cast may cause any of reinterpret, const, dynamic, or static cast and even some of them together unintended, in addiction to that not only cpp-check but also compilers reports that as problem and getting many of those warnings make it difficult to find important ones, so we need to use explicit casting in new code and think as a way to port the old code to the C++ style cast without making it unreadable or the PR impossible to review

PhenomRetroShare:
It's possible to cherry-pick commit by commit...
I cannot make a PR for each modified files.
I don't see a better manner to do it.
And we have to do it before running a coverity scan.
Once done, no need to do it later... or at least they will be very little.

G10h4ck:
Yeah no need of a PR for each file, but it must be reasonable to review the code and maximum ten minutes for this kind (fix compiler warning, style improvements...) of PR

PhenomRetroShare:
Cherry-pick by commits so...

libretroshare/src/ft/ftserver.cc
PhenomRetroShare: OK, but I test if it is a ftServer before.

PR #1383, opened 2018-10-28
WIP: Feature/mailto

hunbernd:
Made possible to create mailto like links

### DIFF
--- a/retroshare-gui/src/gui/Identity/IdDialog.cpp
+++ b/retroshare-gui/src/gui/Identity/IdDialog.cpp
@@ -2462,7 +2462,10 @@ void IdDialog::IdListCustomPopupMenu( QPoint )
 			}
 
 			if (n_selected_items==1)
+			{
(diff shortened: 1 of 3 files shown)


csoler:
I'm not sure that this is functional yet: when you click on the link, does it open a msg box with the proper destination already set? That would be the case if the code was here already. I havn't checked.

hunbernd:
It opens the message composer dialog, with destination set, and optionally a title if you specify one.

It is already used in forums: right click on a message --> Reply to author with a private message.

csoler:
sorry for taking so long to review this. It works, except from one detail: the created link text is "send a new mail to Unknown (some id)". I think there's a bg somewhere when requesting the name when you create the link

defnax:
@hunbernd we need something a rs profile link.

when clicking on a nickname open the profile page of the person.
some time better use profile link and sometime message link.

i

defnax:
any news about this? @csoler @hunbernd

PR #1384, opened 2018-10-28
Removed outdated build scripts and information

hunbernd:


G10h4ck (approved):
Seems a nice cleanup to me, let's wait a bit for @csoler  and @thunder2 feedback

csoler:
I'm ok with it too. I'm just not sure about the .bat. Maybe they are used by thunder?

thunder2:
For me it's ok.
I only use files in build_scripts/Windows and build_scripts/Windows-msys2.

csoler:
thx all for the confirmation. I'm merging this.

PR #1385, opened 2018-10-29
0.6.5 Safer rsGxsChannel API

G10h4ck:
Protect data members with mutexes
Print messages when errors happens

### DIFF
--- a/libretroshare/src/retroshare/rsgxschannels.h
+++ b/libretroshare/src/retroshare/rsgxschannels.h
@@ -32,6 +32,7 @@
 #include "retroshare/rsgxscommon.h"
 #include "serialiser/rsserializable.h"
 #include "retroshare/rsturtle.h"
+#include "util/rsdeprecate.h"
 
 class RsGxsChannels;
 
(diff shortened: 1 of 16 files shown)


libretroshare/src/retroshare/rsgxschannels.h
```diff
@@ -138,180 +199,186 @@ class RsGxsChannels: public RsGxsIfaceHelper, public RsGxsCommentService
...
 	/**
 	 * @brief Share channel publishing key
 	 * This can be used to authorize other peers to post on the channel
-	 * @jsonapi{development}
+	 * @deprecated
 	 * @param[in] groupId Channel id
 	 * @param[in] peers peers to which share the key
 	 * @return false on error, true otherwise
 	 */
+	RS_DEPRECATED_FOR(shareChannelKeys)
```
PhenomRetroShare, line 311: When it is deprecated, is it possible to say how to do now to replace it?

G10h4ck: the `_FOR(shareChannelKeys)` is a suggestion on how to replace it ;)

libretroshare/src/retroshare/rsgxschannels.h
```diff
@@ -138,180 +199,186 @@ class RsGxsChannels: public RsGxsIfaceHelper, public RsGxsCommentService
...
 	 * @brief Request post creation.
 	 * The action is performed asyncronously, so it could fail in a subsequent
 	 * phase even after returning true.
+	 * @deprecated
 	 * @param[out] token Storage for RsTokenService token to track request
 	 * status.
 	 * @param[in] post
 	 * @return false on error, true otherwise
 	 */
+	RS_DEPRECATED
```
PhenomRetroShare, line 353: No _FOR here?

G10h4ck: because the right version to use has the same name

PhenomRetroShare: But, is it possible to get class handler too?
I see that for RsGxsIfaceHelper::getTokenService() which is deprecated for RsGenExchange::getTokenService(), but not easy to know...
One more thing, in /libretroshare/src/util/rsdeprecate.h, you don't take care of clang. It's why I don't have message FOR. ;)

G10h4ck: What is inside the parentheses of `RS_DEPRECATED_FOR()` get directly passed trough per-processor stringizer operator `#` and then printed, so you probably can put almost anything within those parentheses (test it please), about taking care of clang I haven't a comfortable clang based tool-chain to test it so please make a PR to take care of it ;)

PhenomRetroShare: Where did you found __GNUC__ value starting using __deprecated__(message)?
Because here, with Clang v7.0.0, __GNUC__ == 4 && __GNUC_MINOR__ == 2 and message works:
/libretroshare/src/services/p3gxscircles.cc:1636: warning: 'getTokenService' is deprecated: Use 'RsGenExchange::getTokenService()' instead [-Wdeprecated-declarations]
GCC can handle it since 4.9:
https://gcc.gnu.org/gcc-4.9/changes.html#cxx

G10h4ck: Don't remember at all, anyway take in account that `GNUC`, ` GNUC_MINOR` ans similar defined by clang are mostly meaningless, clang has it's own way to check for version, and probably define those macro too for some kind of retro-compatibility, but we must not rely on those when compiling with clang.

G10h4ck: also the page you linked talks about  C++1y deprecated attribute which most probably is not supported when compiling in C++11 mode (which is how we compile)

felisucoibi:
could be nice the same for forums.

PR #1388, opened 2018-11-01
V0.6.5

csoler:
Valgrind work! Please take a look. This small fixes are always prone to making more errors.


retroshare-gui/src/gui/HomePage.cpp
```diff
@@ -50,9 +50,11 @@ HomePage::HomePage(QWidget *parent) :
 	MainPage(parent),
 	ui(new Ui::HomePage)
 {
+    mIncludeAllIPs = false;
```
G10h4ck: this should be moved in the initializers list, like ui(...)

csoler: done

PR #1390, opened 2018-11-01
JSON API to check if PGP id is friend

G10h4ck:


PhenomRetroShare:
Same as #1389

G10h4ck:
> Same as #1389

Same code but different branch, one is against master while this is against v0.6.5 ;)

PR #1394, opened 2018-11-05
Add a build-time switch to GXS to distributes all requested data

sehraf:
The problem:
------------------
When everybody subscribes to a channel/forum with a sync period of **one week** and a store period of **one year**, a new user can only retrieve posts within a week ago but not from a year ago even if that information is available.

The fix:
----------
This option lets GXS send all available (requested) data independent from the *sync period* setting. As a compromise the default is to send everything available to fix the actual problem, while the option allows someone who is a afraid of increased traffic to change back to the old behavior.

### DIFF
--- a/retroshare.pri
+++ b/retroshare.pri
@@ -111,6 +111,11 @@ rs_onlyhiddennode:CONFIG -= no_rs_onlyhiddennode
 CONFIG *= rs_gxs
 no_rs_gxs:CONFIG -= rs_gxs
 
+# To disable GXS distrubuting all available posts independed of the "sync" settings append the following
+# assignation to qmake command line "CONFIG+=no_rs_gxs_sendAll"
+CONFIG *= rs_gxs_sendAll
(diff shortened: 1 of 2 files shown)


retroshare.pri
```diff
@@ -111,6 +111,11 @@ rs_onlyhiddennode:CONFIG -= no_rs_onlyhiddennode
 CONFIG *= rs_gxs
 no_rs_gxs:CONFIG -= rs_gxs
 
+# To disable GXS distrubuting all available posts independed of the "sync" settings append the following
+# assignation to qmake command line "CONFIG+=no_rs_gxs_sendAll"
+CONFIG *= rs_gxs_sendAll
```
G10h4ck: better naming it rs_gxs_send_all

G10h4ck (changes requested):

retroshare.pri
sehraf: done, thx.

G10h4ck (approved):

PR #1400, opened 2018-11-08
add missing DEFINE, fix small bugs

sehraf:
If anybody knows the purpose of the `break;` and things that it should not be removed, let me know. Otherwise it causes unreachable code.

I got a bit lost with all this `pri` files. If there is a better place to add this define, let me know, too.

### DIFF
--- a/retroshare-gui/src/gui/statistics/DhtWindow.cpp
+++ b/retroshare-gui/src/gui/statistics/DhtWindow.cpp
@@ -288,7 +288,7 @@ void DhtWindow::updateNetPeers()
 	rsDht->getNetPeerList(peerIds);
 
 	/* collate peer stats */
-	int nPeers = peerIds.size();
+	size_t nPeers = peerIds.size();
 
 	// from DHT peers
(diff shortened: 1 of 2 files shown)


libretroshare/src/use_libretroshare.pri
```diff
@@ -28,6 +28,7 @@ equals(TARGET, retroshare):equals(TEMPLATE, lib){
 
 bitdht {
     !include("../../libbitdht/src/use_libbitdht.pri"):error("Including")
+    DEFINES *= RS_USE_BITDHT
```
G10h4ck: this should go in retroshare.pri, for example just before
```
rs_async_chat {
    DEFINES *= RS_ASYNC_CHAT
}
```

G10h4ck (changes requested):

libretroshare/src/use_libretroshare.pri
sehraf: done (force pushed)

PR #1413, opened 2018-11-20
0.6.5 Disable deprecated messaging

G10h4ck:
Now that by default own signed GXS identity is spread as gossip to direct friends, direct node chat is not needed anymore.
Improve usability by avoiding user confusion with 4 different type of chats of which 2 are mostly not useful, live only chat rooms and distant chat.
As a bonus now when one double click on an identity on people tab a chat is opened.

### DIFF
--- a/libretroshare/src/chat/p3chatservice.cc
+++ b/libretroshare/src/chat/p3chatservice.cc
@@ -848,9 +848,18 @@ bool p3ChatService::handleRecvChatMsgItem(RsChatMsgItem *& ci)
             RsServer::notify()->AddPopupMessage(popupChatFlag, ci->PeerId().toStdString(), name, message); /* notify private chat message */
         else
         {
-            /* notify public chat message */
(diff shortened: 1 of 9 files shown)


zapek:
What happens for the following case?
- I exchange certificates with a friend and he shows up as a node, we connected successfully
- I want to message him but his identity's nickname is different than his node's nickname

This is a bad idea IMHO. While every node has a default identity, there's nothing that tells which one it is.

zapek:
Another thing, how do I quickly inform all my nodes that I'm going offline because of computer maintenance or holidays or whatever? Before, I would just write it into the "Boradcast" field, now I need to find out all the default identities used by my nodes and send a mail.

zapek:
Ok, last one. How do I chat with a friend directly?

It seems I have to create a private chatroom then invite him here.

Where's the progress in the usability? Before, you just double clicked in the friend node and you could chat. It was the perfect instant messaging replacement. All the friends (even non computer savvy ones) I introduced to Retroshare were able to chat that way. But creating a private chatroom just for that is unnecessary.

PR #1415, opened 2018-11-20
V0.6.5 forum model2

csoler:


felisucoibi:
+1000

libretroshare/src/retroshare/rsgxsforums.h
```diff
@@ -150,6 +150,19 @@ class RsGxsForums: public RsGxsIfaceHelper
...
+	 * @brief Get specific list of messages from a single forums. Blocking API
+	 * @jsonapi{development}
+	 * @param[in] forumId id of the forum of which the content is requested
+	 * @param[in] msgs_to_request list of message ids to request
+	 * @param[out] msgs storage for the forum messages
+	 * @return false if something failed, true otherwhise
+	 */
+	virtual bool getForumsContent(
+            const RsGxsGroupId& forumId,
+            std::set<RsGxsMessageId>& msgs_to_request,
```
G10h4ck, line 178: please use camel case for param name to be consistent with the other JSON API calls, also in JSON API we can't expose different methods with same name, so this need to have a different name too

retroshare-gui/src/gui/gxsforums/GxsForumThreadWidget.cpp
```diff
@@ -2542,26 +2543,129 @@ bool GxsForumThreadWidget::filterItem(QTreeWidgetItem *item, const QString &text
...
+        RsGxsForumGroup group = groups[0];
+
+        // 3 - update the model in the UI thread.
+
+        RsQThreadUtils::postToObject( [group,this]()
+		{
+			/* Here it goes any code you want to be executed on the Qt Gui
+			 * thread, for example to update the data model with new information
+			 * after a blocking call to RetroShare API complete, note that
+			 * Qt::QueuedConnection is important!
```
G10h4ck, line 1931: The QueuedConnection part of the comment is not relevant anymore, so it should be removed ;)

G10h4ck (changes requested):

retroshare-gui/src/gui/gxsforums/GxsForumThreadWidget.cpp
```diff
@@ -2424,40 +2431,60 @@ void GxsForumThreadWidget::flagperson()
...
+
+        msgs_to_request.insert(mThreadId);
+
+		if(!rsGxsForums->getForumsContent(groupId(),msgs_to_request,msgs))
+		{
+			std::cerr << __PRETTY_FUNCTION__ << " failed to retrieve forum group info for forum " << groupId() << std::endl;
+			return;
+        }
+
+        if(msgs.size() != 1)
```
G10h4ck, line 1662: This seems an error condition, so a message should be printed ;)

csoler:
3 things left: sorting (and post pinning), post versions, proper fonts for read/unread to match old system. Apart from that the whole thing seems to work pretty well.

csoler:
Still stuck with a bug: sorting does not work. Seems to be a GUI problem. When clicking on the header's columns, the Qt sorting method is not called.

felisucoibi:
testing but seems forums are not loading.

felisucoibi:
Something is wrong, i know this is not a good way to send you a bug report, but just tell you that rs is crashing, i will try to find where or why but It crashes when entering in forum new posts.

csoler:
I removed the [WIP] statement because it seems final to me. Please test.

felisucoibi:
when you choose a forum and no post is selected, before you click in next message and you went to the first last message posted or kind of, now it does nothing, you have to choose a thread first... this feature was usefull to reach fast to new posts. @Cyril

felisucoibi:
and also the icons are good but maybe they need to be a little more bigger, in a 2K screen are very small, i can't imagine in a 3k monitor.@Cyril

felisucoibi:
and the most important forums still don't load, forum posts... i can see the thread but when i click in a new post it does not load. when i click in a forum post this log appears...
(WW) Checking group signature with old api version 2 : tag 0 result: 1
Clicked on message ID bb6af3843588be5f0480d390c7d51986b0071d9f, index=(0,0,0x555db8845eb0)
  early return because mUpdating=true

I think the forum is LOADING and never ends, and that is the reason why.... is says loading on top of forum.

felisucoibi:
i think more or less is ready for open betateting in master, so we can have a beta widows release and my windows users can test it.

PR #1416, opened 2018-11-21
remove unnecessary warning

pedro-nonfree:
according to G10h4ck the warning to stay under 40 trusted nodes (friends) is no longer needed because this was useful when multiple system calls reached gpg tools; but not now that a c++ pgp library is used

@G10h4ck check if the information I'm saying is OK for the commit


G10h4ck:
the base branch is wrong

At first glance it seems you are only removing the text are you sure this won't break compilation?

pedro-nonfree:
> the base branch is wrong

you are right, is master branch OK?

yes, I see that this is the controller code to remove

https://github.com/RetroShare/RetroShare/blob/master/retroshare-gui/src/gui/connect/ConnectFriendWizard.cpp#L148-L149

(removing too variable `friendCount`)

and this is the UI code to remove

https://github.com/RetroShare/RetroShare/blob/master/retroshare-gui/src/gui/connect/ConnectFriendWizard.ui#L1035-L1115

but I would like to test, so wait:

- squash the commit
- test in my compilation thing (and then I compile retroshare for first time)

G10h4ck:
In that case please prefix your PR witle with WIP (Work In Progress)

pedro-nonfree:
I cannot do that by myself (permissions?)

pedro-nonfree:
OK, I thought you mean to put a label with WIP not in the title (read too fast)

I compiled and works. I think I did it well :]

---

I fixed the branch in my branch as

git merge origin/master

(do the diffs)

git push --force

pedro-nonfree:
@G10h4ck I think this is ready

Issue #1418, opened 2018-11-23
Everybody agrees to substitute Friend concept to Trusted node concept?


pedro-nonfree:
I don't know if you all agree that friend concept is not correct for the retroshare and GXS ecosystem.

In PGP/GPG it is clear that it is a matter of trust (and this is probably the same for retroshare)

I would like to see a +1 before starting a massive substitution in code. The idea for this issue is to close the issue when the concept is fully migrated. If we do everything in one commit it is easier to revert and inspect. The idea is also to automate the substitution so it is easy for everybody to apply to current developing branches, etc.

The proposal is to substitute:

text:

Friend -> Trusted Node

code:

friend -> trustedNode

csoler:
In the UI we tried to substitute "Friend" for "trusted node", since that is the correct idea behind connections, but there's still a few places where the old "Friend" term remains.

Generally speaking, massive substitutions in the code are not accepted.

So in summary, it's ok for you to make a pull request to change "Friend" for "Trusted node" in the UI.

pedro-nonfree:
@csoler so you probably prefer substitute little by little (small commits) parts of user interface, right?

pedro-nonfree:
I commented with @G10h4ck and I will go by parts (yay! I started!)

felisucoibi:
I prefer friend, is more newbies concept.

felisucoibi:
nobody knows what is a trusted node.

pedro-nonfree:
From wikipedia, [friendship](https://en.wikipedia.org/wiki/Friendship):

> Such characteristics include affection; kindness, love, virtue, sympathy, empathy, honesty, altruism, loyalty, mutual understanding and compassion, enjoyment of each other's company, **trust**, and the ability to be oneself, express one's feelings to others, and make mistakes without fear of judgment from the friend.

and this relationships evolves with age (1.1 Childhood, 1.2 Adolescence, 1.3 Adulthood, 1.4 Older adults)

Hence, a friend is a set of a lot of different characteristics that depends on age. Maybe: culture,  sexual orientation (and/or the way you love), religion, politics, etc.

Trust is a thing that may be related to a friend. A friend that you don't trust you may call him your friend. Someone you trust is probably not your friend, for example the guy that is storing your money (and you see, sometimes is an infrastructure, a trademark)... then, if we put retroshare locations, that are not friends.

[...]

PR #1420, opened 2018-11-27
substitute friend to trusted node in statusbar

pedro-nonfree:
progress for #1418


pedro-nonfree:
According to #1418 nobody agrees to go for trusted node, so this commit should be reverted

PR #1423, opened 2018-12-03
Unseen p2p app- Groupchat support

unsenep2pdev:


G10h4ck:
It is impossible to merge a PR like this, we are happy you are doing work on the code but please do smal PR based on official branches, so we can review and consider them for merge

PR #1430, opened 2018-12-13
Added new home icon

defnax:
added new home icon old was not suitable


felisucoibi:
its 4k?

defnax:
128x128 pixel

defnax:
the channel tree with new icons needs to be tested if resizing works properly ?

PR #1433, opened 2018-12-16
fix json parameter name

sehraf:


sehraf:
this is done in d19531cfccf72d2b299a042ef65e1b605381566c

PR #1440, opened 2018-12-24
added button (enabled by default) to automatically add IDs signed by …

csoler:
…friends as contacts


csoler:
This is supposed to help when we remove the node-to-node chat, by having all ids signed by friend nodes to appear in contacts automatically.

felisucoibi:
more than rpobably this setup made my rs, to add to whitelist or give positive opinion some banned users i had, because today i started to receive lots of messages from ttwo banned users.

PR #1455, opened 2019-01-06
Improved Display of Moderator selection box on a useable place

defnax:
* Improved Author & Moderator labels display as link
* Improved display of moderator selection box display on right side & added combobox all/contacts
* Added own tree item "My Circles" to see your own circles easyer then before when you are admin.
* Profile generation dialog moved node type box to advanced section.


csoler:
Thx for all this.

There's a problem in circles: when you belong to a circle for which you're the admin, it should be both in "Circles I belong to" and "My circles". So it's not possible to have both systems: the one with My circles / other circles, and the one with Circles I belong to / other circles.

Since the bullet is green when you belong to a circle, there's no need to have a category "Circles I belong to". If you wan tto keep the categories My circles / Other , it's probably fine.

But try to keep the visual coding:
- belong/invited/nothing to the circle = green/yellow/grey bullet
- bold/normal text = I have admin rights

That I don't approve:
- Profile generation dialog moved node type box to advanced section.

...because the way it was done before allows to create Tor nodes very easily with minimal input. Now if you want to create a Tor node, you need to supply a hell of information which is very confusing. So I really don't like this.

defnax:
that means Tor is more importend and the big tor description too?
last year we did lots of to make easy and user friendly now is destroyed without think why we changing the look and feel for make easy for rs newbies.

> 
> That I don't approve:
> 
> * Profile generation dialog moved node type box to advanced section.
> 
> ...because the way it was done before allows to create Tor nodes very easily with minimal input. Now if you want to create a Tor node, you need to supply a hell of information which is very confusing. So I really don't like this.

defnax:
> Thx for all this.
> 
> There's a problem in circles: when you belong to a circle for which you're the admin, it should be both in "Circles I belong to" and "My circles". So it's not possible to have both systems: the one with My circles / other circles, and the one with Circles I belong to / other circles.
> 
> Since the bullet is green when you belong to a circle, there's no need to have a category "Circles I belong to". If you wan tto keep the categories My circles / Other , it's probably fine.
> 
> But try to keep the visual coding:
> 
> * belong/invited/nothing to the circle = green/yellow/grey bullet
> * bold/normal text = I have admin rights

i need to check this again next time
yes i found that trouble
then better when im in or admin display all in My circles then no more item for  "Circles I belong to"

i wanted as admin own color too red or else, i thinked about for circle distrubution colors public or private.

csoler:
well, the bold is already used in other places for parts you're an admin of, such as identities. So it's better to use the same code here I suppose.

defnax:
> well, the bold is already used in other places for parts you're an admin of, such as identities. So it's better to use the same code here I suppose.

i wanted as red colored icon not text bold

PR #1456, opened 2019-01-09
*Improved Create comment dialog for channel/posted comment replies

defnax:
*Improved Create comment dialog for channel/posted comment replies
*Update forum icons use the new flat one green/black
*Changed the context menu icon for Pin post
*Fixed for "new thread" dialog the title icon
*Added new circle icons for later to use


felisucoibi:
+1000

csoler:
I like the new design. I think there's two (small) problems:
- when you change the "post as" box, the identity on top of the window does not change. Sometimes it is even different when the window is initialized
- it's a big confusing to see the poster's identity in two different places

defnax:
> I like the new design. I think there's two (small) problems:
> 
> * when you change the "post as" box, the identity on top of the window does not change. Sometimes it is even different when the window is initialized
> * it's a big confusing to see the poster's identity in two different places

the avatar has nothing todo with with my own identity.
i think you dont understand how? top is for reply action only.

avatar of the person ( for reply)
name of person ( for reply )
comment of the person ( for reply)

yes normaly we can disable/hide the id box for reply i forget to add string "Reply as" and i doesnt set yet window title to reply. i thinked you understand it.

with that method you know better like on twitter  when you reply
image

defnax:
again more improved:
but the cancel button can be hided we has always a x on window dialog
image

PR #1458, opened 2019-01-11
V0.6 improved gui

csoler:


csoler:
to be tested. There might be some side effects I haven't encountered myself.

PR #1463, opened 2019-01-18
Added for posted links group icons

defnax:
this needs to be tested and fixed there is a issue.

rs

- old versions get the posted groups, 
- this changes doesnt get from old versions the posted groups.
- when both rs clients using with icon feature then there is no problem.

 @csoler @G10h4ck  can you view and fix it ?
please only look on the changes for group icons
here https://github.com/RetroShare/RetroShare/pull/1463/commits/351e54ea0cf108f9eecc449a13704f99fec5873d


defnax:
@hunbernd can you test it this changes and maybe fix it?
here https://github.com/RetroShare/RetroShare/commit/351e54ea0cf108f9eecc449a13704f99fec5873d

PR #1464, opened 2019-01-18
removed HeaderFrame where it's not needed

zapek:
Also changed the font of top tooltips/infolabels (the yellow fixed ones) to be the same everywhere.


defnax:
i dont agree it!

i only allow top bars from mainwinndow!!!!

why you waste your time with remove the usefull headerframes!!!.
again nobody help on the real places! other want add ugly names the second one want remove all headerframes.
you want make retroshare as expert/geek application? geek want remove the good looking stuff.
then you can remove all icons from the retroshare too. then no frames no icons.
remove all my work!

zapek:
I already asked on the forum to give one valid reason why those headerframes are useful and so far you didn't answer.

Here are my reasons:
- they take up unnecessary space (it's an important rule for UI design)
- they repeat an information that is already provided by the window's title bar

Also I made a list on the forum and asked for feedback, Cyril provided some and I only did what he agreed upon.

Now can you provide a valid reason to keep those headerframes other than what I'm guessing is that you're the author of them?

If you're an UI expert, you should know that they evolve over time. Nobody puts icons on OK/Cancel buttons anymore for example. And while those headerframes might have looked good in the past, they really look out of place now (try other Qt software like Quassel or Mumble...).

Another thing is that I already did the work. By closing my pull request to "teach me" on what to work, you're not giving me back the hours I spent and are, in fact, discouraging me from providing further contributions.

csoler:
I fully agree with you zapek. Furthermore,  closing this pull request wasn't a respectful move at all.

defnax:
i dont agree!

PR #1477, opened 2019-02-14
Fix clang warnings

PhenomRetroShare:


retroshare.pri
```diff
@@ -434,6 +434,8 @@ no_rs_deprecatedwarning {
 
 no_rs_cppwarning {
     QMAKE_CXXFLAGS += -Wno-cpp
+    QMAKE_CXXFLAGS += -Wno-inconsistent-missing-override
```
G10h4ck, line 437: I think this warning is useful and we shouldn't suppress it, I have started adding override keyword when appropriate on the code I touch

libretroshare/src/services/p3gxschannels.h
```diff
@@ -132,30 +132,29 @@ virtual bool getChannelDownloadDirectory(const RsGxsGroupId &groupId, std::strin
...
+	virtual bool getCommentData(uint32_t token, std::vector<RsGxsComment> &msgs) override
 	{ return mCommentService->getGxsCommentData(token, msgs); }
 
 	virtual bool getRelatedComments( uint32_t token,
-	                                 std::vector<RsGxsComment> &msgs )
+	                                 std::vector<RsGxsComment> &msgs ) override
 	{ return mCommentService->getGxsRelatedComments(token, msgs); }
 
-virtual bool createComment(uint32_t &token, RsGxsComment &msg)
+	virtual bool createNewComment(uint32_t &token, RsGxsComment &msg) override
```
G10h4ck, line 142: Is this change supposed to be included in this PR ?

G10h4ck (changes requested):

retroshare.pri
PhenomRetroShare: This is removed only with no_rs_cppwarning flag.
So it will be removed when less warnings exists. For now it pollute compiler log.

libretroshare/src/services/p3gxschannels.h
PhenomRetroShare: Yes for sure.
It's on  Fix Warnings for 'RsGxsChannels::createComment' and 'RsGxsChannels::createVote' hides overloaded virtual function
commit.
Do you look at commits comments?

G10h4ck (approved):

PR #1485, opened 2019-02-23
Fix clang warnings

PhenomRetroShare:
All said in commits comment.


retroshare-gui/src/gui/FileTransfer/TransfersDialog.cpp
```diff
@@ -1108,6 +1109,7 @@ void TransfersDialog::activatePage(TransfersDialog::Page page)
 													break ;
 		case TransfersDialog::RemoteSharedFilesTab: ui.tabWidget->setCurrentWidget(remoteSharedFiles) ;
 													break ;
+		case TransfersDialog::DownloadTab: ui.tabWidget->setCurrentWidget(ui.tab) ;
```
G10h4ck, line 1112: `break;` missing, AFAIR if you don't put it following cases are axecuted up to first break, in this case it would jump to `case TransfersDialog::SearchTab:`

G10h4ck (changes requested):
Seems ok exctept for a missing break; in a switch

retroshare-gui/src/gui/FileTransfer/TransfersDialog.cpp
PhenomRetroShare: Are you sure of that?
https://onlinegdb.com/rJVzA7gDV

PhenomRetroShare: Done ;)

G10h4ck (approved):

G10h4ck (approved):

PR #1487, opened 2019-02-25
Added more actions to People V2 Thumbed View

defnax2k:
Added action for send invite
Added action for add to contacts
Added action for person details view
Added menu for circles
Added to display votes on Identity widget

before its lost, my changes from last year 

image


felisucoibi:
+1

Issue #469, opened 2016-08-12
Add random Avatar for Friends (when no own Avatar is set)


defnax:
When friend has not set a own avatar,
without avatars looks boring,
Set a random avatar for friends automaticly, like how on GXS Users, but not use the same or random avatar with different colors.

other idea:
https://github.com/arusanov/avatar-generator

Demo 
http://eightbitavatar.herokuapp.com/?id=userid&s=male&size=400 
Where: id - your user id s - sex (male|female) size - avatar size

PR #1489, opened 2019-03-01
Ui avatar cleanup

zapek:


defnax2k:
that idea for the avatar dialog wasto add later some default avatars for selection and use for rs.
not had time to develope some default avatars for selection, and else upload seperate one
You can help develope a avatar selection dialog?

image

https://github.com/RetroShare/RetroShare/issues/469

zapek:
@defnax2k that can always be done later. Also if direct chat is replaced by gxs chat, this will have to be done for identities anyway.

PR #1491, opened 2019-03-10
Fix jsonapi compilation

hunbernd:
- Fixed issues on Windows
- Fixed issues with multi core compilation
- Fixed Libretroshare compilation starts before headers from restbed are ready, and compilation fails at the first time.

### DIFF
--- a/libretroshare/src/libretroshare.pro
+++ b/libretroshare/src/libretroshare.pro
@@ -862,36 +862,24 @@ rs_jsonapi {
 
     no_rs_cross_compiling {
         DUMMYRESTBEDINPUT = FORCE
+        CMAKE_GENERATOR_OVERRIDE=""
+        win32-g++:CMAKE_GENERATOR_OVERRIDE="-G \"MSYS Makefiles\""
         genrestbedlib.name = Generating libresbed.
         genrestbedlib.input = DUMMYRESTBEDINPUT
         genrestbedlib.output = $$clean_path($${RESTBED_BUILD_PATH}/librestbed.a)
         genrestbedlib.CONFIG += target_predeps combine
         genrestbedlib.variable_out = PRE_TARGETDEPS
-        win32-g++ {
-            genrestbedlib.commands = \
-                cd $${RS_SRC_PATH} && \
-                git submodule update --init --recommend-shallow supportlibs/restbed && \
-                cd $${RESTBED_SRC_PATH} && \
-                git submodule update --init --recommend-shallow dependency/asio && \
-                git submodule update --init --recommend-shallow dependency/catch && \
-                git submodule update --init --recommend-shallow dependency/kashmir && \
-                mkdir -p $${RESTBED_BUILD_PATH}; cd $${RESTBED_BUILD_PATH} && \
-                cmake -DCMAKE_CXX_COMPILER=$$QMAKE_CXX -G \"MSYS Makefiles\" -DBUILD_SSL=OFF \
-                    -DCMAKE_INSTALL_PREFIX=. -B. -H$$shell_path($${RESTBED_SRC_PATH}) && \
-                make
-        } else {
-            genrestbedlib.commands = \
-                cd $${RS_SRC_PATH};\
-                git submodule update --init --recommend-shallow supportlibs/restbed;\
-                cd $${RESTBED_SRC_PATH};\
-                git submodule update --init --recommend-shallow dependency/asio;\
-                git submodule update --init --recommend-shallow dependency/catch;\
-                git submodule update --init --recommend-shallow dependency/kashmir;\
-                mkdir -p $${RESTBED_BUILD_PATH}; cd $${RESTBED_BUILD_PATH};\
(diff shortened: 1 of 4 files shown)


sehraf:
> Libretroshare compilation starts before headers from restbed are ready, and compilation fails at the first time.

This can actually happen on linux, too.

csoler:
@sehraf : this PR works for you on linux?

sehraf:
I can test it on Thursday.

csoler:
ok. Will wait for your feedback. thx.

sehraf:
clean built it two times without problems, :+1:

hunbernd:
Please check if you can use the API.

On Windows it created some empty files. The build was successful, but the API returned only empty messages.
I fixed the problem, hopefully I didn't break it on Linux meanwhile.

sehraf:
API is working as expected (tested with a bunch of my scripts)

libretroshare/src/libretroshare.pro
```diff
@@ -861,45 +861,69 @@ rs_jsonapi {
...
-        QMAKE_EXTRA_TARGETS += restbed
-        libretroshare.depends += restbed
-        PRE_TARGETDEPS *= $${restbed.target}
+        DUMMYRESTBEDINPUT = FORCE
+        genrestbedlib.name = Generating libresbed.
+        genrestbedlib.input = DUMMYRESTBEDINPUT
+        genrestbedlib.output = $$clean_path($${RESTBED_BUILD_PATH}/librestbed.a)
+        genrestbedlib.CONFIG += target_predeps combine
+        genrestbedlib.variable_out = PRE_TARGETDEPS
+        win32-g++ {
```
G10h4ck: Do I miss something or the only difference here between the ` win32-g++` and `else` branch is the usage of `&&` instead of `;` in that case it seems reasonable to me (that is the original writer of that code) to use `&&` for all systems removing the conditional

libretroshare/src/libretroshare.pro
```diff
@@ -861,45 +861,69 @@ rs_jsonapi {
...
+    genjsonapi.output = $${WRAPPERS_INCL_FILE} $${WRAPPERS_REG_FILE}
+    genjsonapi.clean = $${WRAPPERS_INCL_FILE} $${WRAPPERS_REG_FILE}
+    genjsonapi.CONFIG += target_predeps combine no_link
+    genjsonapi.variable_out = HEADERS
+    genjsonapi.commands = \
         mkdir -p $${JSONAPI_GENERATOR_OUT} && \
         cp $${DOXIGEN_CONFIG_SRC} $${DOXIGEN_CONFIG_OUT} && \
-        echo OUTPUT_DIRECTORY=$$shell_path($${JSONAPI_GENERATOR_OUT}) >> $${DOXIGEN_CONFIG_OUT} && \
-        echo INPUT=$$shell_path($${DOXIGEN_INPUT_DIRECTORY}) >> $${DOXIGEN_CONFIG_OUT} && \
+        echo OUTPUT_DIRECTORY=$${JSONAPI_GENERATOR_OUT} >> $${DOXIGEN_CONFIG_OUT} && \
```
G10h4ck, line 910: Are you sure this works better without the `$$shell_path(...)` ?

libretroshare/src/libretroshare.pro
```diff
@@ -861,45 +861,69 @@ rs_jsonapi {
...
+        echo INPUT=$${DOXIGEN_INPUT_DIRECTORY} >> $${DOXIGEN_CONFIG_OUT} && \
```
G10h4ck, line 911: Are you sure this works better without the $$shell_path(...) ?

G10h4ck (changes requested):
Great, this contribution fix an obnoxious problem, that I couldn't manage to solve with my qmake knowledge! Please take a look at my comments inline ;)

libretroshare/src/libretroshare.pro
hunbernd: It makes sense to use && on Linux too, because ; ignores errors in previous steps.

But the significant change is this cmake option, else cmake will try to use VisualC:
-G \"MSYS Makefiles\"

libretroshare/src/libretroshare.pro
hunbernd: With shell_path qmake will print Linux like paths into the doxygen config, and doxygen will throw an error.

libretroshare/src/libretroshare.pro
hunbernd: See above.

libretroshare/src/libretroshare.pro
G10h4ck: So only that part should be generated in an os specific branch, you can define a variable like 
```
CMAKE_GENERATOR_OVERRIDE=""
win32-g++:CMAKE_GENERATOR_OVERRIDE="-G \"MSYS Makefiles\""
```

and the use that in the `cmake` command line

`cmake -DCMAKE_CXX_COMPILER=$$QMAKE_CXX  $${CMAKE_GENERATOR_OVERRIDE} -DBUILD_SSL=OFF \ `

G10h4ck (changes requested):

PR #1493, opened 2019-03-12
V0.6 messaging

csoler:
Implemented an item model for messages. Should allow a much faster loading of messages when there is a lot of them. To be tested.


csoler:
This PR is ready for testing. Please report any bugs.

PR #1494, opened 2019-03-12
set default FT rate to some non zero value. Should start transfers faster. Needs testing!

csoler:


felisucoibi:
Hi, first of all i said in forums and in chatrooms that FT is completely destroyed in rs, don't know exactly since when.

I tested this pr, and here goes what i'm experiencing now.

I start RS and every x seconds, is like it tries to download from lot of sources, but looks like is a fake due to the change of the starting number, so after a few secs this download speeds goes to 0 again.
BUT, seems something changed, so some downloads started to have more "energy" than before the PR, what makes me to thinkg that this PR is just a proof of something is wrong in RS FT, but i have to admit that i will keep this PR, because is doing something, I mean is better than whas before, in fact much better.

But still, is kind of cracy behavour, file downloading starts, gets 400K/s then goes back to 25k/s, then lot of files transdfer at 0.x k/s, some gets faster, again on, and off...

I mean, this pr, seems making file transfer beter thant the last 2 months that stucked my rs dowloading list to HELL.

But FT is broken somehow that needs lot of love.

For example, if there are direct friends connected, the connectin MUST BE STABLE, and using as much BW as possible, since sec 1. It's direct freiwend, is online, and we can share directly. 
What is hapening now, is kind of cracy, up and down of download speed, cracy random file start from the same friend but different files, every 3 secs. File transfer connect and disconnect every sec...

Some TEST POINTS needed to find what is making rs file transfer HELL, but this pr made a big step, the last two months i suffered the same, but with transfer speeds of 1K/s, since this pr some files started to transfer faster... in a very crazy way.

felisucoibi:
After a few days this pr is a must, and should be a 0.6.5.2 with thi pr, because more or less downloads works again. With 0.6.5 is not.

csoler:
@felisucoibi : apparently something else causes some disturbance in the file transfer of your node. This PR should only make transfers start faster. When transfers are stuck you should have a look at upload queues and determine what's blocking the traffic.

felisucoibi:
FT needs lot of love, everybody that use them knows it in RS, also filelists, meanwhile we have just a toy to play with and we use other tools for real work. Nobody really uses RS for file transfers because they are slower than emule. Nobody really uses RS for chats, or for filelist browsing.... the only one thing that works are chatrooms... more or less.

PR #1501, opened 2019-03-21
created a cache for default icons to avoid allocating them everytime …

csoler:
…we need them. Should save a lot of memory

To be tested!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ addons:
 
 before_script:
   - if [ $TRAVIS_OS_NAME == linux ]; then qmake QMAKE_CC=$CC QMAKE_CXX=$CXX; fi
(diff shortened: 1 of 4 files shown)


felisucoibi:
+1 testing.... if i can fix the compilation problem

csoler:
There's something I cannot figure out: the cache holds an extra reference to each image, and because the images data uses shared pointers, there's no extra cost in passing this to Qt. That would mean it's never necessary to clean the cache.
On the other hand, because sometimes Qt doesn't need some icons anymore, we can safely remove them from the cache which deleted the data since it's the last reference.
So I dont know exactly which one to choose between cleaning the cache sometimes or never. For now in this PR it cleans it ever 5mins secs and removes icons that are not requested for  more than 10 mins. Given the above reflexions, the time should be extended a bit further. The risk in deleting the cache items too soon is that the items are re-generated and therefore do not share the pointers with the previous references that are hold by other Qt widgets, which duplicated the meomory used (as before without the cache)

PR #1509, opened 2019-04-03
display session traffic

RetroPooh:


retroshare-gui/src/gui/MainWindow.cpp
```diff
@@ -726,9 +726,12 @@ void MainWindow::updateStatus()
     float downKb = 0;
     float upKb = 0;
     rsConfig->GetCurrentDataRates(downKb, upKb);
+	uint64_t down = 0;
+	uint64_t up = 0;
+	rsConfig->GetTrafficSum(down, up);
 
     if (ratesstatus)
-        ratesstatus->getRatesStatus(downKb, upKb);
+        ratesstatus->getRatesStatus(downKb, down, upKb, up);
```
csoler, line 734: plz rename down->down_total_Kb and up->up_total_Kb. Otherwise because we have downKb and down it sounds like the two variables hold the same quantity with different units. The new down and up are in Kb as well right? If not, just chop of the _Kb part.

retroshare-gui/src/gui/statusbar/ratesstatus.cpp
```diff
@@ -48,13 +49,13 @@ RatesStatus::RatesStatus(QWidget *parent)
     setLayout(hbox);
 }
 
-void RatesStatus::getRatesStatus(float downKb, float upKb)
+void RatesStatus::getRatesStatus(float downKb, uint64_t down, float upKb, uint64_t upl)
```
csoler, line 52: same here

PR #1510, opened 2019-04-03
fixed wrong file hash daylight saving time warnings on windows

zapek:
On Windows, there's a check for +1/-1 hours because NTFS can change the modification time of a file twice per year during the DST change. The warning is wrong though because Retroshare emits it everytime it's launched for every file shared (DST or not).

### DIFF
--- a/libretroshare/src/file_sharing/hash_cache.cc
+++ b/libretroshare/src/file_sharing/hash_cache.cc
@@ -133,12 +133,7 @@ void HashStorage::data_tick()
 
                 if(!mChanged)	// otherwise it might prevent from saving the hash cache
                 {
-                    std::cerr << "Stopping hashing thread." << std::endl;
-                    shutdown();
-                    mRunning = false ;
-                    mTotalSizeToHash = 0;
-                    mTotalFilesToHash = 0;
-                    std::cerr << "done." << std::endl;
+                    stopHashThread();
                 }
 
                 RsServer::notify()->notifyHashingInfo(NOTIFY_HASHTYPE_FINISH, "") ;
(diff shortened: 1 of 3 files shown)


csoler:
this code was originally meant to correct a bug in windows system that I'm sure was meant to be a feature: when you accessed the file time, the system was automatically changing it to be in your current time zone, instead of returning the actual time on the disk. So when the timezone changes (such as daylight saving time shifts twice a year) all files appeared in windows to have a different time than what was registered and got re-hashed.

You;re telling that this feature has disappeared from windows? It sounds like a big change. Are you sure?

csoler:
If the warning shows up, it tells you exactly that: RS will not re-hash whereas it would have rehashed without this special code. So the fact that you see this warning is a good sign that the code is useful. What you can do however is put a static bool in the code to check that the warning is printed only once.

zapek:
What I mean is that the check is wrong because this warning is printed for every file I share every time I launch Retroshare and not twice per year, every time.

```
it->second.time_stamp = now;
if (it->second.time_stamp != (uint64_t)mod_time)
{
    // this will always be true
}
```

From my understanding, the comparison before the warning does the right thing:
`( (uint64_t)mod_time == it->second.modf_stamp || (uint64_t)mod_time+3600 == it->second.modf_stamp ||(uint64_t)mod_time == it->second.modf_stamp+3600)`

This seems correct as it compares the time 3 times (normal, -1 hour, +1 hour).

csoler:
This code is quite subtle. Took me a while to get the problem
What happens is that:
 1 - RS sees that there's a shift and prints a warning
 2 - if there is a shift, the time is not set to now (which would cause rehash next time because it's not the actual time of the file on the disk). In this case it is set to the actual time of modification of the file

The correct line for the if is 
      if(it->second.modf_stamp != (uint64_t)mod_time)
There's no reason indeed to compare the time of last request for the hash to the file modification time

Same below: it should be
 it->second.modf_stamp = (uint64_t)mod_time;

Otherwise what happens is that old files produce a mod_time that is old and therefor it->second.time_stamp is set to that old value, which makes it be removed from the hash cache and therefore the file is rehashed at next reboot (see hash_cache.cc:329)

zapek:
I added a way to start the hashing thread when needed. I tested and now I don't get the file modification warnings anymore.

PR #1512, opened 2019-04-04
Improve on top of the work made by Cyril on channel API

G10h4ck:
Fix compilation, retrocompatibility and enums
    
    Workaround miss-behaviour on old Android phones
    Cleanup indentation a bit
    Consistent param naming
    Introduce default parameter values also for output paramethers

### DIFF
--- a/libretroshare/src/retroshare/rsgxsifacehelper.h
+++ b/libretroshare/src/retroshare/rsgxsifacehelper.h
@@ -314,7 +314,7 @@ LLwaitTokenBeginLabel:
 #if defined(__ANDROID__) && (__ANDROID_API__ < 24)
 		/* Work around for very slow/old android devices, we don't expect this
 		 * to be necessary on newer devices. If it take unreasonably long
-		 * something worser is already happening elsewere amd we return anyway.
+		 * something worser is already happening elsewere and we return anyway.
 		 */
 		if( st > RsTokenService::FAILED && st < RsTokenService::COMPLETE
 		        && maxWorkAroundCnt-- > 0 )
--- a/libretroshare/src/retroshare/rsgxsifacehelper.h
+++ b/libretroshare/src/retroshare/rsgxsifacehelper.h
@@ -4,7 +4,7 @@
  * libretroshare: retroshare core library                                      *
  *                                                                             *
  * Copyright 2011 by Christopher Evi-Parker                                    *
- * Copyright (C) 2018  Gioacchino Mazzurco <gio@eigenlab.org>                  *
+ * Copyright (C) 2018-2019  Gioacchino Mazzurco <gio@eigenlab.org>             *
  *                                                                             *
  * This program is free software: you can redistribute it and/or modify        *
  * it under the terms of the GNU Lesser General Public License as              *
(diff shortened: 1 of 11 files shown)


libretroshare/src/retroshare/rsgxsifacehelper.h
```diff
@@ -292,25 +290,55 @@ struct RsGxsIfaceHelper
...
+		auto wkStartime = std::chrono::steady_clock::now();
+		int maxWorkAroundCnt = 10;
+LLwaitTokenBeginLabel:
+#endif
 		auto timeout = std::chrono::steady_clock::now() + maxWait;
 		auto st = requestStatus(token);
 		while( !(st == RsTokenService::FAILED || st >= RsTokenService::COMPLETE)
 		       && std::chrono::steady_clock::now() < timeout )
 		{
-			std::this_thread::sleep_for(std::chrono::milliseconds(2));
```
csoler, line 305: Sleeping for only 2 ms between two checks is probably far too small. You'll be creating lots of CPU loops. I would advise 100 or 200 ms here.

libretroshare/src/util/rsmemory.h
```diff
@@ -23,8 +23,11 @@
 
 #include <stdlib.h>
 #include <iostream>
+#include <memory>
 #include <util/stacktrace.h>
 
+#define RS_DEFAULT_STORAGE_PARAM(Type) *std::unique_ptr<Type>(new Type)
```
csoler, line 63: Could you add a quick comment about this?

csoler: Congrats for the macro btw ;-)

csoler:
I added a few comments inline. Everything seems clean.
I still think we should rename the old methods *_deprecated() and use the normal names for the new ones. This will force current developers to replace the names if they want to keep using the old format, but it's very easy to do, and it will guarranty that in 2-3 years when we remove the *_deprecated() methods and we hopefully have lot more developers using them, it wont annoy anyone and ensure that the name is the most appropriate.

G10h4ck:
> I added a few comments inline. Everything seems clean.
> I still think we should rename the old methods *_deprecated() and use the normal names for the new ones. This will force current developers to replace the names if they want to keep using the old format, but it's very easy to do, and it will guarranty that in 2-3 years when we remove the *_deprecated() methods and we hopefully have lot more developers using them, it wont annoy anyone and ensure that the name is the most appropriate.

The problem with that is that for JSON API users the things that used to works, would break without much notice, and because we released a stable version with that API I would avoid breaking the API at least for a couple versions, otherwise JSON API will be not attractive at all other projects/developers :-\

csoler:
Come on! It's not compiled by default so people who use it compile RS themselves. Changing the name is not a terrible task. It's anyway better than keeping this V2 name for years.

libretroshare/src/retroshare/rsgxsifacehelper.h
G10h4ck: 100ms sound reasonable for operations like channel creation that take lot of time but it is far too much for simple reads, I have been measuring this value even on old android phones and for reads operations 2ms is more then enough (probably the system does some caching), also this PR doesn't change the default value it just make it configurable, I agree that long operations like channel creation might be optimized by passing a longer value as parameters ;)

G10h4ck:
> Come on! It's not compiled by default so people who use it compile RS themselves. Changing the name is not a terrible task. It's anyway better than keeping this V2 name for years.

It is enabled at compile time in most linux distributions packages, and it is mandatory for retroshare-service ;)

PR #1513, opened 2019-04-06
add AGPL license text

Piraty:
Hi,

many source files mention the `AGPL-3.0-or-later` license in their header, but there is no license file for AGPL in the project.

Also see: https://spdx.org/licenses/AGPL-3.0-or-later.html

[CI-SKIP]

### DIFF
--- a/LICENSE.AGPL.txt
+++ b/LICENSE.AGPL.txt
@@ -1,0 +1,661 @@
+                    GNU AFFERO GENERAL PUBLIC LICENSE
+                       Version 3, 19 November 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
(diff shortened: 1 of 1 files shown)


csoler:
We have recently updated the licence for the whole project. The licence file that is up to date is in build_scripts/Debian/debian/copyright. We will move it to the root of the project.

G10h4ck:
AGPL should  probably be included anyway because that is the license that apply when jsonapi is enabled?

P.S. we probably need to update build_scripts/Debian/debian/copyright specifying jsonapi-generator and jsonapi folder inside libretroshare are AGPLv3+

Piraty:
thanks for merging.
btw more distros than Debian might be [packaging](https://github.com/void-linux/void-packages/pull/10257) your software, so whatever you do in debian/ might not be obvious to everybody else

PR #1519, opened 2019-04-11
Fixed channel API

csoler:


G10h4ck:
please avoid lines longer the 80 columns as per https://retroshare.readthedocs.io/en/latest/developer/coding/ it make diffing side by side quite cumbersome, especially in non-huge screens like my laptop, I'll fix those huge lines and merge the PR

G10h4ck:
Also std::string conversion is not needed in most the cases you used it

csoler:
well, the problem with linefeeds in the middle of comments like this is that it makes the code really hard to read. So if your screen wraps the long lines when you display them, it's mostly your problem. Also, limiting the code to 80 columns is quite an outdated requirement in 2019. I certainly did not contribute to that in the coding requirement you're citing ;-)

G10h4ck:
Also longer then 80 lines make it difficult to read I have to scroll horizontally all the time :-|

G10h4ck:
included into https://github.com/RetroShare/RetroShare/pull/1523

PR #1520, opened 2019-04-12
Broadcast Domain friends IP:Port Discovery

G10h4ck:
Discover peers on the same broadcast domain/LAN
Implements own node announcement
Implements others announcement listening
Implements API to get notifiend when other nodes get discovered
Implements API to get discovered nodes
Add discovered peer locator to rsPeers if it is already friend

More things such as discovered peers friends suggestion will come in another pull request

### DIFF
--- a/libretroshare/src/rsserver/p3face-server.cc
+++ b/libretroshare/src/rsserver/p3face-server.cc
@@ -39,6 +39,9 @@ int rsserverzone = 101;
 
 #include "util/rsdebug.h"
 
+#include "retroshare/rsevents.h"
+#include "services/rseventsservice.h"
+
 
 /****
 #define DEBUG_TICK 1
@@ -81,6 +84,12 @@ RsServer::RsServer() :
     coreMutex("RsServer"), mShutdownCallback([](int){}),
     coreReady(false)
 {
+	{
+		RsEventsService* tmpRsEvtPtr = new RsEventsService();
+		rsEvents.reset(tmpRsEvtPtr);
+		startServiceThread(tmpRsEvtPtr, "RsEventsService");
+	}
+
 	// This is needed asap.
 	//
 	mNotify = new p3Notify() ;
--- a/libretroshare/src/services/broadcastdiscoveryservice.h
+++ b/libretroshare/src/services/broadcastdiscoveryservice.h
@@ -43,14 +43,6 @@ public:
 	/// @see RsBroadcastDiscovery
 	std::vector<RsBroadcastDiscoveryResult> getDiscoveredPeers() override;
 
-	/// @see RsBroadcastDiscovery
-	bool registerPeersDiscoveredEventHandler(
-	        const std::function<void (const RsBroadcastDiscoveryResult&)>&
-	        multiCallback,
-	        rstime_t maxWait = 300,
-	        std::string& errorMessage = RS_DEFAULT_STORAGE_PARAM(std::string)
-	        ) override;
-
 	/// @see RsTickingThread
 	void data_tick() override;
 
@@ -67,16 +59,6 @@ protected:
 	RsMutex mDiscoveredDataMutex;
 
 	RsPeers& mRsPeers; // TODO: std::shared_ptr<RsPeers> mRsPeers;
-
-	typedef std::pair<
-	std::function<void (const RsBroadcastDiscoveryResult&)>,
-	    std::chrono::system_clock::time_point > timedDiscHandlers_t;
-	/** Store peer discovered event handlers with timeout */
-	std::forward_list<timedDiscHandlers_t> mPeersDiscoveredEventHandlersList;
-	RsMutex mPeersDiscoveredEventHandlersListMutex;
-
-	/// Cleanup mPeersDiscoveredEventHandlersList
-	void cleanTimedOutEventHandlers();
 
 	RsBroadcastDiscoveryResult createResult(
 	        const UDC::IpPort& ipp, const std::string& uData );
--- a/libretroshare/src/services/broadcastdiscoveryservice.h
+++ b/libretroshare/src/services/broadcastdiscoveryservice.h
@@ -1,0 +1,83 @@
+/*******************************************************************************
+ * RetroShare Broadcast Domain Discovery                                       *
+ *                                                                             *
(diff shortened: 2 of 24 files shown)


sehraf:
Does this replace zeroconf? If so please consider removing zeroconf

felisucoibi:
very important to improve lan and wan discovery without internet.

G10h4ck:
> Does this replace zeroconf? If so please consider removing zeroconf

Yeah!
Do zeroconf ever worked with retroshare?

G10h4ck:
> > Does this replace zeroconf? If so please consider removing zeroconf
> 
> Yeah!
> Do zeroconf ever worked with retroshare?

It seems that at least 0.6 releases never had zeroconf enabled a compile time, zeroconf seems to offer some NAT port forward functionality configuration like UPnP, but i doub't that would work in most cases... So for me the zeroconf code can be removed, but I think that comes on a pull request on it's own.

csoler:
Your PR has a license problem: libretroshare/* should be LGPLv3+ as we agreed, not AGPLv3+.
See build_scripts/Debian/debian/copyright

G10h4ck:
Ok for me to re-license this PR to LGPL, but JSON API specific code (libretroshare/src/jsonapi) is under libretroshare too, but we agreed that that should be to kept as AGPL, how can we handle that (libretroshare/src/jsonapi) ?

csoler:
you need to add an entry in the Debian copyright file (sorry but I spent a lot of time fixing all this so I try to keep it up to date) before libretroshare/src/. There's already a few of them, just copy+update one of them

csoler:
up to now there was no entry for json api in the copyright file because I removed it from the debian package since it wasn't enabled at compilation time.

pedro-nonfree:
```diff
[submodule "supportlibs/udp-discovery-cpp"]
	path = supportlibs/udp-discovery-cpp
-	url = https://github.com/truvorskameikin/udp-discovery-cpp.git
+	url = https://github.com/RetroShare/udp-discovery-cpp.git
	branch = develop
```

I would feel more safe if retroshare org forks this project and the git submodule points to the forked project. That way the responsability is directly related to retroshare devs that can apply changes on their own and have the chance to apply or not changes from upstream (truvorskameikin/udp-discovery-cpp.git), etc.

EDIT: add that this repo is also in `build_scripts/Android/prepare-toolchain-clang.sh`

G10h4ck:
> ```diff
> [submodule "supportlibs/udp-discovery-cpp"]
> 	path = supportlibs/udp-discovery-cpp
> -	url = https://github.com/truvorskameikin/udp-discovery-cpp.git
> +	url = https://github.com/RetroShare/udp-discovery-cpp.git
> 	branch = develop
> ```
> 
> I would feel more safe if retroshare org forks this project and the git submodule points to the forked project. That way the responsability is directly related to retroshare devs that can apply changes on their own and have the chance to apply or not changes from upstream (truvorskameikin/udp-discovery-cpp.git), etc.
> 
> EDIT: add that this repo is also in `build_scripts/Android/prepare-toolchain-clang.sh`

I was thinking to fork the repo to ensure we have a copy of it, about the submodule it is not a problem to leave it pointing to the original version because submodules are bond to a specific commit, so it checkout the version I reviewed unless I change the commit which the submodule points to in this repository

The android toolchain script is bond to the develop branch instead that can change, but it is ok right now as this is used for development purposes ATM and I what to keep track of upstream changes there, when a release is prepared I'll take care to use a reviewed version of the library (for testing you can already do that defining the version variable to a commit hash in your shell before running the script)

libretroshare/src/rsserver/p3face-server.cc
```diff
@@ -81,6 +84,12 @@ RsServer::RsServer() :
     coreMutex("RsServer"), mShutdownCallback([](int){}),
     coreReady(false)
 {
+	{
+		RsEventsService* tmpRsEvtPtr = new RsEventsService();
```
csoler, line 88: Seems to me that this should be created in rsinit.cc, in the same place than rsnotify is created.

libretroshare/src/services/broadcastdiscoveryservice.h
```diff
@@ -0,0 +1,65 @@
...
+	virtual ~BroadcastDiscoveryService() override;
+
+	/// @see RsBroadcastDiscovery
+	std::vector<RsBroadcastDiscoveryResult> getDiscoveredPeers() override;
+
+	/// @see RsTickingThread
+	void data_tick() override;
+
+protected:
+	constexpr static uint16_t port = 36405;
```
csoler, line 50: what happens is RS connexion port happens to be the same port number??

libretroshare/src/services/broadcastdiscoveryservice.h
```diff
@@ -0,0 +1,65 @@
...
+
+	void updatePublishedData();
+
+	UDC::EndpointParameters mUdcParameters;
+	UDC::Endpoint mUdcEndpoint;
+
+	std::map<UDC::IpPort, std::string> mDiscoveredData;
+	RsMutex mDiscoveredDataMutex;
+
+	RsPeers& mRsPeers; // TODO: std::shared_ptr<RsPeers> mRsPeers;
```
csoler, line 61: that's pretty horrible. Somebody would be able to erase RsPeers from here. I suggest that you use a pointer instead: RsPeers *mRsPeers. Another alternative is to use the global variable rsPeers.

libretroshare/src/services/broadcastdiscoveryservice.h
G10h4ck: the listening socket bind on 255.255.255.255 so it should not conflict with retroshare anyway

libretroshare/src/services/broadcastdiscoveryservice.h
G10h4ck: This is not a reference to the pointer &* it is a reference to the object, so it is equivalent to the plain pointer, except it can't be null and can't be deleted from there either

libretroshare/src/rsserver/p3face-server.cc
G10h4ck: I'll take a look at that, thanks!

PR #1534, opened 2019-04-22
RFC: Add support for short invites

G10h4ck:
With this PR it is possible to add a friend based on her SSL id only so a short invite looks like rs://89a4101580dd61c119e4b3cf2dbe53f7?extAddr=192.0.2.28&extPort=7992&name=Test%2004

This make it possible to fit a  retroshare invite in low size mediums like a QR code or a broadcast UDP packet.

I have successfully tested the functionalities introduced by this pull request through JSON API.

### DIFF
--- a/libretroshare/src/chat/distantchat.h
+++ b/libretroshare/src/chat/distantchat.h
@@ -33,9 +33,6 @@ static const uint32_t DISTANT_CHAT_AES_KEY_SIZE = 16 ;
 class DistantChatService: public RsGxsTunnelService::RsGxsTunnelClientService
 {
 public:
-    // So, public interface only uses DistandChatPeerId, but internally, this is converted into a RsGxsTunnelService::RsGxsTunnelId
-    
-   
     DistantChatService() ;
 
     virtual void triggerConfigSave()=0 ;
@@ -91,9 +88,13 @@ public:
     virtual void connectToGxsTunnelService(RsGxsTunnelService *tunnel_service) ;
     
 private:
-    virtual bool acceptDataFromPeer(const RsGxsId& gxs_id, const RsGxsTunnelService::RsGxsTunnelId& tunnel_id, bool am_I_client_side) ;
-    virtual void notifyTunnelStatus(const RsGxsTunnelService::RsGxsTunnelId& tunnel_id,uint32_t tunnel_status) ;
-    virtual void receiveData(const RsGxsTunnelService::RsGxsTunnelId& id,unsigned char *data,uint32_t data_size) ;
+	virtual bool acceptDataFromPeer(
+	        const RsGxsId& gxs_id, const RsGxsTunnelId& tunnel_id,
+	        bool am_I_client_side);
+	virtual void notifyTunnelStatus(
+	        const RsGxsTunnelId& tunnel_id, uint32_t tunnel_status);
+	virtual void receiveData(
+	        const RsGxsTunnelId& id, unsigned char* data, uint32_t data_size );
 
     // Utility functions.
     
--- a/libretroshare/src/chat/distantchat.h
+++ b/libretroshare/src/chat/distantchat.h
@@ -30,6 +30,10 @@ class RsGixs ;
 
 static const uint32_t DISTANT_CHAT_AES_KEY_SIZE = 16 ;
 
+/**
+ * Public interface only uses DistandChatPeerId, but internally, this is
+ * converted into a RsGxsTunnelId
+ */
 class DistantChatService: public RsGxsTunnelService::RsGxsTunnelClientService
 {
 public:
--- a/libretroshare/src/pgp/rscertificate.cc
+++ b/libretroshare/src/pgp/rscertificate.cc
@@ -3,8 +3,8 @@
  *                                                                             *
  * libretroshare: retroshare core library                                      *
  *                                                                             *
- * Copyright 2016 Cyril Soler <csoler@users.sourceforge.net>                   *
- * Copyright 2018 Gioacchino Mazzurco <gio@eigenlab.org>                       *
+ * Copyright (C) 2016  Cyril Soler <csoler@users.sourceforge.net>              *
+ * Copyright (C) 2018-2019  Gioacchino Mazzurco <gio@eigenlab.org>             *
  *                                                                             *
  * This program is free software: you can redistribute it and/or modify        *
  * it under the terms of the GNU Lesser General Public License as              *
@@ -55,6 +55,7 @@ static bool is_acceptable_radix64Char(char c)
 {
 	return (c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z') || (c >= '0' && c <= '9') || c == '+' || c == '/' ||  c == '=' ;
 }
+
 
 RsCertificate::~RsCertificate()
 {
@@ -167,23 +168,6 @@ std::string RsCertificate::toStdString() const
 	return out2 ;
 }
 
-RsCertificate::RsCertificate(const std::string& str) :
-    location_name(""), pgp_version("Version: OpenPGP:SDK v0.9"),
-    dns_name(""), only_pgp(true)
-{
-	uint32_t err_code;
-	binary_pgp_key = nullptr;
-
-	if(!initializeFromString(str, err_code))
-	{
-		RsErr() << __PRETTY_FUNCTION__ << " is deprecated because it can "
-		        << "miserably fail like this! str: " << str
-		        << " err_code: " << err_code << std::endl;
-		print_stacktrace();
-		throw err_code;
-	}
-}
-
 RsCertificate::RsCertificate(const RsPeerDetails& Detail, const unsigned char *binary_pgp_block,size_t binary_pgp_block_size)
 	:pgp_version("Version: OpenPGP:SDK v0.9")
 {
@@ -262,6 +246,91 @@ RsCertificate::RsCertificate(const RsPeerDetails& Detail, const unsigned char *b
 	}
 }
 
+
+/*static*/ std::unique_ptr<RsCertificate> RsCertificate::fromMemoryBlock(
+        const RsPeerDetails& Detail, const uint8_t* binary_pgp_block,
+        size_t binary_pgp_block_size )
+{
+	if(binary_pgp_block_size == 0 || binary_pgp_block == nullptr)
+	{
+		RsErr() << __PRETTY_FUNCTION__ << " invalid parameters! " << std::endl;
+		print_stacktrace();
+		return nullptr;
+	}
+
+	std::unique_ptr<RsCertificate> crt(new RsCertificate);
+
+	crt->binary_pgp_key = new uint8_t[binary_pgp_block_size];
+	memcpy(crt->binary_pgp_key, binary_pgp_block, binary_pgp_block_size);
+	crt->binary_pgp_key_size = binary_pgp_block_size;
+
+	if(!Detail.isOnlyGPGdetail)
+	{
+		crt->only_pgp = false;
+		crt->location_id = RsPeerId(Detail.id);
+		crt->location_name = Detail.location;
+
+		if (Detail.isHiddenNode)
+		{
+			crt->hidden_node = true;
+			crt->hidden_node_address = Detail.hiddenNodeAddress;
+			rs_sprintf_append(
+			            crt->hidden_node_address, ":%u", Detail.hiddenNodePort);
+
+			memset(crt->ipv4_internal_ip_and_port, 0, 6);
+			memset(crt->ipv4_external_ip_and_port, 0, 6);
+			crt->dns_name = "";
+		}
+		else
+		{
+			crt->hidden_node = false;
+			crt->hidden_node_address = "";
+
+			try
+			{
+				scan_ip( Detail.localAddr, Detail.localPort,
+				         crt->ipv4_internal_ip_and_port );
+			}
+			catch(...)
+			{
+				RsErr() << __PRETTY_FUNCTION__ << " Invalid LocalAddress: "
+				        << Detail.localAddr << std::endl;
+				memset(crt->ipv4_internal_ip_and_port, 0, 6);
+			}
+
+			try
+			{
+				scan_ip( Detail.extAddr, Detail.extPort,
+				         crt->ipv4_external_ip_and_port );
+			}
+			catch(...)
+			{
+				RsErr() << __PRETTY_FUNCTION__ << " Invalid ExternalAddress: "
+				        << Detail.extAddr << std::endl;
+				memset(crt->ipv4_external_ip_and_port, 0, 6);
+			}
+
+			crt->dns_name = Detail.dyndns;
+
+			for(auto&& ipr : Detail.ipAddressList)
+				crt->mLocators.insert(RsUrl(ipr.substr(0, ipr.find(' '))));
+		}
+	}
+	else
+	{
+		crt->only_pgp = true ;
+		crt->hidden_node = false;
+		crt->hidden_node_address = "";
+		crt->location_id = RsPeerId() ;
+		crt->location_name = "" ;
+		memset(crt->ipv4_internal_ip_and_port, 0, 6);
+		memset(crt->ipv4_external_ip_and_port, 0, 6);
+		crt->dns_name = "";
+	}
+
+	return crt; // Implicit move semantic
+}
+
 void RsCertificate::scan_ip(const std::string& ip_string, unsigned short port,unsigned char *ip_and_port)
 {
 	int d0,d1,d2,d3 ;
@@ -278,8 +347,11 @@ void RsCertificate::scan_ip(const std::string& ip_string, unsigned short port,un
 	ip_and_port[5] =  port        & 0xff ;
 }
 
-bool RsCertificate::initializeFromString(const std::string& instr,uint32_t& err_code)
+/*static*/ std::unique_ptr<RsCertificate> RsCertificate::fromString(
+        const std::string& instr, uint32_t& err_code )
 {
+	std::unique_ptr<RsCertificate> crt(new RsCertificate);
+
 	try
 	{
 		std::string str ;
@@ -293,7 +365,7 @@ bool RsCertificate::initializeFromString(const std::string& instr,uint32_t& err_
 				continue ;
 
 			if(! is_acceptable_radix64Char(instr[i]))
-				return false ;
+				return nullptr;
 
 			str += instr[i] ;
 		}
@@ -308,7 +380,7 @@ bool RsCertificate::initializeFromString(const std::string& instr,uint32_t& err_
 		bool checksum_check_passed = false;
 		unsigned char *buf = bf.data();
 		size_t total_s = 0;
-		only_pgp = true;
+		crt->only_pgp = true;
 		uint8_t certificate_version = 0x00;
 
 		while(total_s < size)
@@ -324,7 +396,7 @@ bool RsCertificate::initializeFromString(const std::string& instr,uint32_t& err_
 			if(total_s > size)
 			{
 				err_code = CERTIFICATE_PARSING_ERROR_SIZE_ERROR ;
-				return false ;
+				return nullptr;
 			}
 
 #ifdef DEBUG_RSCERTIFICATE
@@ -336,44 +408,46 @@ bool RsCertificate::initializeFromString(const std::string& instr,uint32_t& err_
 				certificate_version = buf[0];
 				break;
 			case CERTIFICATE_PTAG_PGP_SECTION:
-				binary_pgp_key = new unsigned char[s];
-				memcpy(binary_pgp_key,buf,s);
-				binary_pgp_key_size = s;
+				crt->binary_pgp_key = new unsigned char[s];
+				memcpy(crt->binary_pgp_key, buf, s);
+				crt->binary_pgp_key_size = s;
 				break;
 			case CERTIFICATE_PTAG_NAME_SECTION:
-				location_name = std::string((char *)buf,s);
+				crt->location_name =
+				        std::string(reinterpret_cast<char*>(buf), s);
 				break;
 			case CERTIFICATE_PTAG_SSLID_SECTION:
-				if(s != location_id.SIZE_IN_BYTES)
+				if(s != crt->location_id.SIZE_IN_BYTES)
 				{
 					err_code = CERTIFICATE_PARSING_ERROR_INVALID_LOCATION_ID;
-					return false;
+					return nullptr;
 				}
-				location_id = RsPeerId(buf);
-				only_pgp = false;
+				crt->location_id = RsPeerId(buf);
+				crt->only_pgp = false;
 				break;
 			case CERTIFICATE_PTAG_DNS_SECTION:
-				dns_name = std::string((char *)buf,s);
+				crt->dns_name = std::string(reinterpret_cast<char*>(buf), s);
 				break;
 			case CERTIFICATE_PTAG_HIDDENNODE_SECTION:
-				hidden_node_address = std::string((char *)buf,s);
-				hidden_node = true;
+				crt->hidden_node_address =
+				        std::string(reinterpret_cast<char*>(buf),s);
+				crt->hidden_node = true;
 				break;
 			case CERTIFICATE_PTAG_LOCIPANDPORT_SECTION:
 				if(s != 6)
 				{
 					err_code = CERTIFICATE_PARSING_ERROR_INVALID_LOCAL_IP;
-					return false;
+					return nullptr;
 				}
-				memcpy(ipv4_internal_ip_and_port,buf,s);
+				memcpy(crt->ipv4_internal_ip_and_port, buf, s);
 				break;
 			case CERTIFICATE_PTAG_EXTIPANDPORT_SECTION:
 				if(s != 6)
 				{
 					err_code = CERTIFICATE_PARSING_ERROR_INVALID_EXTERNAL_IP;
-					return false;
+					return nullptr;
 				}
-				memcpy(ipv4_external_ip_and_port,buf,s);
+				memcpy(crt->ipv4_external_ip_and_port, buf, s);
 				break;
 			case CERTIFICATE_PTAG_CHECKSUM_SECTION:
 			{
@@ -381,7 +455,7 @@ bool RsCertificate::initializeFromString(const std::string& instr,uint32_t& err_
 				{
 					err_code =
 					        CERTIFICATE_PARSING_ERROR_INVALID_CHECKSUM_SECTION;
-					return false;
+					return nullptr;
 				}
 				uint32_t computed_crc =
 				        PGPKeyManagement::compute24bitsCRC(bf.data(),size-5);
@@ -390,18 +464,19 @@ bool RsCertificate::initializeFromString(const std::string& instr,uint32_t& err_
 				if(computed_crc != certificate_crc)
 				{
 					err_code = CERTIFICATE_PARSING_ERROR_CHECKSUM_ERROR;
-					return false;
+					return nullptr;
 				}
 				else checksum_check_passed = true;
 				break;
 			}
 			case CERTIFICATE_PTAG_EXTRA_LOCATOR:
-				mLocators.insert(RsUrl(std::string((char *)buf, s)));
+				crt->mLocators.insert(
+				            RsUrl(std::string(reinterpret_cast<char*>(buf), s)));
 				break;
 			default:
-				std::cerr << "(WW) unknwown PTAG 0x" << std::hex << ptag
-				          << std::dec << " in certificate! Ignoring it."
-				          << std::endl;
+				RsErr() << __PRETTY_FUNCTION__ << " unknwown PTAG 0x"
+				        << std::hex << ptag << std::dec
+				        << " in certificate! Ignoring it." << std::endl;
 				break;
 			}
 
@@ -412,30 +487,28 @@ bool RsCertificate::initializeFromString(const std::string& instr,uint32_t& err_
 		if(!checksum_check_passed)
 		{
 			err_code = CERTIFICATE_PARSING_ERROR_MISSING_CHECKSUM ;
-			return false ;
+			return nullptr;
 		}
 
 		if(certificate_version != CERTIFICATE_VERSION_06)
 		{
 			err_code = CERTIFICATE_PARSING_ERROR_WRONG_VERSION ;
-			return false ;
+			return nullptr;
 		}
 #ifdef DEBUG_RSCERTIFICATE
 		std::cerr << "Certificate is version " << (int)certificate_version << std::endl;
 #endif
 
 		if(total_s != size)	
-			std::cerr << "(EE) Certificate contains trailing characters. Weird." << std::endl;
+			RsErr() << __PRETTY_FUNCTION__ << " Certificate contains trailing "
+			        << "characters. Weird." << std::endl;
 
-		return true ;
+		return crt; // Implicit move semantic
 	}
-	catch(std::exception& e)
+	catch(std::exception&)
 	{
-		if(binary_pgp_key != NULL)
-			delete[] binary_pgp_key ;
-
-		err_code = CERTIFICATE_PARSING_ERROR_SIZE_ERROR ;
-		return false ;
+		err_code = CERTIFICATE_PARSING_ERROR_SIZE_ERROR;
+		return nullptr;
 	}
 }
 
@@ -456,6 +529,7 @@ std::string RsCertificate::ext_ip_string() const
 	os << (int)ipv4_external_ip_and_port[0] << "." << (int)ipv4_external_ip_and_port[1] << "." << (int)ipv4_external_ip_and_port[2] << "." << (int)ipv4_external_ip_and_port[3] ;
 	return os.str() ;
 }
+
 std::string RsCertificate::loc_ip_string() const
 {
 	std::ostringstream os ;
@@ -473,28 +547,21 @@ unsigned short RsCertificate::loc_port_us() const
 	return (int)ipv4_internal_ip_and_port[4]*256 + (int)ipv4_internal_ip_and_port[5] ;
 }
 
-bool RsCertificate::cleanCertificate(const std::string& input,std::string& output,Format& format,int& error_code,bool check_content)
+bool RsCertificate::cleanCertificate(
+        const std::string& input, std::string& output, Format& format,
+        int& error_code, bool check_content )
 {
 	if(cleanCertificate(input,output,error_code))
 	{
-		format = RS_CERTIFICATE_RADIX ;
-
-        if(!check_content)
-            return true ;
-
-        try
-        {
-			RsCertificate c(input) ;
-            return true ;
-        }
-        catch(uint32_t err_code)
-        {
-            error_code = err_code ;
-            return false;
-        }
+		format = RS_CERTIFICATE_RADIX;
+		if(!check_content) return true;
+		uint32_t errCode;
+		auto crt = RsCertificate::fromString(input, errCode);
+		error_code = static_cast<int>(errCode);
+		return crt != nullptr;
 	}
 
-	return false ;
+	return false;
 }
 
 std::string RsCertificate::armouredPGPKey() const
--- a/libretroshare/src/pgp/rscertificate.cc
+++ b/libretroshare/src/pgp/rscertificate.cc
@@ -31,6 +31,7 @@
 #include "rscertificate.h"
 #include "util/rsstring.h"
 #include "util/stacktrace.h"
+#include "util/rsdebug.h"
 
 //#define DEBUG_RSCERTIFICATE
 
@@ -175,9 +176,9 @@ RsCertificate::RsCertificate(const std::string& str) :
 
 	if(!initializeFromString(str, err_code))
 	{
-		std::cerr << __PRETTY_FUNCTION__ << " is deprecated because it can "
-		          << "miserably fail like this! str: " << str
-		          << " err_code: " << err_code << std::endl;
+		RsErr() << __PRETTY_FUNCTION__ << " is deprecated because it can "
+		        << "miserably fail like this! str: " << str
+		        << " err_code: " << err_code << std::endl;
 		print_stacktrace();
 		throw err_code;
 	}
@@ -186,8 +187,13 @@ RsCertificate::RsCertificate(const std::string& str) :
 RsCertificate::RsCertificate(const RsPeerDetails& Detail, const unsigned char *binary_pgp_block,size_t binary_pgp_block_size)
 	:pgp_version("Version: OpenPGP:SDK v0.9")
 {
-	if(binary_pgp_block_size == 0 || binary_pgp_block == NULL)
-		throw std::runtime_error("Cannot init a certificate with a void key block.") ;
+	if(binary_pgp_block_size == 0 || binary_pgp_block == nullptr)
+	{
+		RsErr() << __PRETTY_FUNCTION__ << " is deprecated because it can "
+		        << "miserably fail like this! " << std::endl;
+		print_stacktrace();
+		throw std::runtime_error("Cannot init a certificate with a void key block.");
+	}
 
 	binary_pgp_key = new unsigned char[binary_pgp_block_size] ;
 	memcpy(binary_pgp_key,binary_pgp_block,binary_pgp_block_size) ;
--- a/libretroshare/src/pgp/rscertificate.h
+++ b/libretroshare/src/pgp/rscertificate.h
@@ -54,7 +54,7 @@ public:
 	bool initializeFromString(const std::string& str, uint32_t& errCode);
 
 	/// Constructs from binary gpg key, and RsPeerDetails.
-	RsCertificate( const RsPeerDetails& details,
+	RS_DEPRECATED RsCertificate( const RsPeerDetails& details,
 	               const unsigned char *gpg_mem_block,
 	               size_t gpg_mem_block_size );
 
--- a/libretroshare/src/pgp/rscertificate.h
+++ b/libretroshare/src/pgp/rscertificate.h
@@ -3,7 +3,8 @@
  *                                                                             *
  * libretroshare: retroshare core library                                      *
  *                                                                             *
- * Copyright 2016 Cyril Soler <csoler@users.sourceforge.net>                   *
+ * Copyright (C) 2016  Cyril Soler <csoler@users.sourceforge.net>              *
+ * Copyright (C) 2018-2019  Gioacchino Mazzurco <gio@eigenlab.org>             *
  *                                                                             *
  * This program is free software: you can redistribute it and/or modify        *
  * it under the terms of the GNU Lesser General Public License as              *
@@ -23,9 +24,11 @@
 
 #include "retroshare/rstypes.h"
 #include "util/rsurl.h"
+#include "util/rsmemory.h"
 
 #include <set>
 #include <string>
+#include <memory>
 
 struct RsPeerDetails;
 
@@ -35,30 +38,30 @@ public:
 	typedef enum { RS_CERTIFICATE_OLD_FORMAT, RS_CERTIFICATE_RADIX } Format;
 
 	/**
-	 * @brief Costruct an empty certificate, use toghether with
-	 * if(initializeFromString) for safe certificate radix string parsing
+	 * @brief Create certificate object from certificate string
+	 * @param[in] str radix format string
+	 * @param[out] errorCode Optional storage for eventual error code,
+	 *	meaningful only on failure
+	 * @return nullptr on failure, pointer to the generated certificate
+	 *	otherwise
 	 */
-	RsCertificate() :
-	    ipv4_external_ip_and_port{0,0,0,0,0,0},
-	    ipv4_internal_ip_and_port{0,0,0,0,0,0},
-	    binary_pgp_key(nullptr), binary_pgp_key_size(0),
-	    pgp_version("Version: OpenPGP:SDK v0.9"), only_pgp(true),
-	    hidden_node(false) {}
+	static std::unique_ptr<RsCertificate> fromString(
+	        const std::string& str,
+	        uint32_t& errorCode = RS_DEFAULT_STORAGE_PARAM(uint32_t) );
 
 	/**
-	 * @brief Initialize from certificate string
-	 * @param[in] str radix format string
-	 * @param[out] errCode storage for eventual error code
-	 * @return false on failure, true otherwise
+	 * @brief Create certificate object from details and PGP memory block
+	 * @param[in] details peer details
+	 * @param[in] binary_pgp_block pointer to PGP memory block
+	 * @param[in] binary_pgp_block_size size of PGP memory block
+	 * @return nullptr on failure, pointer to the generated certificate
+	 *	otherwise
 	 */
-	bool initializeFromString(const std::string& str, uint32_t& errCode);
+	static std::unique_ptr<RsCertificate> fromMemoryBlock(
+	        const RsPeerDetails& details, const uint8_t* binary_pgp_block,
+	        size_t binary_pgp_block_size);
 
-	/// Constructs from binary gpg key, and RsPeerDetails.
-	RS_DEPRECATED RsCertificate( const RsPeerDetails& details,
-	               const unsigned char *gpg_mem_block,
-	               size_t gpg_mem_block_size );
-
-	virtual ~RsCertificate();
+	~RsCertificate();
 
 	/// Convert to certificate radix string
 	std::string toStdString() const;
@@ -86,10 +89,12 @@ public:
 
 	/**
 	 * @deprecated using this costructor may raise exception that cause
-	 *	crash if not handled, use empty constructor + if(initFromString) for a
-	 *	safer behaviour.
+	 *	crash if not handled.
 	 */
-	RS_DEPRECATED explicit RsCertificate(const std::string& input_string);
+	RS_DEPRECATED_FOR("RsCertificate::fromMemoryBlock(...)")
+	RsCertificate( const RsPeerDetails& details,
+	               const unsigned char *gpg_mem_block,
+	               size_t gpg_mem_block_size );
 
 private:
 	// new radix format
@@ -104,6 +109,14 @@ private:
 
 	RsCertificate(const RsCertificate&) {} /// non copy-able
 	const RsCertificate& operator=(const RsCertificate&); /// non copy-able
+
+	/// @brief Costruct an empty certificate
+	RsCertificate() :
+	    ipv4_external_ip_and_port{0,0,0,0,0,0},
+	    ipv4_internal_ip_and_port{0,0,0,0,0,0},
+	    binary_pgp_key(nullptr), binary_pgp_key_size(0),
+	    pgp_version("Version: OpenPGP:SDK v0.9"), only_pgp(true),
+	    hidden_node(false) {}
 
 	unsigned char ipv4_external_ip_and_port[6];
 	unsigned char ipv4_internal_ip_and_port[6];
--- a/libretroshare/src/pqi/authssl.cc
+++ b/libretroshare/src/pqi/authssl.cc
@@ -3,7 +3,8 @@
  *                                                                             *
  * libretroshare: retroshare core library                                      *
  *                                                                             *
- * Copyright 2004-2008 by Robert Fernie, Retroshare Team.                      *
+ * Copyright (C) 2004-2008  Robert Fernie, Retroshare Team.                    *
+ * Copyright (C) 2019  Gioacchino Mazzurco <gio@altermundi.net>                *
  *                                                                             *
  * This program is free software: you can redistribute it and/or modify        *
  * it under the terms of the GNU Lesser General Public License as              *
@@ -19,6 +20,11 @@
  * along with this program. If not, see <https://www.gnu.org/licenses/>.       *
  *                                                                             *
  *******************************************************************************/
+
+#include <chrono>
+#include <thread>
+#include <memory>
+
 #ifdef WINDOWS_SYS
 #include "util/rswin.h"
 #endif // WINDOWS_SYS
@@ -33,8 +39,8 @@
 #include "util/rsstring.h"
 #include "pgp/pgpkeyutil.h"
 
-#include "retroshare/rspeers.h" // for RsPeerDetails structure 
-#include "retroshare/rsids.h" // for RsPeerDetails structure
+#include "retroshare/rspeers.h"
+#include "retroshare/rsids.h"
 #include "rsserver/p3face.h" 
 
 /******************** notify of new Cert **************************/
@@ -63,8 +69,8 @@ const uint32_t RS_SSL_HANDSHAKE_DIAGNOSTIC_WRONG_SIGNATURE_VERSION     = 0x0a ;
  ***/
 
 // initialisation du pointeur de singleton
-AuthSSL *AuthSSL::instance_ssl = NULL;
-static pthread_mutex_t *mutex_buf = NULL;
+AuthSSL* AuthSSL::instance_ssl = nullptr;
+static pthread_mutex_t* mutex_buf = nullptr;
 
 struct CRYPTO_dynlock_value
 {
@@ -214,29 +220,13 @@ void tls_cleanup()
 	}
 }
 
-/* hidden function - for testing purposes() */
-void AuthSSL::setAuthSSL_debug(AuthSSL *newssl)
-{
-	instance_ssl = newssl;
-}
-
 void AuthSSL::AuthSSLInit()
-{
-	if (instance_ssl == NULL)
-	{
-		instance_ssl = new AuthSSLimpl();
-	}
-}
-  
-AuthSSL *AuthSSL::getAuthSSL()
-{
-	return instance_ssl;
-}
+{ if (!instance_ssl) instance_ssl = new AuthSSLimpl(); }
 
-AuthSSL::AuthSSL()
-{
-	return;
-}
+AuthSSL* AuthSSL::getAuthSSL()
+{ return instance_ssl; }
+
+AuthSSL::~AuthSSL() {}
 
   
 /********************************************************************************/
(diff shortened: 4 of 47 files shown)


defnax:
its great that you work on this direction, add friend wasnt yet easy

G10h4ck:
ATM retroshare-gui doesn't take adavantage of the features introduced in this pull request. Pending PGP peers (peers added through short invite) are not showed in the network tab and there is no way to get a short invite in retroshare-gui. First step to add support for this feature in retroshare-gui is to implemet RsEvents handling in retroshare-gui which is necessary to support the new features.

As the PR is already huge only with libretroshare modification, I would prefer that retroshare-gui related enanchements comes in another PR.

I have tested the functionalities introduced in this PR through JSON API, so I didn't need to modify retroshare-gui to test it.

libretroshare/src/pgp/rscertificate.h
```diff
@@ -54,7 +54,7 @@ class RsCertificate
 	bool initializeFromString(const std::string& str, uint32_t& errCode);
 
 	/// Constructs from binary gpg key, and RsPeerDetails.
-	RsCertificate( const RsPeerDetails& details,
+	RS_DEPRECATED RsCertificate( const RsPeerDetails& details,
```
csoler: only deprecate if you have something in replacement. It's of course required to call that constructor in a try{}catch() block. Not very elegant but safe, and I'm not sure there is a proper way to recover from an error in the constructor. You'd need to add a isValid() member to the certificate object, etc. That's not so nice either.

G10h4ck: I forgot to add a function to replace that constructor like I did with the constructor from sting certificate, I am doing it right now ;)

libretroshare/src/pqi/authssl.cc
```diff
@@ -989,411 +1002,353 @@ X509 *AuthSSLimpl::SignX509ReqWithGPG(X509_REQ *req, long /*days*/)
...
 
 	/* verify GPG signature */
-
 	/*** NOW The Manual signing bit (HACKED FROM asn1/a_sign.c) ***/
 
 #if OPENSSL_VERSION_NUMBER < 0x10100000L || defined(LIBRESSL_VERSION_NUMBER)
 	int (*i2d)(X509_CINF*, unsigned char**) = i2d_X509_CINF;
-	ASN1_BIT_STRING *signature = x509->signature;
-	X509_CINF *data = x509->cert_info;
+	ASN1_BIT_STRING* signature = x509->signature;
```
csoler, line 1026: WTF?

G10h4ck: semantically this change nothing, but * actually belong to the type so it got grouped with it, this line change is not strictly needed

libretroshare/src/pqi/authssl.cc
```diff
@@ -989,411 +1002,353 @@ X509 *AuthSSLimpl::SignX509ReqWithGPG(X509_REQ *req, long /*days*/)
...
-		RsPeerId peerIdstr ;
-		getX509id(x509, peerIdstr) ;
-
-		std::string fpr = pd.fpr.toStdString();
-		std::cerr << "Verified " << sigtypestring << " signature of certificate " << peerIdstr << ", Version " << std::hex << certificate_version
-		          << std::dec << " using PGP key " ;
-		for(uint32_t i=0;i<fpr.length();i+=4)
-			std::cerr << fpr.substr(i,4) << " " ;
-		std::cerr << std::endl;
+		Dbg1() << __PRETTY_FUNCTION__ << " Verified: " << sigtypestring
```
csoler, line 1162: the ostream output of fingerprint has spaces too? I dont think so. That's why I used an explicit output method

libretroshare/src/pqi/authssl.cc
sehraf: nope
`int* a, b; ` != `int *a, *b;`

libretroshare/src/pqi/authssl.h
```diff
@@ -42,109 +42,191 @@
...
-        std::string org;
-        std::string email;
+	/* certificate parameters */
+	RsPeerId id;
+	std::string name;
+	std::string location;
+	std::string org;
 
-        RsPgpId issuer;
-        PGPFingerprintType fpr;
```
csoler, line 71: Why would you like to remove the fingerprint? It's going to be useful. Please keep it for a while. After we switch to 160 bits IDs this is going to help.

libretroshare/src/pqi/authssl.cc
G10h4ck: Ok I'll specialize it so we will always pretty print fingerprints

libretroshare/src/pqi/authssl.h
G10h4ck: It was never used and was misleading to keep it there because it was suggesting that value was populated, while it was always 0. I think it is better to re-add it when we populate it not now.

libretroshare/src/pqi/authssl.cc
G10h4ck: done

libretroshare/src/pgp/rscertificate.h
G10h4ck: done

felisucoibi:
step by step

libretroshare/src/retroshare/rsdisc.h
```diff
@@ -1,9 +1,10 @@
 /*******************************************************************************
- * libretroshare/src/retroshare: rsdht.h                                       *
+ * RetroShare remote peers gossip discovery                                    *
```
csoler, line 2: normally we put the file path here. So it should be libretroshare/src/retroshare/rsdisc.h

G10h4ck: Many times the file path/name there is wrong in our code, also because it doesn't add any information, I do prefer to put a short message descriptive of the file there, but if the file name there is some Debian obnoxious policy I can change that to the path of the file again.

libretroshare/src/chat/distantchat.h
```diff
@@ -33,9 +33,6 @@ static const uint32_t DISTANT_CHAT_AES_KEY_SIZE = 16 ;
 class DistantChatService: public RsGxsTunnelService::RsGxsTunnelClientService
 {
 public:
-    // So, public interface only uses DistandChatPeerId, but internally, this is converted into a RsGxsTunnelService::RsGxsTunnelId
```
csoler, line 36: This is true. Why did you remove it?

libretroshare/src/pgp/rscertificate.cc
```diff
@@ -330,52 +408,54 @@ bool RsCertificate::initializeFromString(const std::string& instr,uint32_t& err_
...
-				memcpy(binary_pgp_key,buf,s);
-				binary_pgp_key_size = s;
+				crt->binary_pgp_key = new unsigned char[s];
+				memcpy(crt->binary_pgp_key, buf, s);
+				crt->binary_pgp_key_size = s;
 				break;
 			case CERTIFICATE_PTAG_NAME_SECTION:
-				location_name = std::string((char *)buf,s);
+				crt->location_name =
+				        std::string(reinterpret_cast<char*>(buf), s);
```
csoler, line 417: I would have used a static_cast, since it offers additional compilation-time checks. Isn't it possible here?

libretroshare/src/pgp/rscertificate.cc
```diff
@@ -406,30 +487,28 @@ bool RsCertificate::initializeFromString(const std::string& instr,uint32_t& err_
...
+			RsErr() << __PRETTY_FUNCTION__ << " Certificate contains trailing "
+			        << "characters. Weird." << std::endl;
 
-		return true ;
+		return crt; // Implicit move semantic
 	}
-	catch(std::exception& e)
+	catch(std::exception&)
 	{
-		if(binary_pgp_key != NULL)
```
csoler, line 428: Why remove this? It cleans the memory that was allocated. I would have added the size to be 0.

libretroshare/src/pqi/p3linkmgr.cc
```diff
@@ -1454,21 +1456,8 @@ void p3LinkMgrIMPL::peerConnectRequest(
...
 /*******************************************************************/
        /*************** External Control ****************/
-bool   p3LinkMgrIMPL::retryConnect(const RsPeerId &id)
-{
-	/* push all available addresses onto the connect addr stack */
-#ifdef LINKMGR_DEBUG
-	std::cerr << "p3LinkMgrIMPL::retryConnect() id: " << id << std::endl;
-#endif
-
-#ifndef P3CONNMGR_NO_TCP_CONNECTIONS
```
csoler: This was actually useful for debugging UDP connections

libretroshare/src/pqi/p3linkmgr.cc
```diff
@@ -1648,79 +1674,59 @@ bool p3LinkMgrIMPL::retryConnectTCP(const RsPeerId &id)
...
-	std::cerr << sockaddr_storage_tostring(addr);
-	std::cerr << ", " << age << ")";
-	std::cerr << std::endl;
-#endif
+	Dbg3() << __PRETTY_FUNCTION__ << " addr: " << addr << std::endl;
 
-	/*
-	 * if it is old - quick rejection 
-	 */
-	if (age > MAX_TCP_ADDR_AGE)
```
csoler, line 1674: I'm wondering why this was here. I suppose you too ;-) I think that old addresses in the contact address list were discarded by this.

libretroshare/src/pqi/p3linkmgr.cc
```diff
@@ -2010,22 +2014,17 @@ bool  p3LinkMgrIMPL::addAddressIfUnique(std::list<peerConnectAddress> &addrList,
 
 
 
-bool  p3LinkMgrIMPL::locked_ConnectAttempt_Complete(peerConnectState *peer)
+bool p3LinkMgrIMPL::locked_ConnectAttempt_Complete(peerConnectState* peer)
 {
-
 	/* flag as last attempt to prevent loop */
 	//add a random perturbation between 0 and 2 sec.
-	peer->lastattempt = time(NULL) + rand() % MAX_RANDOM_ATTEMPT_OFFSET;
```
csoler, line 2018: We should remove rand() everywhere. Just replace it by RSRandom::random_u32(). It's probably a good thing that connect attempts cannot be predicted.

libretroshare/src/pqi/p3peermgr.cc
```diff
@@ -941,18 +939,6 @@ bool p3PeerMgrIMPL::addFriend(const RsPeerId& input_id, const RsPgpId& input_gpg
 			return true;
 		}
 
-		//Authentication is now tested at connection time, we don't store the ssl cert anymore
```
csoler, line 944: /!\  Removing this is potentially harmful: this method is used during discovery when we receive new locations from known friends. It's very important to check that if the supplied PGP id is not null, the friend is not added!! Otherwise it would be potentially easy to sneak in some friends using discovery.

I think that you should create a second method for addign without PGP id, and make sure that the information that no PGP id was supplied *on purpose* is written somwhere, so that this method that accepts no PGP id is not called from a different place by mistake (or attack).

libretroshare/src/pqi/p3peermgr.cc
```diff
@@ -1101,6 +1087,72 @@ bool p3PeerMgrIMPL::removeFriend(const RsPgpId &id)
...
+		/* If in mOthersList -> move over */
+		auto it = mOthersList.find(sslId);
+		if (it != mOthersList.end())
+		{
+			pstate = it->second;
+			mOthersList.erase(it);
+		}
+
+		pstate.id = sslId;
+		if(!dt.gpg_id.isNull())  pstate.gpg_id = dt.gpg_id;
```
csoler, line 1122: if PGP id is null, you should return with an error, since this should not happen.
This is where you could check that the PGP id that was stored as a fingerprint really corresponds to the PGP key that is received.

libretroshare/src/pqi/p3peermgr.cc
```diff
@@ -1101,6 +1087,72 @@ bool p3PeerMgrIMPL::removeFriend(const RsPgpId &id)
...
+		if(!dt.fpr.isNull())     pstate.mPgpFingerprint = dt.fpr;
```
csoler, line 1123: same here

libretroshare/src/pqi/pqiperson.h
```diff
@@ -73,23 +82,38 @@ class pqiconnect : public pqiQoSstreamer, public NetInterface
 
 protected:
 	NetBinInterface *ni;
+
+#if defined(RS_DEBUG_PQICONNECT) && RS_DEBUG_PQICONNECT == 1
```
csoler, line 86: These blocks repeated in every class really look bad. Is there a way to hide them using some kind of macro?

libretroshare/src/pqi/pqissllistener.cc
```diff
@@ -574,55 +575,19 @@ int pqissllistenbase::closeConnection(int fd, SSL *ssl)
...
-  	pqioutput(PQL_DEBUG_BASIC, pqissllistenzone, 
-	  "pqissllistenbase::Extract_Failed_SSL_Certificate() Have Peer Cert - Registering");
-
-	{
-		std::string out;
-		out += "pqissllistenbase::Extract_Failed_SSL_Certificate() from: ";
-		out += sockaddr_storage_tostring(info.addr);
-		out += " Passing Cert to AuthSSL() for analysis";
-		std::cerr << out << std::endl;
+	X509* peercert = SSL_get_peer_certificate(info.ssl);
```
csoler, line 585: be careful with pointers that are extracted out of SSL structures. They may not be valid when the handshake is finished. Is SSL_get_peer_certificate making a deep copy of the pointer? Sometimes you need to go and see the SSL code to know the answer...

libretroshare/src/pqi/pqissllistener.cc
```diff
@@ -830,91 +795,38 @@ int pqissllistener::status()
...
-
-	// Let connected one through as well! if ((npc == NULL) || (npc -> Connected()))
-	if (!certOk)
-	{
-  	        pqioutput(PQL_WARNING, pqissllistenzone, 
-		 "pqissllistener::completeConnection() registerCertificate Failed!");
-
-		// bad - shutdown.
-		// pqissllistenbase will shutdown!
-		X509_free(peercert);
```
csoler, line 870: Removing this is likely to cause a memory leak

libretroshare/src/pqi/pqissllistener.cc
```diff
@@ -830,91 +795,38 @@ int pqissllistener::status()
...
-		/* now need to get GPG id too */
-#if OPENSSL_VERSION_NUMBER < 0x10100000L || defined(LIBRESSL_VERSION_NUMBER)
-		RsPgpId pgpid(std::string(getX509CNString(peercert->cert_info->issuer)));
-#else
-		RsPgpId pgpid(std::string(getX509CNString(X509_get_issuer_name(peercert))));
-#endif
-		mPeerMgr->addFriend(newPeerId, pgpid);
-	
-		X509_free(peercert);
-		return -1;
```
csoler, line 917: If that works, it's good, because ti will avoid the friend to attempt to connect twice. Have you tested this??

libretroshare/src/pqi/pqissllistener.cc
```diff
@@ -830,91 +795,38 @@ int pqissllistener::status()
...
+		mPeerMgr->addFriend(connPeerId, pgpId);
```
csoler, line 829: /!\  No!!! You're adding a friend here without testing the PGP signature. You need to call CheckCertificate (Or ValidateCertificate) before doing this! There's no reason at this point that the issuer field actually corresponds to the signature in the handshake.

libretroshare/src/retroshare/rsdisc.h
csoler: Well, it was just to be consistent wth the remaining 99% of the code ;-)

libretroshare/src/retroshare/rsgxsifacetypes.h
```diff
@@ -34,10 +34,7 @@
 #include "serialiser/rstypeserializer.h"
 #include "util/rstime.h"
 
-typedef GXSGroupId   RsGxsGroupId;
```
csoler, line 37: This is actually used a lot in the code. Why remove it??

libretroshare/src/retroshare/rsids.h
```diff
@@ -19,161 +20,219 @@
...
+ * ID Types with different lengths are not convertible even using explicit
+ * constructor and compilation would fail if that is attempted.
+ *
+ * Warning: never store references to a t_RsGenericIdType accross threads, since
+ * the cached string convertion is not thread safe.
+ */
+template<uint32_t ID_SIZE_IN_BYTES, bool UPPER_CASE, RsGenericIdType UNIQUE_IDENTIFIER>
+struct t_RsGenericIdType
+{
+	using Id_t = t_RsGenericIdType<ID_SIZE_IN_BYTES,UPPER_CASE,UNIQUE_IDENTIFIER>;
```
csoler, line 83: sometimes you just rewrite for the pleasure of rewriting...and at the expense of me spending more time to read .

libretroshare/src/retroshare/rsids.h
```diff
@@ -19,161 +20,219 @@
...
+	/// Explicit constructor from a hexadecimal string
+	explicit t_RsGenericIdType(const std::string& hex_string);
+
+	/**
+	 * @brief Construct from a buffer of at least the size of ID_SIZE_IN_BYTES
+	 * @param[in] buff pointer to the buffer
+	 * @param[in] buffSize buffer size
+	 * @return nullptr on failure pointer to the costructed ID otherwise
+	 */
+	static std::unique_ptr<Id_t> fromBuffer(
```
csoler, line 100: Sorry but that goes beyond what is acceptable. I dont want to have to handle pointers everytime I create an ID from a block in memory. That is more costly, harder to read and ultimately more complicated.

libretroshare/src/retroshare/rsids.h
```diff
@@ -19,161 +20,219 @@
...
-			memcpy(bytes,&((uint8_t*)data)[offset],SIZE_IN_BYTES) ;
-			offset += SIZE_IN_BYTES ;
-			return true ;
-		}
-	private:
-		std::string toStdString(bool upper_case) const ;
+	bool serialise(void* data,uint32_t pktsize,uint32_t& offset) const
+	{
+		if(offset + SIZE_IN_BYTES > pktsize) return false;
+		memmove( &(reinterpret_cast<uint8_t*>(data))[offset],
```
csoler, line 204: no need for memmove here. If the areas overlap, then something's already wrong in the code!

libretroshare/src/retroshare/rsids.h
```diff
@@ -19,161 +20,219 @@
...
+		         SIZE_IN_BYTES );
+		offset += SIZE_IN_BYTES;
+		return true;
+	}
 
-		unsigned char bytes[ID_SIZE_IN_BYTES] ;
+	/** Explicit constructor from a byte array. The array must have size at
+	 * least ID_SIZE_IN_BYTES
+	 * @deprecated This is too dangerous!
+	 * Nothing prevent a buffer of wrong size being passed at runtime!
```
csoler, line 223: True, but it is much less costly and more handy than the unique_ptr stuff. When used in low level code, it's probably preferable.

libretroshare/src/retroshare/rspeers.h
```diff
@@ -476,9 +493,20 @@ class RsPeers
 	 * @param[in] flags service permissions flag
 	 * @return false if error occurred, true otherwise
 	 */
-	virtual bool addFriend( const RsPeerId &sslId, const RsPgpId& gpgId,
+	virtual bool addFriend( const RsPeerId& sslId, const RsPgpId& gpgId,
 	                        ServicePermissionFlags flags = RS_NODE_PERM_DEFAULT ) = 0;
 
+	/**
+	 * @brief Add trusted node pending PGP key
```
csoler, line 500: As said earlier, optional params here do not make sense. This method should *only* be used in a specific case (because it is quite dangerous) so it should expect a static list of parameters.

libretroshare/src/retroshare/rspeers.h
```diff
@@ -599,6 +627,24 @@ class RsPeers
...
+	 * @param[in] formatRadix true to get in base64 format false to get URL.
+	 * @param[in] includeFingerprint true to include PGP fingerprint in the URL.
+	 * @param[in] bareBones true to get smalles invite, which miss also
+	 *	the information necessary to attempt an outgoing connection, but still
+	 *	enough to accept an incoming one.
+	 * @return false if error occurred, true otherwise
+	 */
+	virtual bool getShortInvite(
+	        std::string& invite, const RsPeerId& sslId = RsPeerId(),
+	        bool formatRadix = false, bool includeFingerprint = false,
```
csoler, line 645: Same here. Why have options in the short invite? The format should be decided and enforced.

libretroshare/src/rsitems/rsdiscovery2items.h
```diff
@@ -31,17 +31,24 @@
...
 
+// TODO: port to RsGossipDiscoveryItemType
 const uint8_t RS_PKT_SUBTYPE_DISC_PGP_LIST           = 0x01;
 const uint8_t RS_PKT_SUBTYPE_DISC_PGP_CERT           = 0x02;
 const uint8_t RS_PKT_SUBTYPE_DISC_CONTACT_deprecated = 0x03;
-const uint8_t RS_PKT_SUBTYPE_DISC_SERVICES           = 0x04;
 const uint8_t RS_PKT_SUBTYPE_DISC_CONTACT            = 0x05;
 const uint8_t RS_PKT_SUBTYPE_DISC_IDENTITY_LIST      = 0x06;
 
+enum class RsGossipDiscoveryItemType : uint8_t
```
csoler, line 41: That is typically a case where mixing new style code with old style code is wrong: it makes it hard to see that these two IDs belogn to the list above. Please use the const uint8_t for them as well.

libretroshare/src/rsitems/rsdiscovery2items.h
```diff
@@ -156,38 +160,38 @@ class RsDiscIdentityListItem: public RsDiscItem
 
     std::list<RsGxsId> ownIdentityList ;
 };
-#if 0
-class RsDiscServicesItem: public RsDiscItem
+
+struct RsGossipDiscoveryInviteItem : RsDiscItem
```
csoler, line 164: I dont really understand that name. Why the "Gossip" thing? YOure just sending a certificate here, right?

libretroshare/src/rsserver/p3peers.cc
```diff
@@ -248,6 +248,9 @@ bool    p3Peers::isFriend(const RsPeerId &ssl_id)
         return mPeerMgr->isFriend(ssl_id);
 }
 
+bool p3Peers::isFriendPendingPgp(const RsPeerId& sslId)
+{ return isFriend(sslId) && !isPgpFriend(getGPGId(sslId)); }
```
csoler, line 252: /!\ Why dont you use a flag in the Peer structure to mark the SSL id as not yet validated using PGP? That would prevent the risk of calling isPgpFriend on a Pgp id that is actually not even yet known to be the one that will sign the handshake.
Typical attack vector: I send you a certificate with (SSL Id, PGP if of a friend of yours). This method will class the short invite as a true friend without any checking first.
The safe way to do this is to base the "friendship level" on a flag that is set when you add the friend from short invite and reset after the PGP key is received and actually validates the certificate.

libretroshare/src/rsserver/p3peers.cc
```diff
@@ -1104,10 +1111,139 @@ bool p3Peers::GetPGPBase64StringAndCheckSum(	const RsPgpId& gpg_id,
...
+        bool includeFingerprint, bool bareBones )
+{
+	const RsPeerId& sslId(pSslId.isNull() ? getOwnId() : pSslId);
+
+	RsPeerDetails dts;
+	if (!getPeerDetails(sslId, dts)) return false;
+
+	RsUrl inviteUrl;
+	inviteUrl
+	        .setScheme("rs")
```
csoler, line 1125: why not use the classical retroshare:// ? It exists already and the GUI will recognize it. Furthermore I dont see the point in handlign urls in libretroshare. There's nothing to click in libretroshare. So you should probably keep the invite as a base64 string and let the GUI or any interface make a link out of it. Otherwise you'll have to parse html links  inside libretroshare (and without Qt!!) which is a pain to do securely.

libretroshare/src/rsserver/p3peers.cc
```diff
@@ -1104,10 +1111,139 @@ bool p3Peers::GetPGPBase64StringAndCheckSum(	const RsPgpId& gpg_id,
...
+	return true;
+}
+
 bool p3Peers::acceptInvite( const std::string& invite,
                             ServicePermissionFlags flags )
 {
 	if(invite.empty()) return false;
+	if(acceptLongInvite(invite, flags)) return true;
+
+	RsUrl url(invite);
```
csoler, line 1165: ok, same here. Just remve the url stuff and only keep the invite to be the base64 string.

libretroshare/src/services/p3discovery2.cc
```diff
@@ -243,87 +236,80 @@ void p3discovery2::removeFriend(const SSLID &sslId)
...
+		++nhandled;
 
 #ifdef P3DISC_DEBUG
 		std::cerr << "p3discovery2::handleIncoming() Received Message!" << std::endl;
 		item -> print(std::cerr);
 		std::cerr  << std::endl;
 #endif
 
-		if (nullptr != (contact = dynamic_cast<RsDiscContactItem *> (item)))
+		if((contact = dynamic_cast<RsDiscContactItem *>(item)))
```
csoler, line 274: Well, Phenom will probably complain about this, and I quite agree with him for once: this syntaxe is very misleading. The NULL != (a=b)  was here on purpose. The compiler will simplify it for you.

libretroshare/src/services/p3discovery2.cc
```diff
@@ -243,87 +236,80 @@ void p3discovery2::removeFriend(const SSLID &sslId)
...
 				recvPGPCertificateRequest(pgplist->PeerId(), pgplist);
-			}
-			else
-			{
-                		delete item ;
-			}
+			else delete item;
+		}
+		else if((invite = dynamic_cast<RsGossipDiscoveryInviteItem*>(item)))
+			recvInvite(std::unique_ptr<RsGossipDiscoveryInviteItem>(invite));
```
csoler, line 296: so the delete item is implicit in this line? It's explicit everywhere else => hard to read, raises questions, etc. Please don't. It's cool a C++11 feature, but it makes the code harder to read. Furthermore, the way it is written (with explicit delete) shows that the memory ownership is not transferred to the function.

libretroshare/src/services/p3discovery2.cc
```diff
@@ -1270,7 +1238,72 @@ bool p3discovery2::setPeerVersion(const SSLID &peerId, const std::string &versio
...
+			       << " AUTHSSL_CONNECTION_AUTENTICATION event: " << event
+			       << std::endl;
+
+			const RsPeerId& sslId = evt.mSslId;
+			RsThread::async([sslId]()
+			{ /* Make sure the connection is ready before requesting invite. */
+				auto timeout = std::chrono::system_clock::now() +
+				        std::chrono::seconds(15);
+				while ( rsDisc && rsPeers &&
+				        std::chrono::system_clock::now() < timeout )
```
csoler, line 1281: This is not robust. You should use a queue and tick regularly to send the invite requests. You cannot ensure that your timeout will not be reached because of all sorts of delays that happen. So what you're doing is creatinga thread that will wait for a time that is either too short, or too long making potentially many threads to wait for the right conditions to continue, and eating file descriptors.

csoler (changes requested):

libretroshare/src/retroshare/rsids.h
G10h4ck: this line improve the readability of the following code a lot, without all those template huge line around that make it difficult to read in my little laptop

libretroshare/src/retroshare/rspeers.h
G10h4ck: It is convenient to put more or less information/fields depending on what is the medium of transmission, we have someway similar mechanism with in long invites. About the format we already do something similar with long invites too, depending on the medium of transmission it might be more convenient to have it in URI format or in radix64 format. This implement it at libretroshare level instead of delegating the implementation to the GUI with consequent problems in keeping compatibility between different GUI (when I programmed the Qml app I had an hard time to keep RS links compatibles between RS GUI and Qml App because of this)

libretroshare/src/rsitems/rsdiscovery2items.h
G10h4ck: well either we write well the new code and leave the conversion to enum for the other items for another PR, or we convert all the items to the enum, we already agreed that using all those integer types around is bad, because it make very difficult to understand what is passed around. Converting from enum to const uint8_t is walking back, and I don't do that...

libretroshare/src/rsitems/rsdiscovery2items.h
G10h4ck: Gossip discovery is the name that the whole service should take at some point, just `Disc` is too generic and create confusion with other possible method of peer discover, such as DHT, broadcast etc... Gossip apply well here because RsDisc works just like gossip (=people that say things about common friend each other) https://en.wikipedia.org/wiki/Gossip_protocol

G10h4ck: But yes that item just pass a retroshare certificate

libretroshare/src/rsserver/p3peers.cc
G10h4ck: No need to parse HTML, we have RsUrl class that does it safely. URI, and HTML are two completely different formats. rs:// and not retroshare:// is for two motivations, it make the invite shorter, and most libraries that are used in lot of places to detect URL and make them clickable on applications around, recognize things like VERY_SHORT_STRING://something quite commonly, while they usually don't for things like LONGER_STRING://something. This way if someone send an RS invite trough whatsapp or telegram or mail it is much more likely that it get clickable and the used do not need to copy paste, which on mobile phones range from a pain in the ass to impossible. The point to handle it in the library is that all applications that uses libretroshare will have the same format, while if we let them handle each of them by them self is duplicated code and incompatibility (like it happened with Qml app as the URL library there behave different then QUrl I had to sweat a lot to make them interoperate). Anyway we should plan to register both retroshare:// and rs:// as URI schemes to IANA

libretroshare/src/rsserver/p3peers.cc
G10h4ck: NACK

libretroshare/src/services/p3discovery2.cc
G10h4ck: It just make the line longer and add a negation to the condition so it take a little bit more time to understanding what is happening, but I promise I will not change it in all the code :angel: just this time that I had to read it a lot because I was making big changes.

libretroshare/src/services/p3discovery2.cc
G10h4ck: No it is not deleted in all places... around retroshare code we have switches where some cases the ownership is transferred to the function (the delete is not there) and in some cases not, and one never know if it is a leak or not unless one doesn't follow the whole code...
Instead in this case it is explicit in all the places declaration, implementation and call, that the function take the pointer ownership because it take an unique_ptr and even the compiler enforce that (you cannot pass a raw ptr).
This code is safer and more readable the the old one, I think you just need to get a bit more familiar with C++11 pointer types, they are much safer and don't have performance impact in most of the cases especially unique_ptr have no performance impact at all.

libretroshare/src/services/p3discovery2.cc
G10h4ck: Once a peer is authenticated the event is emitted, and the code there actually listen for that event, what happens is that even when rsPeers->isOnline(...) returns true the send keep failing for a bit more, and there is no event that tells when it become possible to send ATM and I could not find (damn unreadable code) at exactly what point it become possible to send an item to a peer without failing 100% of the times. Those sleeps are ugly to me too but I have added them empirically because I found no way to determine when it is possible to send an item to a peer (which apparently is a bit after RsPeer start reporting it as online -_- ). Also only one thread is created, so no problem with pid getting exhausted ;)

libretroshare/src/chat/distantchat.h
G10h4ck: Most probably I intended to move it else were, but then forgot to paste it in the place were it seemed more relevant.

libretroshare/src/pgp/rscertificate.cc
G10h4ck: The compiler wont let you do that because it is conversion between unrelated pointers type ;) also because of this C++ cast is safer the C cast, because you have to be explicit when you are doing something potentially risky ;)

libretroshare/src/pgp/rscertificate.cc
G10h4ck: It get deleted in the destructor. In fact this code was also dangerous because despite checking the pointer to not be null ( that is pure distraction as it is safe to delete a null pointer), the pointer was not set to null after the delete, causing a double delete when the explicit delete ~RsCertificate() is called, with unpredictable effects... Again this catchy constructors are very dangerous in many subtle ways and we should not rely on them at all...

libretroshare/src/pqi/p3linkmgr.cc
G10h4ck: Ok I'll add it back.

libretroshare/src/pqi/p3linkmgr.cc
G10h4ck: This only ignored them for attempt, but do not delete them, which generate very counter-intuitive situations for the user (when I read that line I said DAMN!! it was because of that the some of the address in the listwere never attempted!!!). The deletion on old addresses is handled elsewhere, BTW it is a bad idea to ignore old addresses that we are storing anyway. For example this cause to people that did not open RS for lot of time to loose the possibility to connect to anyone even nodes with static IP... So I am really happy to have found that line and deleted it.

libretroshare/src/pqi/p3linkmgr.cc
G10h4ck: I'll substitute this with RsRandom, but I wonder if when we do attempt NAT traversal if the delay is added too... in that case the NAT traversal success probability go down a lot, and could explain why RS NAT traversal is so bad compared with other applications...

G10h4ck: Altought MAX_RANDOM_ATTEMPT_OFFSET it's just 6 and lastattempt is integer type so don't know if using RsRandom here really make a difference...

libretroshare/src/pqi/p3peermgr.cc
G10h4ck: The PGP fingerprint is not mandatory in short invites as it doesn't really add security there because it is transmitted through the same channel as the SSL id, and make short invite too long depending on the medium (think of a whatsapp mobile chat...)

libretroshare/src/pqi/p3peermgr.cc
G10h4ck: The PGP fingerprint is not mandatory in short invites as it doesn't really add security there because it is transmitted through the same channel as the SSL id, and make short invite too long depending on the medium (think of a whatsapp mobile chat...)

libretroshare/src/pqi/pqiperson.h
G10h4ck: Didn't came up with a more elegant idea for that yet, but I would keep it for a later PR.

libretroshare/src/pqi/pqissllistener.cc
G10h4ck: don't know if a deep copy is made but FailedCertificate(...) check the pointer. Anyway this will be deleted soon, but let me do that on another PR

libretroshare/src/pqi/pqissllistener.cc
G10h4ck: How could that be possible?
VerifyX509Callback is called much before AFAIU, so if the peer is not a verified friend or a pending PGP friend we never get there AFAIU.

libretroshare/src/pqi/pqissllistener.cc
G10h4ck: Didn't understand to what you are referring.

libretroshare/src/retroshare/rsgxsifacetypes.h
G10h4ck: Because now it have the proper name in the first place so no need to alias it anymore ;)

libretroshare/src/retroshare/rsids.h
G10h4ck: Well it should not be that common to create an ID from a block of memory, and when done must be done safely anyway. With the wild pointer constructor In the moment you write the code you probably take care to make it safely, but what happens when another person after some months/years read that code, she says WTF, and need to read a lot to understand it is safe or not, also someone less experiences could use the wild pointer  constructor with less care than you intended in the fist time, and because it is so difficult to follow this kind of code it would even pass a review without much fuss. Passing back a pointer and taking the size of buffer as parameter make explicit to everyone even after years the code was written that this function must be used with care, about the buffer size and the possible failure.

libretroshare/src/retroshare/rsids.h
G10h4ck: I agree it is not strictly needed, but AFAIK on modern C++ run-times memmove() have same efficiency as memcpy(). This way the readers don't have to wonder if the call is safe there or not, and can put more attention in more important details.

libretroshare/src/retroshare/rspeers.h
G10h4ck: It make sense here as the only strict requirement is the sslid

libretroshare/src/rsserver/p3peers.cc
G10h4ck: I made id this way because isPgpFriend properly check if a PGP id is a friend or not, while adding a flag, doesn't make sure that all the previously written code would take it in account. I didn't understand the attack case, can you rephrase it?

libretroshare/src/pqi/p3peermgr.cc
G10h4ck: To create a completely independent method was my original idea too, but later I came to understand that this method is used also when the list of friend is loaded from config files, and it just dropped SSL-only friends, as they are not PGP friend. It happens in p3PeerMgrIMPL::loadList(...), I can put the check in it's place again and use the other method for SSL-only friends in p3PeerMgrIMPL::loadList(...).

libretroshare/src/pqi/pqissllistener.cc
G10h4ck: nope the cert is properly deleted at line 833 don't know why github diff hide just that line ;)

libretroshare/src/retroshare/rsids.h
G10h4ck: I have been thinking about this, in this case as t_RsGenericIdType is trivially copy-able a method that return by value could be offered instead of the one that return the pointer, the method signature would be `static t_RsGenericIdType fromBufferUnsafe(const uint8_t* buff);` Including unsafe in the name suggests to the user and reader that must be used with special care and that documentation should be read before, instead of an apparently innocuous constructor that hide a big danger inside. Also the compiler would optimize away the eventual extra copying on assignation so the performance would be the same as the dangerous costructor. If this alternative is appealing to you I'll change it tomorrow.

libretroshare/src/services/p3discovery2.cc
G10h4ck: In the same file you can see the situation I am talking about, some functions takes ownership and other not, at lines 277, 278, 286, 290, 292 the functions calls seems to take ownership of the pointer, at lines 282, 290 the functions calls seems to not take ownership, and to check if that is correct or not one have to dig deep into the code... while ´recvInvite´ make it explicit that it's taking ownership, and the compiler enforce that too.

libretroshare/src/pqi/p3peermgr.cc
G10h4ck: BTW the comment there was misleading, I am not putting that in place again ;)

G10h4ck: I have put the check in place again with e5519fcc459608a52d41b0fd5694bc225a309082  should be fixed now, mark this conversation as resolved when you double check it ;) Thanks for the careful review ;)

libretroshare/src/chat/distantchat.h
G10h4ck: Solved in 519f6a4a95cf65516ee93fee9bc8854d809a27a5

libretroshare/src/rsserver/p3peers.cc
csoler: No it does not. IsPGPFriend only looks for the PGP fingerprint in a list of accepted peers, but you certainly cannot use that function to allow access to a SSL certificate the signature of which youhave not tested yet. Here you are doing a typical mistake: use a trick that makes coding shorter but at the expense of security.

libretroshare/src/rsserver/p3peers.cc
csoler: If you want the invite to be short, you'd rather use the same format than the one we've been using for a while now in libretroshare: field id + size (in 1,2,5 format) + data, and parse it the same way we parse certificates. It has several advantages over this url thing:
- it is more compact (no something=, just an id, so plenty of space gained, not to mention the rs://)
- no url structure that is totally awkward inside libretroshare and needs some careful parsing, so it's much more robust.
- no parsing ambiguity problems due to encoding because it's all binary. What happens if you have a & or a space in the name of the invite? You'll need to encode this using %foo objects, etc. What about utf8 caracters etc? This is going to create additional weaknesses for no additional value.

libretroshare/src/rsserver/p3peers.cc
csoler: This is really not resolved.

libretroshare/src/services/p3discovery2.cc
csoler: Now that you understand it and do not need to re-write it to understand it, please revert that to the safer version. I bet that someone in the future will replace the = by a == without the person responsible for proof readign the PR being able to notice it.

libretroshare/src/services/p3discovery2.cc
csoler: No really, that is terrible. You see why right? The proper way to handle this here is to post the events into a queue and parse this queue in the data_tick() or tick() function of the service.

libretroshare/src/pgp/rscertificate.cc
csoler: There's a basic rule in crypto: always wipe as soon as possible the data that is not useful. That avoids unintentional leak of that data. So I'de much prefer that you set the pointer to NULL after deleting it.

libretroshare/src/pqi/authssl.h
csoler: ok

libretroshare/src/pqi/p3linkmgr.cc
csoler: ok. I think the idea behind it was to avoid an attempt onto an old address untl that address gets re-activated by a discovery message or a direct connection.

libretroshare/src/pqi/p3linkmgr.cc
csoler: True. I didn't know that this value was so small. Also is there a connection between this and synchronized UDP NAT traversal attempts?

libretroshare/src/pqi/p3peermgr.cc
csoler: I dont like this "optional" format fields. Either the protocol needs the FPR, or it doesn't. And when we write and review the security of all this (which is not done yet) we'll have a much simpler and safer task if there are not such options because they have a lot of implications. I would say that both system (with or without FPR) are possible, but havign both of them at once is not nice at all.

libretroshare/src/pqi/p3peermgr.cc
csoler: Then don't allow it and make sure that everything is safe.

libretroshare/src/pqi/pqiperson.h
csoler: hopefully, because normally you prefer not to add tons of boiler plate code ;-) and this one really is.

libretroshare/src/pqi/pqissllistener.cc
csoler: ok.

libretroshare/src/pqi/pqissllistener.cc
csoler: Never mind: it's crypto world here. Never assume that something is right. Verify it. It doesn't cost much (happens at handshake) and robustly ensures that you're not doing a mistake or that you let a potential weakness. Furthermore, this only happens when you get a connection attempt with a cert that is unknown but signed by a known pgp key. That happens very rarely, so you can afford to take the time to double check this.

libretroshare/src/retroshare/rsgxsifacetypes.h
csoler: mmmh. The proper name is RsGxsGroupId, it's not GXSGroupId. The later was some kind of intermediate proxy and should be the one to remove. 99.99% of the code uses RsGxsGroupId.

libretroshare/src/retroshare/rspeers.h
csoler: I dont understand when you would add a friend without a name to display it nor an address or a DNS to contact it. Addign the SSL id alone should actually never happen.

libretroshare/src/retroshare/rspeers.h
csoler: See my remark above about security protocol. Yu need to make a choice and stick to it.

libretroshare/src/rsitems/rsdiscovery2items.h
csoler: There is also one basic rule: when you modify a given file to add a single bit to it, try to respect the syntax, simply because it is more readable. For instance I've been using p3* for services without understanding why they are named this way because of consistency, same for serialization (before we rewrote *all* the serialization). Otherwise you're messing up the code more and more until it gets impossible to debug or even properly understand.

libretroshare/src/rsitems/rsdiscovery2items.h
csoler: RsDiscoveryCertificateItem and RsDiscoveryCertificateRequestItem would make more sense I would say.

libretroshare/src/pqi/authssl.cc
csoler: exactly. I used to think like you Gio 20 years ago. But sehraf here is right.

libretroshare/src/pqi/authssl.cc
```diff
@@ -989,411 +1002,353 @@ X509 *AuthSSLimpl::SignX509ReqWithGPG(X509_REQ *req, long /*days*/)
...
+	            OPENSSL_malloc(static_cast<unsigned int>(inl)) );
+	unsigned char* p = nullptr;
 #else
-	unsigned char *buf_in=NULL;
+	unsigned char* buf_in = nullptr;
 	inl=i2d_re_X509_tbs(x509,&buf_in) ;	// this does the i2d over x509->cert_info
 #endif
 
-	if(buf_in == NULL)
+	if(buf_in == nullptr)
```
csoler, line 1053: it's the same here: replaceing all NULL by nullptr is seriously not useful at all and wastes the time of the reviewer.

csoler (changes requested):
Overall, the cryptographic security of this PR (which is really critical) is impossible to assess, because this PR is polluted by small syntax changes that just come here to satisfy your love of C++11. This really doesn't help at all. I'm still trying to read authssl and pqissl.

libretroshare/src/pqi/authssl.cc
G10h4ck: I know the @sehraf case, and it is a language pitfail...  In this case it semantically change nothing as we are declaring just one variable, and conceptually * is still part of the type, as `signature` is a variable of type `ASN1_BIT_STRING*`.

libretroshare/src/pgp/rscertificate.cc
G10h4ck: I can do that but take in account it is deleted right there anyway, because it goes out of scope right there.

libretroshare/src/pqi/p3linkmgr.cc
G10h4ck: I am not 100% sure due to the fact that this part of the code is really difficult to follow, but I am driven to believe this code is involved in UDP connections too.

libretroshare/src/pqi/p3peermgr.cc
G10h4ck: Ok in this case, as the idea behind this to have very short invites (we already have the long ones in place), that other application might easily recognize as something the user might want to click on (if they are not clickable they are practically unusable in the mobile phones, I can make concrete use case examples in favor of making them something that should be clickable). I'll remove the PGP fingerprint.

libretroshare/src/pqi/pqissllistener.cc
G10h4ck: Ok I'll add the verification again there, but with comment explaining.

libretroshare/src/retroshare/rsgxsifacetypes.h
G10h4ck: Yeah I agree the one I have removed is `GXSGroupId` despite to what looking at this line seems to suggest ;). I have named it `RsGxsGroupId` when it is defined instead of having that alias around, no no one get tempted to use `GXSGroupId` and `follow symbol under cursor` hop directly to the definition instead of having to jump over 3 file to get to the real type :p

libretroshare/src/retroshare/rspeers.h
G10h4ck: Adding it only with SSL id would give only the possibility to accept incoming connection attempts, but I agree is not the most desirable, way of adding someone, actually even when `bareBones` is set to `true` the produced invite include SSL id and PGP name.

libretroshare/src/rsserver/p3peers.cc
G10h4ck: Well it is what I expect from `isPgpFriend(PgpId)` just check in the list of friend PGP ids and return false if it is not a friend. We cannot verify the PGP signature there because we don't have the PGP key as expected. Perhaps `isFriendPendingPgp(RsPeerId)` isn't a good name... What this method is supposed to do is to check if an RsPeerId is an SSL-only friend, so it return false for PGP friends, and for non-friends, while it return true only for SSL-only friends. Is `isSslOnlyFriend(RsPeerId)` more understandable?

libretroshare/src/services/p3discovery2.cc
G10h4ck: I agree that the sleep thing is ugly, but putting it in a queue, doesn't help with the problem that we don't really know when the peer become available in terms of `sendItem`. RsPeers::isOnline(...) start returning true a little bit before `sendItem` would success... Where is the place where we know for sure it is possible to send an item to a peer? In that pace I could emit an event, and listen for that, without any sleep

libretroshare/src/pqi/authssl.cc
G10h4ck: It just avoid extra compiler warnings, so hopefully at some point we get to 0 or a number so small we easily notice when adding some code we introduce a new warning.

G10h4ck:
> Overall, the cryptographic security of this PR (which is really critical) is impossible to assess, because this PR is polluted by small syntax changes that just come here to satisfy your love of C++11. This really doesn't help at all. I'm still trying to read authssl and pqissl.

To make it easier I could split the PR in smaller PR, as an example I can isolate the rsids refactor in a single PR, the debugging helpers in another PR, the RsDisc refactor in another PR and so on, and the short invites very specific one would be the last one. But before doing that huge splitting work I'd prefer to be sure we agree on a short invites *prototype* implementation. So for this PR if you can focus on the security aspects, and do not pay much attention to the other things it would be acceptable for me to not merge this as is but to start working on splitting it in smaller PRs once we come to a coded agreement ;).

libretroshare/src/services/p3discovery2.cc
csoler: In tick() (or data_tick()?) you can query rsPeers->isPeerOnline() and when it says it is online, you can sent items without problem. We've been doing this for a while. If it was not working we would have noticed already.
Also, most of the time, when handling this kind of async queued events, we call the function that does this only once in a while (e.g. every 5, 10 or even 60 secs depending on the situation) and it's far likely that the peer is connected since several secs

libretroshare/src/rsserver/p3peers.cc
csoler: Moreover, I dont see a single usage case where you would like to create the invite yourself: the invite is created by libretroshare and parsed by libretroshare, which means it doesn't need to have a "readable" url format (with all the risks that come with it). For instance in a webUI you'd call JSonAPI->sendMeAnInvite, you get the base64 string and paste it as needed. Same when adding a friend. If needed,you may implement a API function to parse it for you, such as returning the name, SSL Id, etc for display. That's what we do with certificates.

libretroshare/src/pqi/authssl.cc
csoler: Well, that you use this funny writing in your own code, fine. Although I would not do so myself, I dont really mind. But dont spend time re-writing it in existing code!

G10h4ck: yeah don't worry I don't plan to do that ;)

libretroshare/src/rsserver/p3peers.cc
G10h4ck: The point to make it look like a normal URL is not programmer comfort to generate it outside libretroshare (on the contrary I want to avoid that!! Right now retroshare URLs are generated in retroshare-gui that created compatibility problems between Qml app and retroshare-gui!!!!), it is for user comfort. To make other applications, like a web browsers, whatsapp, telegram, whatever that may be used to give/get the invite to/from other that you cannot yer reach over RS, recognize the invites as something that might be clicked, because copy paste an invite in mobile phones is really bad and even impossible in some cases, such as when you use whatsapp (that is the app 99% of user uses today). Frankly I don't think that prepending `rs://sslid/` in front of whatever format we agree to use (that could be the plain old radix) make it insecure... I think you should take a look at RsUrl code... it is simple (because it doesn't intend to support every URL in the word but only simple URLs as the one we uses) and safe.

libretroshare/src/rsserver/p3peers.cc
G10h4ck: yeah but let's keep this discussion only in one place

libretroshare/src/services/p3discovery2.cc
G10h4ck: > you can query rsPeers->isPeerOnline() and when it says it is online, you can sent items without problem

No it is not like that, here I had to add the `sleep_for`
```
if(rsPeers->isOnline(sslId))
					{
						// Need to wait a bit more even if online
						std::this_thread::sleep_for(std::chrono::seconds(2));
						if(rsDisc && rsDisc->requestInvite(sslId, sslId))
``` 
becase the `sendItem(item)` inside `requestInvite` was failing (i see that becaus eit print an helpful message when that happens)

So doing this inside `tick()` maybe seems more elegant, but would work by chance anyway, most probably because tick is called once in a while... still we have the problem that `rsPeers->isOnline(sslId)` returns true a bit before we are able to send items

csoler: the same test isPeerOnline() is used *everywhere* and we have never watched items not being sent because the peer is actually not online. What may have happenned in your case is that the item is of too low priority and therefore got stuck for a while before is was actually sent.
If you want to make sure to wait for some time after the peer is online, then add a field to your queue item with the last time you checked that it is online. When this time is long enough, send your item. But I dont even think that is necessary at all.
For the record the async[] trick is nice and works best in GUI situations especially because of Qt thread issues, but here it's really not what you need.

libretroshare/src/rsserver/p3peers.cc
csoler: What's insecure is to use the url format to store multiple fields some of which are strings. This is the open door to parsing problems and attack vectors. Remember what happenned with xml ;-)

Now what you're describing is exactly my point: to allow all applications to understand the invite in the best level of compatibility, you need to set it to be a base64 string that is unbreakable, and let libretroshare parse it. Having urls in libretroshare itself (especially if it has a single field that is the base64 string itself) doesnt make sense to me.
Then if you want to display this in a way or another in your own app, fine. At least you wont have to parse it in order to extract the actual information you need to re-form your own url format.

csoler: For the record, retroshare certificate in plain PGP key format and named fields have always been a pain to parse and caused many problems and errors in different applications. Now that we have a certificate format that is a single base64 string, all these problems are gone. It's more robust, more secure and easier to handle. Nothing prevents you in your own app to display the cert a rs://certificate/[base64 string]

G10h4ck: The problem is not in an App that we control of course there we can do whatever we want and not use URL at all eventually. The problem is in the apps we don't control such as whatsapp, that is what most users could use in mobile phones to exchange the invites, so we can avoid dangerous fields in the URL, but it is important the string is passed around looks like an URL and can be parsed as a simple URL, so those apps "make it blue" and clickable, that automatically trigger an OS handler that look for the application able to handle that URI scheme and pass it to the correct app for handling.

csoler: Of course. But then how do you get the url to arrive in a whatsapp application? It needs to come from soem UI of some sort. It cannot come directly from libretroshare. So that's where you add the rs://thing

G10h4ck: > For the record, retroshare certificate in plain PGP key format and named fields have always been a pain to parse and caused many problems and errors in different applications. Now that we have a certificate format that is a single base64 string, all these problems are gone. It's more robust, more secure and easier to handle. Nothing prevents you in your own app to display the cert a rs://certificate/[base64 string]

The point in doing it in libretroshare, is that hen all apps developed on top of libretroshare use the same compatible format, so a link generated by customAndroidRsApp is exactly the same by one generated by retroshare-gui avoiding fragmentation

csoler: sure. And that's pretty easy to do. You dont need to enforce it, given the cost of it. A simple comment around the method that give you the invite to suggest how to convert it into a url would do the job.

G10h4ck: > Of course. But then how do you get the url to arrive in a whatsapp application? It needs to come from soem UI of some sort. It cannot come directly from libretroshare. So that's where you add the rs://thing

That means that then one should implement it in the same way in retroshare-gui, elrepo.io, retroshare-qml-app, emoty... that all uses different libraries and are maintained by different persons... and it was already a pain to make it compatible in the past that was only between retroshare-gui and retroshare-qml-app...

G10h4ck: > sure. And that's pretty easy to do. You dont need to enforce it, given the cost of it. A simple comment around the method that give you the invite to suggest how to convert it into a url would do the job.

It is not enforced at all, the `formatRadix` option is there just for that... I don't see which is the cost to do that in the library... XML and RsUrl two completely different beasts, a simple URL won't introduce any vulnerability

G10h4ck:
Obsoleted by https://github.com/RetroShare/RetroShare/pull/1565

PR #1535, opened 2019-04-22
Added color feature for channels & posted

defnax:
- Added color feature for Channels & posted header frames
- Added header frame for posted links too

by the way the colors can be used on forums too but i doesnt enabled yet there.
please test it and feedbacks are welcome

image

image


defnax:
please test this pr
22 days over and nobody checking it?
@csoler @hunbernd @G10h4ck @felisucoibi @PhenomRetroShare @RetroPooh

felisucoibi:
i tested and is not working, i can st a color but in my rs nothing happens. the channles keeps as usual, i'm using dark style.

defnax:
> i tested and is not working, i can st a color but in my rs nothing happens. the channles keeps as usual, i'm using dark style.

please test without darkstyle too

felisucoibi:
it works well even in darkstyle, thanks.

csoler:
@felisucoibi : did you test backward compatibility? I mean:
- does old RS still loads the channel created on the new RS (without color of course)
- is switching back and forth between old/new RS in the same location causing any disruption (beyond losing the color information)?

felisucoibi:
I did not tested backwards, just the latest build it works, with different skins. Like i edited some channels, i wil expect some feedback if a problem is found.

csoler:
Generally, I am quite doubtful about the advantage of having colors in widgets that propagate to friends. Depending on your own skin, displaying the channels/forums/etc will sometimes look terrible, and everyone in the network will have to cope with a look and feel that depends on the own choice of an anonymous person without your consent. What will your UI look like when everyone decide to use a different color for his own channels? Because of this, it seems to me that it is a bad idea to change the color or widgets at friends. 

Besides, the way you define the color clearly shows that you want to do this locally only:  you store the color as a string that is extracted from the button in the GUI. That means the color text is only valid for a person with the same language, otherwise the button will just look different. So there's no chance this can make its way to friends properly.

Even worse is the way you are using to re-inject the color in the html code as a string (see PostedListWidget.cpp:333). That clearly is a security hole. Anyone could create a forum/channel with some nasty code (e.g. lol bombs, images, etc) in the color string and use that to corrupt everyone's GUI in the network.

That being said, some people (including you) might want to display the channels/forums/posted on their own node using a particular  color. We can see this as e.g. a personal tool to easily locate some particular channel in the list. This is what we do for mails for instance, and the color of mails is obviously not "sent" with the mail.

So I strongly suggest that you do not send the color information and only keep it as a local property of the channel/forum/posted UI (nothing to be put inside the GroupData for instance!). It should only be stored in the Qt preferences file.

Also, remove .gitignore from the PR. It's probably added by mistake.

defnax:
> That being said, some people (including you) might want to display the channels/forums/posted on their own node using a particular color. We can see this as e.g. a personal tool to easily locate some particular channel in the list. This is what we do for mails for instance, and the color of mails is obviously not "sent" with the mail.

my idea is not for locally only, for locally dont need this feature, i mean not spent time my own retroshare. But you mean its a security hole with this, then when thats right it makes no sences with colors.
i thinked instead of background picture which takes lots of size, thinked use a color string.

idea i get from Youtube/Reddit/twitter, you can show your subscribers your own design.
Each Channel or Posted can have a own color, when Channel admin want it only optional.
Experts, anon people, geeks dont need to use colors or pics.

With the color you can attractive/promote your channel when needed with your color branding.
Channels/Posted will be look more interesting, or without boring.
example: You has a nice Red Logo and want a nice color for your channel which looks good to your logo.

@felisucoibi what you think?
i wanted only get rs more atractive to have more nice have features, i dont want a boring rs.

defnax:
> Even worse is the way you are using to re-inject the color in the html code as a string (see PostedListWidget.cpp:333). That clearly is a security hole. Anyone could create a forum/channel with some nasty code (e.g. lol bombs, images, etc) in the color string and use that to corrupt everyone's GUI in the network.

maybe you know a better way how use in a secure way to send the hex colors?
i know nobody doing something things like my ideas, color features example.

G10h4ck:
From a short review of the code it seems to me this can break retro-compatibility because it changes the item format (it add a string at the end of the item). About sending the color in a safe way you can use an uint32_t but again I think it is not a good idea to break retro-compatibility.

G10h4ck:
Also you should not remove `.gitignore` either, live it as it was because you can use `git checkout master -- .gitignore` to restore it as it was

G10h4ck (changes requested):
See previous comments

defnax:
> uint32_t



> From a short review of the code it seems to me this can break retro-compatibility because it changes the item format (it add a string at the end of the item). About sending the color in a safe way you can use an uint32_t but again I think it is not a good idea to break retro-compatibility.

`const uint16_t TLV_TYPE_STR_COLOR     = 0x00bb;
`

i thinked this helped to be backward compatible then im wrong:
	
```
if(j == RsGenericSerializer::DESERIALIZE && ctx.mOffset == ctx.mSize)
return ;

if((j == RsGenericSerializer::SIZE_ESTIMATE || j == RsGenericSerializer::SERIALIZE) && mColor.empty())
return ;
```

maybe @csoler or you can fix it the compatiblity and make it safe?
but from the beginning i wanted backward compatible.

defnax:
> Even worse is the way you are using to re-inject the color in the html code as a string (see PostedListWidget.cpp:333). That clearly is a security hole. Anyone could create a forum/channel with some nasty code (e.g. lol bombs, images, etc) in the color string and use that to corrupt everyone's GUI in the network.

i dont know maybe you can make it safe?
maybe use without stylesheet?
i doesnt get worked with QPalette
https://wiki.qt.io/How_to_Change_the_Background_Color_of_QWidget
```
	QString colorstring = QString::fromUtf8(group.mColor.c_str());
	//ui->headFrame->setStyleSheet(QString("QFrame#headFrame{background-color: %1 };").arg(colorstring));
	
	QPalette pal = palette();

	// set background color
	pal.setColor(QPalette::Background, QColor(colorstring));
	ui->headFrame->setAutoFillBackground(true);
	ui->headFrame->setPalette(pal);
```

defnax:
any help on this pr would be nice

PR #1542, opened 2019-04-24
Darkstylesheet fixes

defnax:
Fixed some wrong url path.
Fixed the background of the labels to be transparent


felisucoibi:
+10000

defnax:
> +10000

normally this things can be fixed by you :)

felisucoibi:
i tried to modify dark themes to change blue link color (a href) to another color and i was not able, i tried hard, maybe you can do it.
default blue with balck backgruond is impossible to read.

Also there are some missing or invisible icons in tree filelists and other tree componetns of rs.

defnax:
> i tried to modify dark themes to change blue link color (a href) to another color and i was not able, i tried hard, maybe you can do it.
> default blue with balck backgruond is impossible to read.

that links color is hardcoded but it can be changed to use light blue
> 
> Also there are some missing or invisible icons in tree filelists and other tree componetns of rs.

i has fixed the tree icons which was missed, i doesnt looked on filelist.

felisucoibi:
i really triend to remove this blue, is in lot of places, but i read wasnot possible globally, i really tried and i'm not a programmer of this century at least. :page_facing_up: 
I think is not fixed in fiellists.

defnax:
> I think is not fixed in fiellists.
 
why you not make screenshots? then we see it better

here with the changes:
image

felisucoibi:
I'm using qdarkstye

imagen

felisucoibi:
The blue in news feed for example impossible to read.
imagen
And the blue in links in forums etc.
imagen

defnax:
> I'm using qdarkstye
> 
> imagen

thats is fixed on master

defnax:
> The blue in news feed for example impossible to read.
> imagen
> And the blue in links in forums etc.
> imagen

maybe need to check if using a dark styleesheet or else use a different href color light blue or else

PR #1551, opened 2019-05-01
Added country emojis

defnax:
added country emojis

image


csoler:
how many megabytes is that??

PR #1554, opened 2019-05-02
Gossip discovery refinement

G10h4ck:
Rename from RsDisc to more descriptive RsGossipDiscovery
Keep full retro-compatibility suggesting usage of RsGossipDiscovery
Add capability to send and receive full RetroShare invitation
Emit event when receiving a full invitation
Start using new debug utilities
Use enum class instead of defines and constant where appropriate

### DIFF
--- a/libretroshare/src/gossipdiscovery/gossipdiscoveryitems.cc
+++ b/libretroshare/src/gossipdiscovery/gossipdiscoveryitems.cc
@@ -1,9 +1,10 @@
 /*******************************************************************************
- * libretroshare/src/rsitems: rsdiscitems.cc                                   *
+ * Gossip discovery service items                                              *
  *                                                                             *
  * libretroshare: retroshare core library                                      *
  *                                                                             *
- * Copyright 2007-2008 by Robert Fernie <retroshare@lunamutt.com>              *
+ * Copyright (C) 2007-2008  Robert Fernie <retroshare@lunamutt.com>            *
+ * Copyright (C) 2019  Gioacchino Mazzurco <gio@eigenlab.org>                  *
  *                                                                             *
  * This program is free software: you can redistribute it and/or modify        *
  * it under the terms of the GNU Lesser General Public License as              *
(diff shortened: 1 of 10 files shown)


libretroshare/src/gossipdiscovery/gossipdiscoveryitems.cc
```diff
@@ -19,60 +20,49 @@
...
-	case RS_PKT_SUBTYPE_DISC_SERVICES           : return new RsDiscServicesItem();      //= 0x04;
-#endif
-	case RS_PKT_SUBTYPE_DISC_CONTACT            : return new RsDiscContactItem();       //= 0x05;
-	case RS_PKT_SUBTYPE_DISC_IDENTITY_LIST      : return new RsDiscIdentityListItem();  //= 0x06;
-    default:
-    return NULL ;
-    }
+	if(service != RS_SERVICE_TYPE_DISC) return nullptr;
+
+	switch(static_cast<RsGossipDiscoveryItemType>(item_subtype))
```
csoler, line 36: I have to say this is a bit ugly. Using enums is generally good, but RsItem have their own sub-item IDs which are only used inside the item class, so it's not really a problem that they are actually uint8_t, and it makes the code *much* more readable.

libretroshare/src/gossipdiscovery/gossipdiscoveryitems.cc
```diff
@@ -19,60 +20,49 @@
...
 }
 
-void RsDiscPgpListItem::serial_process(RsGenericSerializer::SerializeJob j,RsGenericSerializer::SerializeContext& ctx)
+void RsDiscPgpListItem::serial_process(
+        RsGenericSerializer::SerializeJob j,
+        RsGenericSerializer::SerializeContext& ctx )
 {
-    RsTypeSerializer::serial_process<uint32_t>(j,ctx,mode,"mode") ;
-    RsTypeSerializer::serial_process<RsTlvItem>(j,ctx,pgpIdSet,"pgpIdSet") ;
+	RS_SERIAL_PROCESS(mode);
```
csoler, line 64: Isn't that dangerous?? It seemed to me that for int types, specifying which type it is forces the serializer to use the right method. This is very important if e.g. in an other system the ints have a different size.
In this particular case: DIST_PGP_LIST_MODE_NONE is a uint8_t so it was on purpose serialized as a uint32_t for compatibility on all machines.  Now it is an enum but you have no guarranty which type the enum will be on the system it's compiling. So it's better to force the serialization with uint32_t.

libretroshare/src/gossipdiscovery/gossipdiscoveryitems.h
```diff
@@ -31,79 +31,77 @@
...
 
-const uint8_t RS_PKT_SUBTYPE_DISC_PGP_LIST           = 0x01;
-const uint8_t RS_PKT_SUBTYPE_DISC_PGP_CERT           = 0x02;
-const uint8_t RS_PKT_SUBTYPE_DISC_CONTACT_deprecated = 0x03;
-const uint8_t RS_PKT_SUBTYPE_DISC_SERVICES           = 0x04;
-const uint8_t RS_PKT_SUBTYPE_DISC_CONTACT            = 0x05;
-const uint8_t RS_PKT_SUBTYPE_DISC_IDENTITY_LIST      = 0x06;
+enum class RsGossipDiscoveryItemType : uint8_t
+{
+	PGP_LIST           = 1,
```
csoler: We use a hexa notation everywhere. It makes the code cleaner ;-)

libretroshare/src/gossipdiscovery/gossipdiscoveryitems.h
```diff
@@ -31,79 +31,77 @@
...
+	virtual ~RsDiscItem();
+};
 
-#define DISC_PGP_LIST_MODE_NONE		0x00
-#define DISC_PGP_LIST_MODE_FRIENDS	0x01
-#define DISC_PGP_LIST_MODE_GETCERT	0x02
+/// uint32_t overkill just for retro-compatibility
+enum class RsGossipDiscoveryPgpListMode : uint32_t
+{
+	NONE    = 0,
```
csoler: Same here.

libretroshare/src/retroshare/rsdisc.h
```diff
@@ -19,66 +19,14 @@
...
-	 * @jsonapi{development}
-	 * @param[out] sendCount number of queued outgoing packets
-	 * @param[out] recvCount number of queued incoming packets
-	 * @return true on success false otherwise
-	 */
-	virtual bool 	getWaitingDiscCount(size_t &sendCount, size_t &recvCount) = 0;
-};
-
-#endif
+#define rsDisc rsGossipDiscovery.get()
```
csoler, line 32: You read a book about design patterns recently ?? ;-)

libretroshare/src/gossipdiscovery/gossipdiscoveryitems.cc
G10h4ck: To keep retro-compatibility I have enforced that enumeration underlining type to be `uint32_t` as you can see in the header `enum class RsGossipDiscoveryPgpListMode : uint32_t` and the serializer onor that. About `uint8_t` is safe to use an all architecture ;) so we don't need to use bigger types when unneeded except for retro-compatibility as in this case ;)

G10h4ck: Also the documentation of `rstypeserializer` says that in order to serialize enumeration types in a cross-platform way the underlining type must be specified at declaration time (like in this case ;) )

libretroshare/src/gossipdiscovery/gossipdiscoveryitems.cc
G10h4ck: I don't agree about the much more readable, in the end it is only one cast, with the benefit that one can follow the symbol under cursor and get to the place where the type is declared being sure all the constants are there, this is valid also for method signatures, using uin8_t from the method declaration one has no way to know what values might be passed to that, while having the enum type from the method declaration one can know all the possible values (this is good also for _in future_ testing the code)

G10h4ck: Also using the enumeration give the compiler the possibility to be stricter, the other positive thing is that using the enumeration the namespace pollution is much less

libretroshare/src/gossipdiscovery/gossipdiscoveryitems.h
G10h4ck: Ok I can change that to my eyes it is equivalent ;)
I used decimal notation because soon we will have tons of JavaScript kiddies reading our code as they fiddle with JSON API and lot of them doesn'T even know that when a number start with 0x it is to be read as hexadecimal, sad to say but it is true :(

libretroshare/src/retroshare/rsdisc.h
G10h4ck: LOL no :p

G10h4ck: but offer a more consistent API, with more uniform patterns, will make life easier to new developers ;)

sehraf: > life easier to new developers

I thought that's exactly what we try to avoid :wink:

libretroshare/src/gossipdiscovery/gossipdiscoveryitems.h
gmarcos87: Very sad :D

PR #1559, opened 2019-05-02
Cleanup huge amount of AuthSSL and related files cruft

G10h4ck:
Make reduntant auth check in pqi effective (before even if redundant only siganture was checked but friendess wasn't)
Evidence redundant auth check in pqi by putting it inside #ifdef this way the beavior being the same with and without redundant check can be tested easier
Solve lot of compiler warnings and made code more readable
Remove dangerous sslcert wrapper
Remove misleading messages and notification about peer not giving cert, FailedCertificate logic is wrong since many years as authentication is fully handled inside VerifyX509Callback

### DIFF
--- a/libretroshare/src/pqi/authssl.cc
+++ b/libretroshare/src/pqi/authssl.cc
@@ -974,9 +974,9 @@ bool AuthSSLimpl::AuthX509WithGPG(X509 *x509, uint32_t& diagnostic)
 	RsPeerDetails pd;
 	if (!AuthGPG::getAuthGPG()->getGPGDetails(issuer, pd))
 	{
-		RsErr() << __PRETTY_FUNCTION__ << " X509 NOT authenticated : "
-		        << "AuthGPG::getAuthGPG()->getGPGDetails(" << issuer
-		        << ",...) returned false." << std::endl;
+		RsInfo() << __PRETTY_FUNCTION__ << " X509 NOT authenticated : "
+		         << "AuthGPG::getAuthGPG()->getGPGDetails(" << issuer
+		         << ",...) returned false." << std::endl;
 		diagnostic = RS_SSL_HANDSHAKE_DIAGNOSTIC_ISSUER_UNKNOWN;
 		return false;
 	}
@@ -1125,11 +1125,11 @@ bool AuthSSLimpl::AuthX509WithGPG(X509 *x509, uint32_t& diagnostic)
 			goto err;
 		}
 
-		Dbg1() << __PRETTY_FUNCTION__ << " Verified: " << sigtypestring
-		       << " signature of certificate sslId: "
-		       << RsX509Cert::getCertSslId(*x509)
-		       << ", Version " << std::hex << certificate_version << std::dec
-		       << " using PGP key " << pd.fpr << " " << pd.name << std::endl;
+		RsInfo() << __PRETTY_FUNCTION__ << " Verified: " << sigtypestring
+		         << " signature of certificate sslId: "
+		         << RsX509Cert::getCertSslId(*x509)
+		         << ", Version " << std::hex << certificate_version << std::dec
+		         << " using PGP key " << pd.fpr << " " << pd.name << std::endl;
 	}
 
 	EVP_MD_CTX_destroy(ctx);
@@ -1140,9 +1140,9 @@ bool AuthSSLimpl::AuthX509WithGPG(X509 *x509, uint32_t& diagnostic)
 
 	return true;
 
-err:
-	RsErr() << __PRETTY_FUNCTION__ << " X509 PGP authentication failed with "
-	        << "diagnostic: " << diagnostic << std::endl;
+err: // TODO: this label is very short and might collide every easly
+	RsInfo() << __PRETTY_FUNCTION__ << " X509 PGP authentication failed with "
+	         << "diagnostic: " << diagnostic << std::endl;
 
 	if(buf_in) OPENSSL_free(buf_in);
 
@@ -1232,7 +1232,7 @@ int AuthSSLimpl::VerifyX509Callback(int /*preverify_ok*/, X509_STORE_CTX* ctx)
 	}
 
 	uint32_t auth_diagnostic;
-	if (!AuthX509WithGPG(x509Cert, auth_diagnostic))
+	if(!AuthX509WithGPG(x509Cert, auth_diagnostic))
 	{
 		std::string errMsg = "Certificate was rejected because PGP "
 		                     "signature verification failed with diagnostic: "
--- a/libretroshare/src/pqi/authssl.cc
+++ b/libretroshare/src/pqi/authssl.cc
@@ -3,7 +3,8 @@
  *                                                                             *
  * libretroshare: retroshare core library                                      *
  *                                                                             *
- * Copyright 2004-2008 by Robert Fernie, Retroshare Team.                      *
+ * Copyright (C) 2004-2008  Robert Fernie <retroshare@lunamutt.com>            *
+ * Copyright (C) 2019  Gioacchino Mazzurco <gio@eigenlab.org>                  *
  *                                                                             *
  * This program is free software: you can redistribute it and/or modify        *
  * it under the terms of the GNU Lesser General Public License as              *
@@ -62,9 +63,7 @@ const uint32_t RS_SSL_HANDSHAKE_DIAGNOSTIC_WRONG_SIGNATURE_VERSION     = 0x0a ;
  * #define AUTHSSL_DEBUG 1
  ***/
 
-// initialisation du pointeur de singleton
-AuthSSL *AuthSSL::instance_ssl = NULL;
-static pthread_mutex_t *mutex_buf = NULL;
+static pthread_mutex_t* mutex_buf = nullptr;
 
 struct CRYPTO_dynlock_value
 {
@@ -214,31 +213,17 @@ void tls_cleanup()
 	}
 }
 
-/* hidden function - for testing purposes() */
-void AuthSSL::setAuthSSL_debug(AuthSSL *newssl)
+/*static*/ AuthSSL& AuthSSL::instance()
 {
-	instance_ssl = newssl;
+	static AuthSSLimpl mInstance;
+	return mInstance;
 }
 
-void AuthSSL::AuthSSLInit()
-{
-	if (instance_ssl == NULL)
-	{
-		instance_ssl = new AuthSSLimpl();
-	}
-}
-  
-AuthSSL *AuthSSL::getAuthSSL()
-{
-	return instance_ssl;
-}
+AuthSSL* AuthSSL::getAuthSSL() { return &instance(); }
 
-AuthSSL::AuthSSL()
-{
-	return;
-}
+AuthSSL::~AuthSSL() = default;
 
-  
+
 /********************************************************************************/
 /********************************************************************************/
 /*********************   Cert Search / Add / Remove    **************************/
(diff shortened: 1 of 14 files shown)


libretroshare/src/pqi/authssl.cc
```diff
@@ -984,416 +968,330 @@ X509 *AuthSSLimpl::SignX509ReqWithGPG(X509_REQ *req, long /*days*/)
 }
 
 
-/* This function, checks that the X509 is signed by a known GPG key,
```
csoler, line 987: Why remove that comment??

libretroshare/src/pqi/authssl.cc
```diff
@@ -984,416 +968,330 @@ X509 *AuthSSLimpl::SignX509ReqWithGPG(X509_REQ *req, long /*days*/)
...
 
 	/* verify GPG signature */
-
 	/*** NOW The Manual signing bit (HACKED FROM asn1/a_sign.c) ***/
 
 #if OPENSSL_VERSION_NUMBER < 0x10100000L || defined(LIBRESSL_VERSION_NUMBER)
 	int (*i2d)(X509_CINF*, unsigned char**) = i2d_X509_CINF;
-	ASN1_BIT_STRING *signature = x509->signature;
-	X509_CINF *data = x509->cert_info;
+	ASN1_BIT_STRING* signature = x509->signature;
```
csoler, line 992: come on!!

libretroshare/src/pqi/authssl.cc
```diff
@@ -984,416 +968,330 @@ X509 *AuthSSLimpl::SignX509ReqWithGPG(X509_REQ *req, long /*days*/)
...
-	unsigned char *buf_in=NULL;
+	unsigned char* buf_in = nullptr;
 	inl=i2d_re_X509_tbs(x509,&buf_in) ;	// this does the i2d over x509->cert_info
 #endif
 
-	if(buf_in == NULL)
+	if(buf_in == nullptr)
 	{
-		fprintf(stderr, "AuthSSLimpl::AuthX509: ASN1err(ASN1_F_ASN1_SIGN,ERR_R_MALLOC_FAILURE)\n");
+		RsErr() << __PRETTY_FUNCTION__
```
csoler, line 1021: haven't you put the __PRETTY_FUNCTION__ inside the macro already?

libretroshare/src/pqi/authssl.cc
```diff
@@ -984,416 +968,330 @@ X509 *AuthSSLimpl::SignX509ReqWithGPG(X509_REQ *req, long /*days*/)
...
-		RsPeerId peerIdstr ;
-		getX509id(x509, peerIdstr) ;
-
-		std::string fpr = pd.fpr.toStdString();
-		std::cerr << "Verified " << sigtypestring << " signature of certificate " << peerIdstr << ", Version " << std::hex << certificate_version
-		          << std::dec << " using PGP key " ;
-		for(uint32_t i=0;i<fpr.length();i+=4)
-			std::cerr << fpr.substr(i,4) << " " ;
-		std::cerr << std::endl;
+		Dbg1() << __PRETTY_FUNCTION__ << " Verified: " << sigtypestring
```
csoler: This is not really a "debug" message. I would classify this into "information". It shouldn't be omitted e.g. when the software is released, whereas debug message should IMHO.

libretroshare/src/pqi/authssl.cc
```diff
@@ -984,416 +968,330 @@ X509 *AuthSSLimpl::SignX509ReqWithGPG(X509_REQ *req, long /*days*/)
...
+	        << "diagnostic: " << diagnostic << std::endl;
 
-	if(buf_in != NULL)
-		OPENSSL_free(buf_in) ;
+	if(buf_in) OPENSSL_free(buf_in);
 
 	return false;
 }
 
-
```
csoler, line 1192: This wasn't used?

libretroshare/src/pqi/authssl.cc
```diff
@@ -984,416 +968,330 @@ X509 *AuthSSLimpl::SignX509ReqWithGPG(X509_REQ *req, long /*days*/)
...
+	 * About preverify_ok OpenSSL documentation doesn't tell which value is
+	 * passed to the first callback in the authentication chain, it just says
+	 * that the result of previous step is passed down, so I have tested it
+	 * and we get passed 0 always so in our case as there is no other
+	 * verifications step vefore we ignore it completely */
+
+	constexpr int verificationFailed = 0;
+	constexpr int verificationSuccess = 1;
+
+	using Evt_t = RsAuthSslConnectionAutenticationEvent;
```
csoler, line 1174: This design adds shared pointers where they are actually not needed. A single pointer corresponding to an event would do fine as soon as the documentation clearly states who owns the memory.

libretroshare/src/pqi/authssl.cc
```diff
@@ -984,416 +968,330 @@ X509 *AuthSSLimpl::SignX509ReqWithGPG(X509_REQ *req, long /*days*/)
...
-#ifdef AUTHSSL_DEBUG
-    std::cerr << "AuthSSLimpl::VerifyX509Callback(preverify_ok: " << preverify_ok
-              << " Err: " << err << " Depth: " << depth << std::endl;
-#endif
+		if(rsEvents)
+		{
+			ev->mSslCn = sslCn;
+			ev->mPgpId = pgpId;
+			ev->mErrorMsg = errMsg;
+			rsEvents->postEvent(std::move(ev));
```
csoler, line 1210: same here.

libretroshare/src/pqi/authssl.cc
```diff
@@ -984,416 +968,330 @@ X509 *AuthSSLimpl::SignX509ReqWithGPG(X509_REQ *req, long /*days*/)
...
-#ifdef AUTHSSL_DEBUG
-        fprintf(stderr, "A normal certificate is probably a security breach attempt. We sould fail it !!!\n");
-#endif
-        preverify_ok = false;
-    }
+		return verificationFailed;
+	}
 
-    if (preverify_ok) {
+	AuthSSL::instance().setCurrentConnectionAttemptInfo(pgpId, sslId, sslCn);
```
csoler: So you still use this?

libretroshare/src/pqi/authssl.cc
```diff
@@ -984,416 +968,330 @@ X509 *AuthSSLimpl::SignX509ReqWithGPG(X509_REQ *req, long /*days*/)
...
-
-		getX509id(x509,sslid);
-
-		if(sslid.isNull()) 
+		if(rsEvents)
 		{
-			std::cerr << "verify_x509_callback(): wrong PGP id \"" << sslcn << "\"" << std::endl;
-			return false ;
+			ev->mErrorMsg = errMsg;
+			rsEvents->postEvent(std::move(ev));
```
csoler, line 1189: I really have the feeling that you use this because you like shared pointers, but it seems unnecessary to me in this situation (authssl).

csoler: All this is because your event type derives from RsEvent so you need a pointer to handle subclassing?

libretroshare/src/pqi/pqissl.cc
```diff
@@ -1300,13 +1231,18 @@ int pqissl::accept( SSL *ssl, int fd,
...
-#endif
+	Dbg3() << __PRETTY_FUNCTION__ << std::endl;
+
+	constexpr int failure = -1;
+	constexpr int success = 1;
 
+#ifdef RS_PQISSL_BANLIST_REDUNDANT_CHECK
+	/* TODO: It make no sense to check banlist at this point, as we are actively
+	 * attempting the connection, we decide the address to which to connect to,
+	 * banned addresses should never get here */
```
csoler: Well, the thing is: they do. This test is not costly at all, and in a very critical place. So you really should keep it!

libretroshare/src/pqi/sslfns.cc
```diff
@@ -711,67 +705,39 @@ uint32_t getX509RetroshareCertificateVersion(X509 *cert)
...
-	X509 *x509 = PEM_read_X509(tmpfp, NULL, NULL, NULL);
+	X509* x509 = PEM_read_X509(tmpfp, nullptr, nullptr, nullptr);
 	fclose(tmpfp);
 
-	// check the certificate.
-	bool valid = false;
-	if (x509)
+	if(!x509)
 	{
-                valid = CheckX509Certificate(x509);
```
csoler, line 740: normally this function should check the self-signature of the certificate. It should never accept to return the PGP ID without checking the signature. In the existing implementation, that was apparently removed, but it should be added back!

libretroshare/src/pqi/pqissl.cc
csoler: Generally speaking, because we're in a place where we check the crypto, we should never assume things. We should check them, especially if it costs nothing. It's like in a plane: you have 3 times every vital component and most of the time you only use the 1st one.

csoler (changes requested):
Nice cleaning work. I mostly have comments, but there's 2-3 important things to fix near the end.

libretroshare/src/pqi/pqissl.cc
G10h4ck: I expected you would think it like that, so I kept it, the ifdef there is just to make it easy to test if RS actually behave the same if redundant checks are disabled

libretroshare/src/pqi/authssl.cc
G10h4ck: we discussed that much after I prepared this PR ;)

libretroshare/src/pqi/authssl.cc
G10h4ck: it is not a shared_ptr it is an unique_ptr

libretroshare/src/pqi/authssl.cc
G10h4ck: we talked this in chat

libretroshare/src/pqi/authssl.cc
G10h4ck: It prints it as information just after, when the verification either success or fail

libretroshare/src/pqi/authssl.cc
G10h4ck: Yeah it is basically a redundant duplicate of AuthX509WithGPG

libretroshare/src/pqi/authssl.cc
G10h4ck: This PR is not meant to change the behavior, only to make the code more readable, by deleting old cruft, and evidence redundant checks/code

libretroshare/src/pqi/authssl.cc
G10h4ck: I moved it as Doxygen documentaion in athssl.h were it should stay ;)

libretroshare/src/pqi/authssl.cc
G10h4ck: I am not really sure about the usage of those member variables so I didn't touch that

libretroshare/src/pqi/sslfns.cc
G10h4ck: NO. `CheckX509Certificate` did nothing only returned true, so I removed that from the code, because it was just misleading readers to believe we where checking something while we were checking nothing as the function implementation was literally `return true;`

csoler: That's what i'm saying! Instead of removing it, we should make it right. Now, not only is is not called but the code saying "please fix this" is gone. That's exactly what we dont want ;-)

libretroshare/src/pqi/pqissllistener.cc
```diff
@@ -830,91 +763,73 @@ int pqissllistener::status()
...
-	RsPeerId newPeerId;
-
+	RsPgpId pgpId = RsX509Cert::getCertIssuer(*peercert);
+	RsPeerId newPeerId = RsX509Cert::getCertSslId(*peercert);
 
-	/****
-	 * As the validation is actually done before this...
-	 * we should only need to call CheckCertificate here!
-	 ****/
+#ifdef RS_PQISSL_AUTH_REDUNDANT_CHECK
```
csoler: you should remove that ifdef too. These checks are here for double-check security reasons. Your comment below will tell that the check is redundant, but it is still important to do it.

libretroshare/src/pqi/sslfns.cc
G10h4ck: AFAIU RetroShare SSL certificates are not self signed certificates in the SSL strict sense, because `issuer` (PGP id) and `subject` (location name?) are different, the SSL cert is actually signed with the PGP key not with it's own private key, and we have `AuthX509WithGPG` to validate PGP signature.

libretroshare/src/pqi/pqissl.cc
```diff
@@ -1300,13 +1231,20 @@ int pqissl::accept( SSL *ssl, int fd,
...
 {
-#ifdef PQISSL_DEBUG
-	std::cerr << __PRETTY_FUNCTION__ << std::endl;
-#endif
+	Dbg3() << __PRETTY_FUNCTION__ << std::endl;
 
+	constexpr int failure = -1;
+	constexpr int success = 1;
+
+#ifdef RS_PQISSL_BANLIST_REDUNDANT_CHECK
```
csoler: Same here: the comment tells that the check is redundant but we will always perform this check whatsoever, so the ifdef is not useful.

libretroshare/src/pqi/pqissllistener.cc
G10h4ck: RS_PQISSL_AUTH_REDUNDANT_CHECK = 1,  the ifdef is there only to make it easy to test that those checks are redundant, but enabling/disabling it in a testing session and verifying the behaviour is the same without changing the code

G10h4ck: RS_PQISSL_AUTH_REDUNDANT_CHECK = 1,  the ifdef is there is enabled and only to make it easy to test that those checks are redundant, but enabling/disabling it in a testing session and verifying the behaviour is the same without changing the code

libretroshare/src/pqi/pqissl.cc
G10h4ck: RS_PQISSL_BANLIST_REDUNDANT_CHECK = 1, the ifdef is there is enabled and only to make it easy to test that those checks are redundant, but enabling/disabling it in a testing session and verifying the behaviour is the same without changing the code

csoler: yes, until someone removes them in order to "optimize" because they are called "redundant".

G10h4ck: well they are, but we would review that PR too so no problem here ;)

zapek: Just call it DOUBLE_CHECK? It's more correct.

G10h4ck: If Cyril like that I can change redundant to double

csoler: Double check is ok with me. My fear about "redundant" is that someone will remove it non necessarily on purpose.

libretroshare/src/pqi/pqissllistener.cc
G10h4ck: Renamed to RS_PQISSL_AUTH_DOUBLE_CHECK

libretroshare/src/pqi/pqissl.cc
G10h4ck: Renamed to RS_PQISSL_BANLIST_DOUBLE_CHECK

libretroshare/src/pqi/sslfns.cc
G10h4ck: resolved in 16d606b51388ddf7fa2b70dbffd31d894f52978d

PR #1561, opened 2019-05-05
Prepare RsBanList for JSON API

G10h4ck:


libretroshare/src/services/p3banlist.cc
```diff
@@ -1304,4 +1299,4 @@ int p3BanList::printBanSources_locked(std::ostream &out)
     return true ;
 }
 
-
+RsBanList::~RsBanList() = default;
```
csoler, line 1302: In terms of how much size is used and code clarity, I prefer the "virtual ~RsBanList() {}" in the .h: it removes the need to go into the cpp file to figure out what this destructor is actually doing to discover that it does nothing.

csoler (approved):
Not much to say ;-) See my only comment about the destructor.

libretroshare/src/services/p3banlist.cc
csoler: ...and it spares the linking job to the compiler too.

G10h4ck: I do prefer too (`=default` can be used in the header too), but the compiler print a warning about the vtable being emitted in every compilation unit that includes the headed if no method is defined outside the header, so moved the empty destructor in an the implementation file

PR #1565, opened 2019-05-12
Short invites support for libretroshare

G10h4ck:
With this PR it is possible to add a friend based on her SSL id only so the following are all valid equivalents short invites:

AImkEBWA3WHBGeSzzy2+U/cBAAAAB1Rlc3QgMDSSHDs7XB84
https://retroshare.me?rsInvite=AImkEBWA3WHBGeSzzy2%2BU%2FcBAAAAB1Rlc3QgMDSSHDs7XB84
retroshare:///?rsInvite=AImkEBWA3WHBGeSzzy2%2BU%2FcBAAAAB1Rlc3QgMDSSHDs7XB84
rs:///?rsInvite=AImkEBWA3WHBGeSzzy2%2BU%2FcBAAAAB1Rlc3QgMDSSHDs7XB84
https://elrepo.io/?rsInvite=AImkEBWA3WHBGeSzzy2%2BU%2FcBAAAAB1Rlc3QgMDSSHDs7XB84
http://whadeverdomain.tld/whatewerpath/?rsInvite=AImkEBWA3WHBGeSzzy2%2BU%2FcBAAAAB1Rlc3QgMDSSHDs7XB84

Those invites contains SSL id + PGP name + external IP, hidden locations and dyndns are supported too, passing `bareBones == true` the link would contain only  SSL id + PGP name so the link is even shorter but only incoming connections are possible. 

This make it possible to fit a retroshare invite in low size mediums like a QR code or a broadcast UDP packet, or to sneak it as an common looking link into other Apps messages etc, also this way it is more flexible to register retroshare invites handling across platforms.

I have successfully tested the functionalities introduced by this pull request through JSON API.

### DIFF
--- a/libretroshare/src/gossipdiscovery/p3gossipdiscovery.cc
+++ b/libretroshare/src/gossipdiscovery/p3gossipdiscovery.cc
@@ -27,6 +27,7 @@
 #include "retroshare/rsiface.h"
 #include "rsserver/p3face.h"
 #include "util/rsdebug.h"
+#include "retroshare/rspeers.h"
 
 /****
  * #define P3DISC_DEBUG	1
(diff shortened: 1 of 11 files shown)


felisucoibi:
i love this pr, somethin we needed long time ago

PR #1566, opened 2019-05-16
added new qdarkstyle version 2.6.8

defnax:
added new version of qdarkstyle to retroshare, nobody doing this things now created from the latest 
https://github.com/ColinDuquesnoy/QDarkStyleSheet/releases

@felisucoibi check this looks little bit different then the older version


felisucoibi:
is ok

defnax:
> is ok

but you are right the blue rs links are not readable good
other issue is with bold notify text.
maybe there needs some yellow text color when there is new count then its better for your eyes.

i using now often qdarkstyle and seeing some issues

PR #1569, opened 2019-05-19
Fix clang warnings

PhenomRetroShare:


G10h4ck:
I have reviewed changes made to libretroshare seems ok, for this one i do prefer @csoler to boule check https://github.com/RetroShare/RetroShare/pull/1569/commits/729cb56d9ab5eedb67ac9c830df80365fbfb942b     about RetroShare-gui I leave that for @csoler to review ;)

G10h4ck (approved):
See previous comment

libresapi/src/api/PeersHandler.cpp
```diff
@@ -643,7 +643,7 @@ void PeersHandler::handleWildcard(Request &req, Response &resp)
 			std::string error_string;
 			RsPeerDetails peerDetails;
 
-			RsPgpId own_pgp = mRsPeers->getGPGOwnId();
+			/*RsPgpId own_pgp =*/ mRsPeers->getGPGOwnId();
```
csoler: I dont see the point in leaving this statement. It's not doing anything, is it? If so, just remove the line.

libretroshare/src/pqi/authssl.cc
```diff
@@ -47,13 +47,13 @@
 
 /* SSL connection diagnostic */
 
-const uint32_t RS_SSL_HANDSHAKE_DIAGNOSTIC_UNKNOWN                     = 0x00 ;
+//const uint32_t RS_SSL_HANDSHAKE_DIAGNOSTIC_UNKNOWN                     = 0x00 ;
```
csoler, line 50: Please keep these 3. Maybe they are not used for now but they should in the future when we add more feedback on certificates. Also the first one is here to have a name to uninitialized/default values.

libretroshare/src/rs_upnp/UPnPBase.cpp
```diff
@@ -369,6 +369,8 @@ m_direction           (upnpLib.Element_GetChildValueByTag(argument, "direction")
 m_retval              (upnpLib.Element_GetFirstChildByTag(argument, "retval")),
 m_relatedStateVariable(upnpLib.Element_GetChildValueByTag(argument, "relatedStateVariable"))
 {
+	/* remove unused private field warnings */
+	(void) m_UPnPControlPoint;
```
csoler, line 373: I would really prefer that you keep the variables into /*  */ in the function prototype. The effect on compilation will be the same but it spares the use of a statement that does nothing. Maybe it's a matter of taste but I find this very ugly ;-)

csoler: The same happens in many places in this PR. Can you update them all?

libretroshare/src/services/p3idservice.cc
```diff
@@ -4336,7 +4336,7 @@ void p3IdService::generateDummy_OwnIds()
 
 	/* grab all the gpg ids... and make some ids */
 
-	RsPgpId ownId = mPgpUtils->getPGPOwnId();
+	mPgpUtils->getPGPOwnId();
```
csoler: just remove the line if ownId is not used.

retroshare-gui/src/gui/settings/ServerPage.cpp
```diff
@@ -62,11 +62,11 @@
 ///
 
 // Tabs numbers *after* non relevant tabs are removed. So do not use them to add/remove tabs!!
-const static uint32_t TAB_HIDDEN_SERVICE_OUTGOING = 0;
+//const static uint32_t TAB_HIDDEN_SERVICE_OUTGOING = 0;
```
csoler, line 65: Please keep these. It prevents the next code cleaning to actually rename them from 0 to n-1 and therefore cause bugs in the addressing of tabs.

csoler (changes requested):
Thx for all the fixes. I left a few requests for things to change.

libresapi/src/api/PeersHandler.cpp
PhenomRetroShare: Just in case getGPGOwnId initialize something...
But doesn't seems the case.

libretroshare/src/pqi/authssl.cc
PhenomRetroShare: They stay in comment so you can uncomment them when needed.
They are const not enum.

libretroshare/src/rs_upnp/UPnPBase.cpp
PhenomRetroShare: These m_UPnPControlPoint is not used in our code but seems to be in UPnP one.
If not initialized, UPnP doesn't work.

libretroshare/src/services/p3idservice.cc
PhenomRetroShare: That's here, that tell to me that getPGPOwnId initialize something as this function only call it for now.
I remove it.

retroshare-gui/src/gui/settings/ServerPage.cpp
PhenomRetroShare: Comment left to not doing that.
And this not prevents to add a new intermediate Tab in ui file, like it happens sometimes.

libretroshare/src/rs_upnp/UPnPBase.cpp
csoler: But you're not initializing here. You're just tricking the compiler in not giving a warning. This statement should just be removed.

PhenomRetroShare: This commit is here to remove these warnings.
Do you have a better way to do it?

PhenomRetroShare:
Any news?

Issue #638, opened 2017-01-11
Implement New Invite System


defnax:
#**New Invite System**

Allow one way Add Friend via Key, on other side Friend gets Friend Reqeust (dont need both sides to add key)Not sure when this will be implemented? use DHT?
If we develope a RetroShare for all, 12 to 60 years old people must be able to use it, it must be easy to reach more people with RetroShare.

**Introduction**
This page is the Overview for a Proposed Invitation System for RetroShare.
It is supposed to list the desired requirements for the System, and how closely the proposals meet these requirements.

**Use Cases**
1 - Two people find retroshare, and want to use it.
2 - An existing retroshare user wants to invite a friend to use retroshare.
3 - two retroshare users (unconnected via gpg graph) want to be friends.
4 - two retroshare users (connected via gpg graph) want to be friends.
5 - a retroshare user wants to introduce two friends to each other.
6 - a retroshare user wants an introduction to a friend of a friend.

Requirements

**Mandatory**
1) Authentication Reliable

**Essential**
1) Ability to attach a personal message.
2) Simple to Use.
3) Works for Firewalled Peers.
4) Works for Offline Peers
5) Works with Unknown Peers.
6) Works with Friends of Friends.
7) Indication if Invitation impossible

**Desirable**
1) Unique Code to indicate Invitation.
2) Small UniqueCode Size.
3) Single Click.
4) Invitation Times Out.
5) Invitation Completed Immediately (No Delay between Invitation Received and Signing Key)
6) Allows Introductions. 

 **Proposals Overview** 
![rs](https://cloud.githubusercontent.com/assets/9952056/21830717/f1f5ccec-d79e-11e6-830d-c742f1803c47.png)

**## Invitation system: DHT IP STORE**

For me the invitation should be an easy way to make friends, while keeping the key signature paradigm.

**Two typical cases:**

1 - A adds a new friend B that does not understand much about gpg, signatures, etc. In such a case the "invite friend" button should propose to generate a small string/passwd/whatever that A sends to B

2 - A wants to invite a peer from his network tab, and let him know

I think it's possible to handle both cases this way:

**A wants to make friend with B**

-> A generates an invitation code.
-> B gets the code, add it to RS. RS pops a window at B, asking to sign A's key
-> A gets informed that B did. RS pops a window at A, asking to sign B's key

Depending on the configuration (1 or 2, as listed above) some of these steps might be skipped

**Proposed Algorithm**
One option for implementing this is:

[...]

PR #1571, opened 2019-05-28
V0.6 short invites

csoler:
WIP: short invites. GUI part is on its way.


libretroshare/src/rsserver/p3peers.cc
```diff
@@ -1342,6 +1344,7 @@ bool p3Peers::parseShortInvite(const std::string& inviteStrUrl, RsPeerDetails& d
 		default:
 			RsWarn() << __PRETTY_FUNCTION__ << " got unkown field type: "
 			         << static_cast<uint32_t>(fieldType) << std::endl;
+            return false;
```
G10h4ck: In case we get an unknown field we should just print a warning and ignore it, just in case we decide to add some new field in newer versions, so we can stay compatible with the past versions. Did you remember we had to do the same change on old certs? Maybe we need to add filed length too for achieving that?

G10h4ck (changes requested):

libretroshare/src/rsserver/p3peers.cc
csoler: I agree on the principle. In this case I had an invite that did not parse and therefore the parser is actually going through the entire data byte by byte. This is the consequence of not storing the size of each field (which we actually do in RS certificate format) in which case it's more easy to distinguish random data (that has inconsistent size) from correct field with unknown tag (that has consistent size). So it seems to me that because we cannot decide, it's safer to return false otherwise, when parsing random data you'd easily end up with a random supposedly valid SSL id and PGP fingerprint.

csoler: Another way to see the problem is that if you have 1 unknown field in the middle of the invite, the parser will not be able to skip it. It will instead try to interpret each and every byte in this field as a potential tag. That's likely to make the parsing completely wrong and even prevent the correct reading of the remaining fields. It seems to me that it would be safer to add sizes (which is 1 byte most of the time in the 1-2-5 encoding).

libretroshare/src/gossipdiscovery/gossipdiscoveryitems.h
```diff
@@ -34,11 +35,12 @@
...
 {
 	PGP_LIST           = 0x1,
-	PGP_CERT           = 0x2,
+	PGP_CERT           = 0x2,		// deprecated
 	CONTACT            = 0x5,
 	IDENTITY_LIST      = 0x6,
-	INVITE             = 0x7,
-	INVITE_REQUEST     = 0x8
+	INVITE             = 0x7,		// deprecated
+	INVITE_REQUEST     = 0x8,		// deprecated
```
G10h4ck: We should remove this two AFAIR those two never got into an official release

libretroshare/src/gossipdiscovery/gossipdiscoveryitems.cc
```diff
@@ -36,14 +36,11 @@ RsItem *RsDiscSerialiser::create_item(
...
 	case RsGossipDiscoveryItemType::CONTACT:  return new RsDiscContactItem();
-	case RsGossipDiscoveryItemType::IDENTITY_LIST:
-		return new RsDiscIdentityListItem();
-	case RsGossipDiscoveryItemType::INVITE:
-		return new RsGossipDiscoveryInviteItem();
-	case RsGossipDiscoveryItemType::INVITE_REQUEST:
-		return  new RsGossipDiscoveryInviteRequestItem();
+	case RsGossipDiscoveryItemType::IDENTITY_LIST: return new RsDiscIdentityListItem();
+    default:
+        return NULL;
```
G10h4ck: please return `nullptr` instead of `NULL` so the compiler doesn't complain

libretroshare/src/gossipdiscovery/gossipdiscoveryitems.h
```diff
@@ -80,20 +82,18 @@ class RsDiscPgpListItem: public RsDiscItem
...
 	void clear() override;
-	void serial_process(
-	        RsGenericSerializer::SerializeJob j,
-	        RsGenericSerializer::SerializeContext& ctx) override;
+	void serial_process( RsGenericSerializer::SerializeJob j, RsGenericSerializer::SerializeContext& ctx) override;
 
-	RsPgpId pgpId;
-	std::string pgpCert;
+	RsPgpId pgpKeyId;				// duplicate information for practical reasons
+	RsTlvBinaryData pgpKeyData;
```
G10h4ck: If possible please avoid using TLV stuff as they are completely opaque to JSON serialization, and we might regret using them at some point.

libretroshare/src/gossipdiscovery/p3gossipdiscovery.cc
```diff
@@ -293,28 +290,18 @@ int p3discovery2::handleIncoming()
...
-		{
-			sendInvite(inviteReq->mInviteId, item->PeerId());
-			delete item;
-		}
 		else
 		{
-			RsWarn() << __PRETTY_FUNCTION__ << " Received unknown item type! "
-			         << std::endl << item << std::endl;
+			RsWarn() << __PRETTY_FUNCTION__ << " Received unknown item type " << (int)item->PacketSubType() << "! " << std::endl ;
+            RsWarn() << item << std::endl;
```
G10h4ck, line 296: This will cause the warning prefix and timestamp being printed again, probably it is better to remove the second `RsWarn()` and `;` in the first line so it is part of the same message

libretroshare/src/gossipdiscovery/p3gossipdiscovery.cc
```diff
@@ -376,6 +364,21 @@ void p3discovery2::recvOwnContactInfo(const SSLID &fromId, const RsDiscContactIt
...
 	std::cerr << std::endl;
 #endif
+	// Check that the "own" ID sent corresponds to the one we think it should be.
+	// Some of these checks may look superfluous but it's better to risk to check twice than not check at all.
+
+    // was obtained using a short invite. , and that the friend is marked as "ignore PGP validation" because it
+	RsPeerDetails det ;
+	if(!rsPeers->getPeerDetails(fromId,det))
+	{
+		std::cerr << "(EE) Cannot obtain details from " << fromId << " who is supposed to be a friend! Dropping the info." << std::endl;
```
G10h4ck: You should use `RsErr() << __PRETTY_FUNCTION__ <<` instead of `std::cerr` here

libretroshare/src/gossipdiscovery/p3gossipdiscovery.cc
```diff
@@ -376,6 +364,21 @@ void p3discovery2::recvOwnContactInfo(const SSLID &fromId, const RsDiscContactIt
...
+		return;
+	}
+	if(det.gpg_id != item->pgpId)
+	{
+		std::cerr << "(EE) peer " << fromId << " sent own details with PGP key ID " << item->pgpId << " which does not match the known key id " << det.gpg_id << ". Dropping the info." << std::endl;
```
G10h4ck: You should use `RsErr() << __PRETTY_FUNCTION__ <<` instead of `std::cerr` here

libretroshare/src/gossipdiscovery/p3gossipdiscovery.cc
```diff
@@ -388,6 +391,17 @@ void p3discovery2::recvOwnContactInfo(const SSLID &fromId, const RsDiscContactIt
 
 	updatePeerAddresses(item);
 
+    // if the peer is not validated, we stop the exchange here
+
+    if(det.skip_signature_validation)
+    {
+#ifdef P3DISC_DEBUG
+		std::cerr << "p3discovery2::recvOwnContactInfo() missing PGP key  " << item->pgpId << " from short invite friend " << fromId << ". Requesting it." << std::endl;
```
G10h4ck, line 402: You should use `DbgX() << __PRETTY_FUNCTION__ <<` instead

libretroshare/src/gossipdiscovery/p3gossipdiscovery.cc
```diff
@@ -683,13 +686,13 @@ void p3discovery2::processPGPList(const SSLID &fromId, const RsDiscPgpListItem *
...
 	{
-		std::set<PGPID>::const_iterator fit;
+		std::set<RsPgpId>::const_iterator fit;
 		for(fit = item->pgpIdSet.ids.begin(); fit != item->pgpIdSet.ids.end(); ++fit)
 		{
 			if (!AuthGPG::getAuthGPG()->isGPGId(*fit))
 			{
 #ifdef P3DISC_DEBUG
-				std::cerr << "p3discovery2::processPGPList() requesting PgpId: " << *fit;
+				std::cerr << "p3discovery2::processPGPList() requesting certificate for PgpId: " << *fit;
```
G10h4ck, line 707: You should use `DbgX() << __PRETTY_FUNCTION__ <<` instead

libretroshare/src/gossipdiscovery/p3gossipdiscovery.cc
```diff
@@ -768,16 +771,16 @@ void p3discovery2::updatePeers_locked(const SSLID &aboutId)
...
 					mit != ffit->second.mSslIds.end(); ++mit)
 			{
-				SSLID sslid = mit->first;
+				RsPeerId sslid = mit->first;
 				if (mServiceCtrl->isPeerConnected(getServiceInfo().mServiceType, sslid))
 				{
 					// TODO IGNORE if sslid == aboutId, or sslid == ownId.
 #ifdef P3DISC_DEBUG
-					std::cerr << "p3discovery2::updatePeer_locked() Adding Online SSLID: " << sslid;
+					std::cerr << "p3discovery2::updatePeer_locked() Adding Online RsPeerId: " << sslid;
```
G10h4ck, line 791: You should use `DbgX() << __PRETTY_FUNCTION__ <<` instead

libretroshare/src/gossipdiscovery/p3gossipdiscovery.cc
```diff
@@ -882,14 +885,14 @@ void p3discovery2::sendContactInfo_locked(const PGPID &aboutId, const SSLID &toI
 		else
 		{
 #ifdef P3DISC_DEBUG
-			std::cerr << "p3discovery2::sendContactInfo_locked() SSLID Hidden";
+			std::cerr << "p3discovery2::sendContactInfo_locked() RsPeerId Hidden";
```
G10h4ck, line 894: You should use `DbgX() << __PRETTY_FUNCTION__ <<` instead

libretroshare/src/gossipdiscovery/p3gossipdiscovery.cc
```diff
@@ -905,12 +908,12 @@ void p3discovery2::processContactInfo(const SSLID &fromId, const RsDiscContactIt
...
 	}
 
-	std::map<PGPID, DiscPgpInfo>::iterator it;
+	std::map<RsPgpId, DiscPgpInfo>::iterator it;
 	it = mFriendList.find(item->pgpId);
 	if (it == mFriendList.end())
 	{
 #ifdef P3DISC_DEBUG
-		std::cerr << "p3discovery2::processContactInfo(" << fromId << ") PGPID: ";
+		std::cerr << "p3discovery2::processContactInfo(" << fromId << ") RsPgpId: ";
```
G10h4ck, line 924: You should use `DbgX() << __PRETTY_FUNCTION__ <<` instead

libretroshare/src/gossipdiscovery/p3gossipdiscovery.cc
```diff
@@ -1022,30 +1025,86 @@ void p3discovery2::recvPGPCertificateRequest(
...
+    unsigned char *bin_data;
+    size_t bin_len;
 
-	{
-		RS_STACK_MUTEX(mDiscMtx);
-		mPendingDiscPgpCertOutList.push_back(item);
-	}
+	if(!AuthGPG::getAuthGPG()->exportPublicKey(aboutId,bin_data,bin_len,false,true))
+    {
+        std::cerr << "(EE) cannot export public key " << aboutId << " requested by peer " << toId << std::endl;
```
G10h4ck, line 1061: You should use `RsErr() << __PRETTY_FUNCTION__ <<` instead

libretroshare/src/gossipdiscovery/p3gossipdiscovery.cc
```diff
@@ -1022,30 +1025,86 @@ void p3discovery2::recvPGPCertificateRequest(
...
 {
+    // 1 - check that the cert structure is valid.
+
+	RsPgpId cert_pgp_id;
+	std::string cert_name;
+	std::list<RsPgpId> cert_signers;
+
+	if(!AuthGPG::getAuthGPG()->getGPGDetailsFromBinaryBlock( (unsigned char*)item->pgpKeyData.bin_data,item->pgpKeyData.bin_len, cert_pgp_id, cert_name, cert_signers ))
+	{
+		std::cerr << "(EE) cannot parse own PGP key sent by " << fromId << std::endl;
```
G10h4ck, line 1081: You should use `RsErr() << __PRETTY_FUNCTION__ <<` instead

libretroshare/src/gossipdiscovery/p3gossipdiscovery.cc
```diff
@@ -1022,30 +1025,86 @@ void p3discovery2::recvPGPCertificateRequest(
...
+		return;
+	}
+
+	if(cert_pgp_id != item->pgpKeyId)
+	{
+		std::cerr << "(EE) received a PGP key from " << fromId << " which ID (" << cert_pgp_id << ") is different from the one anounced in the packet (" << item->pgpKeyId << ")!" << std::endl;
```
G10h4ck, line 1087: You should use `RsErr() << __PRETTY_FUNCTION__ <<` instead

libretroshare/src/gossipdiscovery/p3gossipdiscovery.cc
```diff
@@ -1022,30 +1025,86 @@ void p3discovery2::recvPGPCertificateRequest(
...
+		return;
+	}
+
+    // 2 - check if the peer who is sending us a cert is already validated
+
+    RsPeerDetails det;
+    if(!rsPeers->getPeerDetails(fromId,det))
+    {
+        std::cerr << "(EE) cannot get peer details from friend " << fromId << ": this is very wrong!"<< std::endl;
```
G10h4ck, line 1096: You should use `RsErr() << __PRETTY_FUNCTION__ <<` instead

libretroshare/src/gossipdiscovery/p3gossipdiscovery.cc
```diff
@@ -1022,30 +1025,86 @@ void p3discovery2::recvPGPCertificateRequest(
...
+        return;
+    }
+
+    // We treat own pgp keys right away when they are sent by a friend for which we dont have it. This way we can keep the skip_pgg_signature_validation consistent
+
+    if(det.skip_signature_validation)
+    {
 #ifdef P3DISC_DEBUG
-	std::cerr << __PRETTY_FUNCTION__ << " queuing for Cert loading" << std::endl;
+		std::cerr << __PRETTY_FUNCTION__ << " Received own full certificate from short-invite friend " << fromId << std::endl;
```
G10h4ck, line 1105: You should use `DbgX() << __PRETTY_FUNCTION__ <<` instead

libretroshare/src/gossipdiscovery/p3gossipdiscovery.cc
```diff
@@ -1022,30 +1025,86 @@ void p3discovery2::recvPGPCertificateRequest(
...
 #endif
-	/* push this back to be processed by pgp when possible */
-	RS_STACK_MUTEX(mDiscMtx);
-	mPendingDiscPgpCertInList.push_back(item);
+        // do some extra checks. Dont remove them. They cost nothing as compared to what they could avoid.
+
+		if(item->pgpKeyId != det.gpg_id)
+        {
+			std::cerr << "(EE) received a PGP key with ID " << item->pgpKeyId << " from non validated peer " << fromId << ", which should only be allowed to send his own key " << det.gpg_id << std::endl;
```
G10h4ck, line 1111: You should use `RsErr() << __PRETTY_FUNCTION__ <<` instead

libretroshare/src/gossipdiscovery/p3gossipdiscovery.cc
```diff
@@ -1022,30 +1025,86 @@ void p3discovery2::recvPGPCertificateRequest(
...
+			return;
+		}
+	}
+	RsPgpId tmp_pgp_id;
+	std::string error_string;
+
+#ifdef P3DISC_DEBUG
+	std::cerr << __PRETTY_FUNCTION__ << "Received PGP key " << cert_pgp_id << " from from friend " << fromId << ". Adding to keyring." << std::endl;
```
G10h4ck, line 1119: You should use `DbgX() << __PRETTY_FUNCTION__ <<` instead

libretroshare/src/gossipdiscovery/p3gossipdiscovery.cc
```diff
@@ -1138,51 +1197,13 @@ bool p3discovery2::getDiscFriends(const RsPeerId& id, std::list<RsPeerId> &proxy
 bool p3discovery2::getWaitingDiscCount(size_t &sendCount, size_t &recvCount)
 {
 	RS_STACK_MUTEX(mDiscMtx);
-	sendCount = mPendingDiscPgpCertOutList.size();
-	recvCount = mPendingDiscPgpCertInList.size();
+	sendCount = 0;//mPendingDiscPgpCertOutList.size();
+	recvCount = 0;//mPendingDiscPgpCertInList.size();
```
G10h4ck, line 1223: Should we remove this method completely instead?

libretroshare/src/pqi/p3peermgr.cc
```diff
@@ -338,17 +338,31 @@ bool p3PeerMgrIMPL::isFriend(const RsPeerId& id)
...
         bool ret = (mFriendList.end() != mFriendList.find(id));
 #ifdef PEER_DEBUG_COMMON
                 std::cerr << "p3PeerMgrIMPL::isFriend(" << id << ") returning : " << ret << std::endl;
 #endif
         return ret;
 }
+bool p3PeerMgrIMPL::isSslOnlyFriend(const RsPeerId& id)
+{
+#ifdef PEER_DEBUG_COMMON
+                std::cerr << "p3PeerMgrIMPL::isFriend(" << id << ") called" << std::endl;
```
G10h4ck, line 351: You should use `DbgX() << __PRETTY_FUNCTION__ <<` instead

libretroshare/src/pqi/p3peermgr.cc
```diff
@@ -338,17 +338,31 @@ bool p3PeerMgrIMPL::isFriend(const RsPeerId& id)
...
+#endif
+		RS_STACK_MUTEX(mPeerMtx);
+        auto it = mFriendList.find(id);
+        bool ret = it != mFriendList.end() && it->second.skip_pgp_signature_validation ;
+
+#ifdef PEER_DEBUG_COMMON
+                std::cerr << "p3PeerMgrIMPL::isFriend(" << id << ") returning : " << ret << std::endl;
```
G10h4ck, line 358: You should use `DbgX() << __PRETTY_FUNCTION__ <<` instead

libretroshare/src/pqi/p3peermgr.cc
```diff
@@ -896,6 +909,28 @@ bool    p3PeerMgrIMPL::haveOnceConnected()
 }
 
 
+bool p3PeerMgrIMPL::notifyPgpKeyReceived(const RsPgpId& pgp_id)
+{
+	RsStackMutex stack(mPeerMtx); /****** STACK LOCK MUTEX *******/
```
G10h4ck, line 901: you should use `RS_STACK_MUTEX(...)` instead

libretroshare/src/pqi/p3peermgr.cc
```diff
@@ -896,6 +909,28 @@ bool    p3PeerMgrIMPL::haveOnceConnected()
...
+
+    bool changed = false;
+
+    for(auto it(mFriendList.begin());it!=mFriendList.end();++it)
+    {
+        if(it->second.gpg_id == pgp_id)
+        {
+            std::cerr << "(WW) notification that full key " << pgp_id << " is available. Reseting short invite flag for peer " << it->first << std::endl;
```
G10h4ck, line 909: You should use `RsWarn() << __PRETTY_FUNCTION__ <<` instead

libretroshare/src/pqi/p3peermgr.cc
```diff
@@ -915,9 +950,7 @@ bool p3PeerMgrIMPL::addFriend(const RsPeerId& input_id, const RsPgpId& input_gpg
 
 		if (id == AuthSSL::getAuthSSL()->OwnId())
 		{
-#ifdef PEER_DEBUG
-			std::cerr << "p3PeerMgrIMPL::addFriend() cannot add own id as a friend." << std::endl;
-#endif
+			RsErr() << "p3PeerMgrIMPL::addFriend() cannot add own id as a friend. That's a bug!" << std::endl;
```
G10h4ck, line 940: use `__PRETTY_FUNCTION__` instead

libretroshare/src/pqi/p3peermgr.cc
```diff
@@ -932,13 +965,17 @@ bool p3PeerMgrIMPL::addFriend(const RsPeerId& input_id, const RsPgpId& input_gpg
...
-#ifdef PEER_DEBUG
-			std::cerr << "p3PeerMgrIMPL::addFriend() Already Exists" << std::endl;
-#endif
-			/* (1) already exists */
-			return true;
+            // The friend may already be here, including with a short invite (meaning the PGP key is unknown).
+
+			if(it->second.gpg_id != input_gpg_id)// already exists as a friend with a different PGP id!!
+            {
+                RsErr() << "Trying to add SSL id (" << id << ") that is already a friend with existing PGP key (" << it->second.gpg_id << ") but using a different PGP key (" << input_gpg_id << "). This is a bug!" << std::endl;
```
G10h4ck, line 961: add `__PRETTY_FUNCTION__`

libretroshare/src/pqi/p3peermgr.cc
```diff
@@ -973,6 +1011,15 @@ bool p3PeerMgrIMPL::addFriend(const RsPeerId& input_id, const RsPgpId& input_gpg
 			it->second.netMode  = netMode;
 			it->second.lastcontact = lastContact;
 
+			if(!it->second.gpg_id.isNull() && it->second.gpg_id != input_gpg_id)// already exists as a friend with a different PGP id!!
+            {
+                RsErr() << "Trying to add SSL id (" << id << ") that is already known (but not friend) with existing PGP key (" << it->second.gpg_id << ") but using a different PGP key (" << input_gpg_id << "). This is a bug!" << std::endl;
```
G10h4ck: add `__PRETTY_FUNCTION__`

libretroshare/src/pqi/p3peermgr.cc
```diff
@@ -1029,6 +1079,93 @@ bool p3PeerMgrIMPL::addFriend(const RsPeerId& input_id, const RsPgpId& input_gpg
 	return true;
 }
 
+
+bool p3PeerMgrIMPL::addSslOnlyFriend( const RsPeerId& sslId, const RsPgpId& pgp_id,const RsPeerDetails& dt )
+{
+	if(sslId.isNull() || sslId == getOwnId())
+    {
+        RsErr() <<"Attempt to add yourself or a null ID as SSL-only friend (id=" << sslId << ")" << std::endl;
```
G10h4ck, line 1078: add `__PRETTY_FUNCTION__`

libretroshare/src/pqi/p3peermgr.cc
```diff
@@ -1029,6 +1079,93 @@ bool p3PeerMgrIMPL::addFriend(const RsPeerId& input_id, const RsPgpId& input_gpg
...
+			pstate = it->second;
+			mOthersList.erase(it);
+		}
+
+
+	} // RS_STACK_MUTEX(mPeerMtx);
+
+	if(!pstate.gpg_id.isNull() && AuthGPG::getAuthGPG()->isGPGAccepted(pstate.gpg_id))
+	{
+		RsErr() << "Trying to add as SSL-only friend a peer which PGP id is already a friend. This means the code is inconsistent. Not doing this!" << std::endl;
```
G10h4ck, line 1100: add `__PRETTY_FUNCTION__`

libretroshare/src/pqi/p3peermgr.cc
```diff
@@ -1029,6 +1079,93 @@ bool p3PeerMgrIMPL::addFriend(const RsPeerId& input_id, const RsPgpId& input_gpg
...
+		return false;
+	}
+
+    if(pgp_id.isNull())
+    {
+        RsErr() << "Null pgp id for friend added with skip_pgp_signature_validaiton flag. This is not allowed." << std::endl;
```
G10h4ck, line 1106: add `__PRETTY_FUNCTION__`

libretroshare/src/pqi/p3peermgr.cc
```diff
@@ -1602,7 +1739,7 @@ bool p3PeerMgrIMPL::addCandidateForOwnExternalAddress(const RsPeerId &from, cons
 	sockaddr_storage_clear(addr_filtered) ;
 	sockaddr_storage_copyip(addr_filtered,addr) ;
 
-#ifdef PEER_DEBUG
+#ifndef PEER_DEBUG
```
G10h4ck, line 1714: You should use `DbgX() << __PRETTY_FUNCTION__ <<` instead

libretroshare/src/retroshare/rsgossipdiscovery.h
```diff
@@ -64,6 +64,8 @@ struct RsGossipDiscoveryFriendInviteReceivedEvent : RsEvent
 class RsGossipDiscovery
 {
 public:
+    virtual ~RsGossipDiscovery() {}
```
G10h4ck: `= default;` should be used instead if `{}` because the compiler can do more optimizations

libretroshare/src/retroshare/rspeers.h
```diff
@@ -227,6 +238,7 @@ struct RsPeerDetails : RsSerializable
 	uint32_t trustLvl;
 	uint32_t validLvl;
 
+    bool skip_signature_validation;
```
G10h4ck: should this be renamed to `skip_pgp_signature_validation` for easier understanding?

libretroshare/src/retroshare/rspeers.h
```diff
@@ -466,6 +505,11 @@ class RsPeers
 	virtual bool getAssociatedSSLIds(const RsPgpId& gpg_id, std::list<RsPeerId>& ids) = 0;
 	virtual bool gpgSignData(const void *data, const uint32_t len, unsigned char *sign, unsigned int *signlen, std::string reason = "") = 0;
 
+    virtual RsPgpId pgpIdFromFingerprint(const RsPgpFingerprint& fpr) = 0;
+
+    // Note: the two methods below could be unified. The fact that one of them can take an optional RsPeerDetails struct as parameter
+    // seems quite inconsistent.
```
G10h4ck, line 511: Is this still relevant? The optional parameter is to pass optional extra data such as ipv6 addresses etc.

libretroshare/src/rsserver/p3peers.cc
```diff
@@ -570,6 +574,22 @@ std::string p3Peers::getGPGName(const RsPgpId &gpg_id)
...
 
+bool p3Peers::isSslOnlyFriend(const RsPeerId& sslId)
+{
+    bool has_ssl_only_flag = mPeerMgr->isSslOnlyFriend(sslId) ;
+
+    if(has_ssl_only_flag)
+    {
+        if(isPgpFriend(getGPGId(sslId)))
+        {
+            RsErr() << "Peer " << sslId << " has SSL-friend-only flag but his PGP id is in the list of friends. This is inconsistent (Bug in the code). Returning false for security reasons." << std::endl;
```
G10h4ck: should add `__PRETTY_FUNCTION__`

G10h4ck (changes requested):
See comments inline

libretroshare/src/pqi/authssl.cc
```diff
@@ -1260,8 +1272,9 @@ int AuthSSLimpl::VerifyX509Callback(int /*preverify_ok*/, X509_STORE_CTX* ctx)
 		return verificationFailed;
 	}
 
-	if ( !isSslOnlyFriend && pgpId != AuthGPG::getAuthGPG()->getGPGOwnId() &&
-	     !AuthGPG::getAuthGPG()->isGPGAccepted(pgpId) )
+    if(isSslOnlyFriend && pgpId !=
```
G10h4ck: This line breaks compilation, how it end up there? This hunk (line 1275-1277) change nothing apart breaking compilation

G10h4ck (changes requested):

G10h4ck:
TravisCI for linux seems broken it report no error, but it didn't compile at all, did we disable that check at some point?

libretroshare/src/retroshare/rspeers.h
csoler: I dont understand this comment. Nothing to do with IPs here.

libretroshare/src/pqi/authssl.cc
csoler: probably a copy+paste issue. The test of same PGP id is alredy there line 1210

libretroshare/src/rsserver/p3peers.cc
csoler: done

csoler:
This PR needs to be tested. See internal devs forums. Do not trust the security of this PR. It's to be considered flawed until proven otherwise.

defnax:
short invites key great, but the real easy send/accept friend request will never come to rs?
Two side must copy and paste the keys to get be friends ?

`- no discovery information is exchanged with nodes with short invites before the public PGP key has been received.`

If i click accept request, then it must be have a feature to send the key to other site, that will not come?

G10h4ck:
With my initial proposal it was possible since one could add a friend just by knowing his own SSL-id that is available when connection is attempted. I have just checked the current PR that @csoler built on top of my initial work and it is not possible anymore since now to add a friend also PGP-fingerprint is needed which is not available at connection attempt (while PGP-id is available). So either some change is made or the comfortable "Accept" button is not possible to implement anymore -_-. One possible solution could be to send the full PGP-fingerprint in `sslCn` field, but AFAIU this would break connection level retro-compatibility. Another one is add the possibility to add a friend knowing only SSL-id + PGP-id (be careful that PGP-id is just a piece of the full PGP-fingerprint). Anyway we have been discussing to port the whole code to full PGP-fingerprint instead of PGP-id, so at some point (Retroshare-0.7 ?) we will break retro-compatibility anyway. IMHO this possibility to add someone as friend on attempt is a really important usability feature, even in one of the most common social interaction people ask your phone number, and the give you a miscall so you have their number instead of blabbering it...

libretroshare/src/retroshare/rspeers.h
G10h4ck: my comment is about the inline comment, not about `pgpIdFromFingerprint`

G10h4ck:
IMHO the possibility to add someone as friend on attempt is a really important usability feature, even in one of the most common social interaction people ask your phone number, and then give you a miss call so you have their number instead of blabbering it to you...

So I have double checked the code, and it is still possible to implement an accept button after a failed connection attempt notification at user interface level.

I have explained the details on a way to implement that in https://github.com/RetroShare/RetroShare/issues/638
But I think that proposal (only UI work) should be eventually implemented in another PR after this one is merged.

libretroshare/src/pqi/authssl.cc
```diff
@@ -1198,6 +1198,21 @@ int AuthSSLimpl::VerifyX509Callback(int /*preverify_ok*/, X509_STORE_CTX* ctx)
 	std::string sslCn = RsX509Cert::getCertIssuerString(*x509Cert);
 	RsPgpId pgpId(sslCn);
 
+    RsPeerDetails det;
+    if(!rsPeers->getPeerDetails(sslId,det))
```
G10h4ck: This extra check deny connection to a new location of a PGP friend, IMHO we should remove it

libretroshare/src/pqi/authssl.cc
```diff
@@ -1198,6 +1198,21 @@ int AuthSSLimpl::VerifyX509Callback(int /*preverify_ok*/, X509_STORE_CTX* ctx)
...
+    {
+        std::cerr << "Nothing known about peer " << sslId << " trying to connect! Refusing connection." << std::endl;
+        return verificationFailed;
+    }
+
+	bool isSslOnlyFriend = det.skip_pgp_signature_validation;
+
+    if(det.gpg_id != pgpId)
```
G10h4ck: This extra check deny connection to a new location of a PGP friend, IMHO we should remove it. PGP-id matching is correctly done at line 1275.

G10h4ck (changes requested):

libretroshare/src/pqi/authssl.cc
csoler: not at all, because it should have the same PGP id.

G10h4ck: the problem is that `det` is not populated if sslid is not known

libretroshare/src/pqi/authssl.cc
```diff
@@ -1127,11 +1127,12 @@ bool AuthSSLimpl::AuthX509WithGPG(X509 *x509, uint32_t& diagnostic)
 			goto err;
 		}
 
-		RsInfo() << __PRETTY_FUNCTION__ << " Verified: " << sigtypestring
-		         << " signature of certificate sslId: "
-		         << RsX509Cert::getCertSslId(*x509)
-		         << ", Version " << std::hex << certificate_version << std::dec
-		         << " using PGP key " << pd.fpr << " " << pd.name << std::endl;
+        if(verbose)
+			std::cerr<< " Verified: " << sigtypestring
```
G10h4ck: Please keep `RsInfo() << __PRETTY_FUNCTION__ ` instead of `std::cerr`

G10h4ck (changes requested):

libretroshare/src/pqi/authssl.cc
csoler: Well, here I removed it because it prints too much. We don't need the time nor the function name.

G10h4ck: If you don'T want the function name just keep `RsInfo()` at least it will print well on all platforms, plain `std::cXXX` doesn't work on Android so I had to do a dirty trick to be able to get those messages too, but it doesn'T wok always, while if you use the thins defined in `util/rsdebug.h` it works well on all platforms

libretroshare/src/pqi/authssl.cc
```diff
@@ -1233,8 +1243,38 @@ int AuthSSLimpl::VerifyX509Callback(int /*preverify_ok*/, X509_STORE_CTX* ctx)
...
 
+	bool isSslOnlyFriend = false;
+
+    // For SSL only friends (ones added through short invites) we check that the fingerprint
+    // in the key (det.gpg_id) matches the one of the handshake.
+	{
+		RsPeerDetails det;
+
+		if(rsPeers->getPeerDetails(sslId,det))
+			bool isSslOnlyFriend = det.skip_pgp_signature_validation;
```
G10h4ck: `bool` must be removed otherwise we are redefining and thus shadowing `isSslOnlyFriend` in that scope and not assigning the variable defined at line 1246

G10h4ck (changes requested):

libretroshare/src/pqi/authssl.cc
csoler: thx.

defnax:
> I have explained the details on a way to implement that in #638
> But I think that proposal (only UI work) should be eventually implemented in another PR after this one is merged.

that would be great

defnax:
hi gui part need/must look now more simple
we can remove the big cert view from Home ?
best use Retroshare ID or Short ID? 

QR Code comes too`? maybe a button to view the QR image

### Did you received a Retroshare ID from friend? Add friend (Button)

### **This is your RetroShare ID. Copy and share with your friends!**

look on screenshot on jami looks very simple
image

unsenep2pdev:
You can try out our pre-release Unseen Communicator based on RS: https://communicator.unseen.is/ <https://communicator.unseen.is/>

csoler:
@defnax : I agree to simplify the Home panel as much as possible. The example you give here is very nice. However, it's important to keep the possibility to use long certs for backward compatibility, at least for a few months.

G10h4ck:
Also please keep all that GUI work into another PR, so we can merge this ASAP without reviewing the code again and again

csoler:
@G10h4ck: of course.

defnax:
> You can try out our pre-release Unseen Communicator based on RS: https://communicator.unseen.is/ <https://communicator.unseen.is/>

Hi, i always posted your fork on RetroShare Forums, you need to look there, i know since last year about Unseen :)
@unsenep2pdev  You are not active on RetroShare network?

PR #1574, opened 2019-05-29
Fix sanitize error

PhenomRetroShare:


csoler:
Ready to merge, IMO. I'm surprised that the group data can be empty. Maybe for some services.

libretroshare/src/services/p3banlist.cc
```diff
@@ -615,9 +615,10 @@ void p3BanList::getDhtInfo()
 
         addBanEntry(ownId, ad, RSBANLIST_ORIGIN_SELF, int_reason, time_stamp);
```
csoler, line 617: That's not the error.
The problem is that 366 calls mBanRanges.find() without mutex protection. Same for the calls afterwards.
So The fix is to add a mutex call in the beginning of the function isAddressAccepted(), given that it doesn't call any other functions of the service. 
And for the record, adding {} here doesn't change anything.

libretroshare/src/gxs/gxssecurity.cc
```diff
@@ -1110,19 +1110,28 @@ bool GxsSecurity::validateNxsGrp(const RsNxsGrp& grp, const RsTlvKeySignature& s
 		grpMeta.serialise(metaData, metaDataLen,api_versions_to_check[i]);
 
 		// copy msg data and meta in allmsgData buffer
-		memcpy(allGrpData, grp.grp.bin_data, grp.grp.bin_len);
-		memcpy(allGrpData+(grp.grp.bin_len), metaData, metaDataLen);
+		if(grp.grp.bin_data && grp.grp.bin_len>0)
```
csoler, line 1113: This change basically means that if the group has no data you do not verify the signature. That is just wrong. You need to put the if only around the memcpy of the grp.grp.bin_data.

csoler (changes requested):
Nice catches, but some of the fixes are wrong/dangerous. Please fix them according to comments.

libretroshare/src/gxs/gxssecurity.cc
```diff
@@ -1112,26 +1112,20 @@ bool GxsSecurity::validateNxsGrp(const RsNxsGrp& grp, const RsTlvKeySignature& s
 		// copy msg data and meta in allmsgData buffer
 		if(grp.grp.bin_data && grp.grp.bin_len>0)
 		{
```
csoler: No! The metadata need *always* to be copied. As I told you: only the grp.grp.bin_data.

PhenomRetroShare: Done.

PhenomRetroShare:
Any News?

csoler:
These "fixes" are generally hazardous, painful to review and have a rather low value. So yes, they are low priority. I will try to flush them when I'm back online.

PR #1576, opened 2019-06-03
attempt to use a cache for user-defined avatars. Should reduce memory…

csoler:
… footprint. To be tested


csoler:
I'm not 100% sure that the cache system is useful because Qt also has its own cache system. The question to solve is whether Qt figures out that two pixmaps allocated independently that contain the same image should be shared or not. 
The correct setting of a pixmap size depending on where it's needed is actually a significant win.

PR #1586, opened 2019-06-20
Implement deep indexing for files through Xapian

G10h4ck:
ATM it there are few indexers implemented which supports extracting most metadata from Ogg and FLAC files, while a more general indexer supports indexing of and common tags among all formats supported by TagLib.
The system has been designed to be easly extensible to more file formats
  registering more indexer functions which just need to extract metadata
  from a certain type of file and feed it to Xapian.
The system has been integrated into existent file search system to
  through generric search requests and results, it keep a good level of
  retro-compatibility due to some tricks.
The indexing system is released under AGPLv3  so when libretroshare is compiled
  with deep search enabled AGPLv3 must be onored instead of LGPLv3-or-later.
Cleaned up the debian copyright file using non-deprecated license
  code-names.


G10h4ck:
All tests seems to fail after compilation which seems to b successful, with this error `8:18 failed 'stat builddir/*/share/app-info/xmls/cc.retroshare.retroshare-gui.xml.gz'` any clue of what/how that file should be generated?

PR #1607, opened 2019-07-29
Fix CMark compilation

PhenomRetroShare:
Remove cmark_node_mem as cmark_node_free do the job.
Look at man:
https://github.com/commonmark/cmark/blob/master/man/man3/cmark.3#L141

### DIFF
--- a/retroshare-gui/src/retroshare-gui.pro
+++ b/retroshare-gui/src/retroshare-gui.pro
@@ -43,6 +43,46 @@ rs_jsonapi {
     FORMS *= gui/settings/JsonApiPage.ui
 }
 
+# Auto detect installed version of cmark
+rs_gui_cmark {
+	DEFINES *= USE_CMARK
+	no_rs_cross_compiling {
+		message("Using compiled cmark")
+		CMARK_SRC_PATH=$$clean_path($${RS_SRC_PATH}/supportlibs/cmark)
+		CMARK_BUILD_PATH=$$clean_path($${RS_BUILD_PATH}/supportlibs/cmark/build)
+		INCLUDEPATH *= $$clean_path($${RS_SRC_PATH}/src/)
+		DEPENDPATH *= $$clean_path($${RS_SRC_PATH}/src/)
+		QMAKE_LIBDIR *= $$clean_path($${CMARK_BUILD_PATH}/)
+		# Using sLibs would fail as libcmark.a is generated at compile-time
+		LIBS *= -L$$clean_path($${CMARK_BUILD_PATH}/src/) -lcmark
+
+		DUMMYCMARKINPUT = FORCE
+		CMAKE_GENERATOR_OVERRIDE=""
+		win32-g++:CMAKE_GENERATOR_OVERRIDE="-G \"MSYS Makefiles\""
+		gencmarklib.name = Generating libcmark.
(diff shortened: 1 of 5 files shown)


retroshare-gui/src/retroshare-gui.pro
```diff
@@ -43,6 +43,23 @@ rs_jsonapi {
...
+# Auto detect installed version of cmark
+cmark {
+	CMARK_AVAILABLE = $$system(pkg-config --exists libcmark && echo yes)
+	isEmpty(CMARK_AVAILABLE) {
+		exists("gui/../../../supportlibs/cmark/src/") {
+			message("Using built-in cmark")
+			CONFIG += cmark_builtin
+			INCLUDEPATH += "gui/../../../supportlibs/cmark/src/"
+		} else {
+			error("CMark not found. Run `git submodule update --init` in trunk folder or install it.")
```
G10h4ck: better to do this automatically like it is done for restbed

retroshare-gui/src/retroshare-gui.pro
```diff
@@ -1391,42 +1408,43 @@ gxsgui {
...
+		  ../../supportlibs/cmark/src/iterator.c								 \
+		  ../../supportlibs/cmark/src/latex.c										 \
+		  ../../supportlibs/cmark/src/man.c											 \
+		  ../../supportlibs/cmark/src/node.c										 \
+		  ../../supportlibs/cmark/src/references.c							 \
+		  ../../supportlibs/cmark/src/render.c									 \
+		  ../../supportlibs/cmark/src/scanners.c								 \
+		  ../../supportlibs/cmark/src/utf8.c										 \
+		  ../../supportlibs/cmark/src/xml.c											 \
+	}
```
G10h4ck: I think this should be avoided, seems cmark uses cmake as build system, so it can be built with it's own build system as we do for restbed

retroshare.pri
```diff
@@ -38,8 +38,8 @@ no_retroshare_nogui:CONFIG -= retroshare_nogui
 
 # To disable cmark append the following 
 # assignation to qmake command line "CONFIG+=no_cmark"
-CONFIG *= no_cmark
-cmark:CONFIG -= no_cmark
+CONFIG *= cmark
+no_cmark:CONFIG -= cmark
```
G10h4ck: If it is a retroshare-gui specific stuff the config name should reflect it for example rs_gui_cmark

G10h4ck (changes requested):

retroshare-gui/src/retroshare-gui.pro
PhenomRetroShare: Like you've done here?
https://github.com/RetroShare/RetroShare/blob/master/libretroshare/src/libretroshare.pro#L839
But this is done on every case, even if library is installed by system?

retroshare.pri
PhenomRetroShare: OK

retroshare-gui/src/retroshare-gui.pro
G10h4ck: yeah like that, but in your case do it only when cmark is not installed in the system

PhenomRetroShare:
Done

G10h4ck:
Looks good to me!

G10h4ck (approved):

PR #1609, opened 2019-08-01
Implement GXS forum reasonable JSON API

G10h4ck:
Remove misleading comments around, and add some TODO of things we should
  fix at some point

### DIFF
--- a/libretroshare/src/retroshare/rsgxsifacetypes.h
+++ b/libretroshare/src/retroshare/rsgxsifacetypes.h
@@ -140,14 +140,20 @@ struct RsMsgMetaData : RsSerializable
     std::string mMsgName;
     rstime_t      mPublishTs;
 
-    /// the lower 16 bits for service, upper 16 bits for GXS
-    uint32_t    mMsgFlags;
+	/** the lower 16 bits for service, upper 16 bits for GXS
+	 * @todo mixing service level flags and GXS level flag into the same member
+	 * is prone to confusion, use separated members for those things, this could
+	 * be done without breaking network retro-compatibility */
+	uint32_t mMsgFlags;
 
     // BELOW HERE IS LOCAL DATA, THAT IS NOT FROM MSG.
     // normally READ / UNREAD flags. LOCAL Data.
 
-    /// the first 16 bits for service, last 16 for GXS
-    uint32_t    mMsgStatus;
+	/** the first 16 bits for service, last 16 for GXS
+	 * @todo mixing service level flags and GXS level flag into the same member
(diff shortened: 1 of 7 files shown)


libretroshare/src/retroshare/rsgxsifacetypes.h
```diff
@@ -140,14 +140,20 @@ struct RsMsgMetaData : RsSerializable
...
 
     // BELOW HERE IS LOCAL DATA, THAT IS NOT FROM MSG.
     // normally READ / UNREAD flags. LOCAL Data.
 
-    /// the first 16 bits for service, last 16 for GXS
-    uint32_t    mMsgStatus;
+	/** the first 16 bits for service, last 16 for GXS
+	 * @todo mixing service level flags and GXS level flag into the same member
+	 * is prone to confusion, use separated members for those things, this could
+	 * be done without breaking network retro-compatibility */
```
csoler, line 155: It's possible but complicated since the DB expects the flags as a single uint32_t. So you'd need to pack them together when storing and unpack the when reading, but then you cannot ensure that flags will be kept in the allowed range when developing new services.

libretroshare/src/services/p3gxsforums.cc
```diff
@@ -844,3 +1005,6 @@ bool RsGxsForumGroup::canEditPosts(const RsGxsId& id) const
 	        id == mMeta.mAuthorId;
 }
 
+RsGxsForumGroup::~RsGxsForumGroup() = default;
```
csoler, line 1008: That's useless IMHO: just add {} after the delcaration in the .h, which wouldn't let the reader think that there's something so special in the declared destructor for it to be postponed to the .cpp

libretroshare/src/retroshare/rsgxsifacetypes.h
G10h4ck: In that case when we change this at some point, we should change it in the database too, and probably write some migration code, to migrate from old DB format to the new one

libretroshare/src/services/p3gxsforums.cc
G10h4ck: It is to avoid the compiler complaining about the `vtable` of that class being emitted in every compilation unit, also `default` and `{}` are treated different, because of C++ specification if the costructor/destructor is omitted or declared default the `vtable` lookup is not necessary improving performances, while even if the method is defined `{}` the compiler won't apply that optimization. So the rationale behind this line that I would prefer not to have to write too, is the less costly thing we can do to avoid the `vtable` being emitted in every compilation unit.

PR #1623, opened 2019-08-28
Protect p3BanList::isAddressAccepted with mutex

G10h4ck:
This method is called from other threads and apparently caused a sporadic crash caught on Android.
See retroshare://forum?name=Got%20crash%20on%20Android%20with%20GDB%20backtrace&id=95de1451952d8c38cb1cdfdb85eed986&msgid=ac8c9d41f2cd0c9e8e290433c7f296fecb2d62b3


csoler:
Seems right to me.

PR #1624, opened 2019-08-29
V0.6 retroshare service

csoler:
This PR is ready:

- allows retroshare-service to be started from the console and possibly handling user selection and PGP passphrase callback in the console (works in this PR)
- launch the webUI possibly asking for a password
- cleans up the mess with commandline options. 
- disables retroshare-nogui and libresapi, now replaced by retroshare-service and jsonapi
- make retroshare-service compile on linux+mac+win


libretroshare/src/jsonapi/jsonapi.h
```diff
@@ -52,6 +52,8 @@ extern JsonApiServer* jsonApiServer;
  */
 struct JsonApiServer : RsSingleJobThread, p3Config
 {
+	static const uint16_t DEFAULT_PORT = 9092 ;
```
G10h4ck, line 55: if you want to put it like that, use `constexpr` instead of `const` ;)

libretroshare/src/rsserver/rsinit.cc
```diff
@@ -22,6 +22,8 @@
 /// RetroShare initialization and login API implementation
 
 #include <unistd.h>
+#include <math.h>
```
G10h4ck: Should use `<cmath>` instead of `<math.h>`

retroshare-service/src/retroshare-service.cc
```diff
@@ -17,66 +17,166 @@
...
 	            [](){
 		if(RsControl::instance()->isReady())
 			RsControl::instance()->rsGlobalShutDown(); } );
 
 	return app.exec();
+#else
+    if(jsonApiServer)
+		jsonApiServer->authorizeToken("webui:"+webui_pass1);
+
+	while(true)
```
G10h4ck: How do you get out of this? You need `rsControl->setShutdownCallback` also on other systems not only in android, you can use an `std::atomic<bool> keepRunning` captured by reference in the lambada passed as callback that get set to false in the callback

retroshare-service/src/retroshare-service.cc
```diff
@@ -17,66 +17,166 @@
...
+	   sleep(1);
```
G10h4ck: use `std::this_thread::sleep_for(std::chrono::seconds(5));` instead

G10h4ck (changes requested):
Isn't easier using `RsLoginHelper::attemptLogin` instead of fighting with the deprecated RsNotify system?

csoler:
> Isn't easier using `RsLoginHelper::attemptLogin` instead of fighting with the deprecated RsNotify system?

yes, you're perfectly right. Still, we need to install a notify component so as to handle the PGP passphrase. But the the rest of it, you're right. This is going to simplify the code.

retroshare-service/src/retroshare-service.cc
csoler: this is still pending. I'll follow your advice.

retroshare-service/src/retroshare-service.cc
```diff
@@ -17,66 +17,274 @@
...
+
+    ch = getchar();
+
+    tcsetattr(STDIN_FILENO, TCSANOW, &t_old);
+    return ch;
+}
+
+std::string readStringFromKeyboard(const char *prompt, bool show_asterisk=true)
+{
+  const char BACKSPACE=127;
```
G10h4ck: use `constexpr` here too

G10h4ck (changes requested):

libretroshare/src/retroshare/rsinit.h
```diff
@@ -45,6 +46,38 @@ struct RsLoginHelper;
...
+    // required
+
+	std::string main_executable_path;/* this should be set to argv[0] */
+
+    // Optional. Only change if needed.
+
+	bool autoLogin;                  /* try auto-login */
+
+	bool udpListenerOnly;			 /* only listen to udp */
+    std::string forcedInetAddress; 	 /* inet address to use.*/
```
sehraf, line 67: How do you manage to screw up indention so bad? :laughing:

retroshare-service/src/retroshare-service.cc
csoler: done

retroshare-service/src/retroshare-service.cc
csoler: done

libretroshare/src/rsserver/rsinit.cc
csoler: done

libretroshare/src/jsonapi/jsonapi.h
csoler: done

retroshare-service/src/retroshare-service.cc
csoler: constexpr is actually not needed for integral expressions. See https://stackoverflow.com/questions/14116003/difference-between-constexpr-and-const

PR #1625, opened 2019-08-30
fixed connection for "deny friend" function in FriendList

csoler:
trying to fix all bugs reported about the new friendlist.


csoler:
This should hopefully solve the crashes. I also removed the regular 5-secs update of the model (which I'm not sure is needed). If you see any discrepancy between what's displayed and who's actully connected, tell me. This may happen for instance when coming out of suspend, etc.

PR #1630, opened 2019-09-04
Fix ChatWidget History UserName

PhenomRetroShare:
Wait some time (maxi 2sec) that Id cache is loaded.

### DIFF
--- a/retroshare-gui/src/gui/chat/ChatWidget.cpp
+++ b/retroshare-gui/src/gui/chat/ChatWidget.cpp
@@ -379,31 +379,42 @@ void ChatWidget::init(const ChatId &chat_id, const QString &title)
 
 		if (messageCount > 0)
 		{
-            rsHistory->getMessages(chatId, historyMsgs, messageCount);
+			rsHistory->getMessages(chatId, historyMsgs, messageCount);
 
 			std::list<HistoryMsg>::iterator historyIt;
 			for (historyIt = historyMsgs.begin(); historyIt != historyMsgs.end(); ++historyIt)
-            {
-                // it can happen that a message is first added to the message history
-                // and later the gui receives the message through notify
-                // avoid this by not adding history entries if their age is < 2secs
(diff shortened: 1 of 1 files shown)


retroshare-gui/src/gui/chat/ChatWidget.cpp
```diff
@@ -379,31 +379,42 @@ void ChatWidget::init(const ChatId &chat_id, const QString &title)
...
-                    else
-                        name = QString::fromUtf8(historyIt->peerName.c_str());
-                } else {
-                    name = QString::fromUtf8(historyIt->peerName.c_str());
-                }
+			{
+				// it can happen that a message is first added to the message history
+				// and later the gui receives the message through notify
+				// avoid this by not adding history entries if their age is < 2secs
+				if ((time(nullptr)-2) <= historyIt->recvTime)
```
csoler: time(NULL) - 2 is a bit dangerous because you're substracting to an unsigned int, so it will be converted to int. Generally, prefer    if(time(nullptr) <= 2+historyIt->recvTime)

PhenomRetroShare: Done, even if not mine. ;)

ef5fca5e5e (AsamK                        2015-08-07 19:08:10 +0200)                 if ((time(NULL)-2) <= historyIt->recvTime)

defnax:
soon will be checked?

i often see this here:
image

csoler:
Seems to me that it would be even better to implement directly a synchronous call to getIdDetails. For instance:
- add an optional param to getIdDetails(); bool wait_cache_load=false
- in p3idservice.cc:698, before the if, add a if(force_cache_load) while(!mKeyCache.fetch(id)) cache_request_load(id);
I'm not 100% it would work though. To be tested.

defnax:
at the moment to get usernames visible , you must leave chat and join again.

PR #1638, opened 2019-09-14
Provide more usable method to create GXS circles

G10h4ck:
Provide more usable method to create GXS circles
Fix some compiler warnings with out of definition default destructors

### DIFF
--- a/libretroshare/src/retroshare/rsgxscircles.h
+++ b/libretroshare/src/retroshare/rsgxscircles.h
@@ -3,8 +3,8 @@
  *                                                                             *
  * libretroshare: retroshare core library                                      *
  *                                                                             *
- * Copyright (C) 2012  Robert Fernie <retroshare@lunamutt.com>                 *
- * Copyright (C) 2018  Gioacchino Mazzurco <gio@eigenlab.org>                  *
+ * Copyright (C) 2012-2014  Robert Fernie <retroshare@lunamutt.com>            *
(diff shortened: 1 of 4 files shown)


libretroshare/src/retroshare/rsgxscircles.h
```diff
@@ -95,17 +94,17 @@ struct RsGxsCircleGroup : RsSerializable
 		RS_SERIAL_PROCESS(mInvitedMembers);
 		RS_SERIAL_PROCESS(mSubCircles);
 	}
+
+	~RsGxsCircleGroup() override;
```
csoler, line 98: Why move this to the .cc ?? When I read the .h it lets me think that something special is done in the destructor whereas it's not.

csoler (commented):
Can the current GUI still create circles? It looks like you removed the createCircle(GxsGroupMeta&) that it was using.

libretroshare/src/retroshare/rsgxscircles.h
G10h4ck: It is to avoid the compiler warning I talked to you some time ago, about the vtable being emitted in every compilation unit when no method is defined outside the .h where the class is declared, moving the default destructor outside is the less invasive thing we can do to avoid that

G10h4ck:
> Can the current GUI still create circles? It looks like you removed the createCircle(GxsGroupMeta&) that it was using.

yes no one used that the GUI call `createGroup` directly

csoler:
ok, let's merge this then!

PR #1640, opened 2019-09-16
Provide usable API to send RS mails

G10h4ck:
Return trackid to sendmail caller
Post RsEvents when sent mail change status
Deprecate old MessageSend

### DIFF
--- a/libretroshare/src/retroshare/rsmsgs.h
+++ b/libretroshare/src/retroshare/rsmsgs.h
@@ -161,34 +161,6 @@ class MsgAddress
 		std::string _addr_string ;
 };
 
-class MessageInfo_v2
-{
-	public:
-		//MessageInfo_v2() {}
-
-		unsigned int msgflags;
-
-		//RsMessageId msgId;
-		MsgAddress from ;
-
-		std::list<MsgAddress> rcpt ;
-
-		// Headers
-		//
-		std::string subject;
-		std::string msg;
-		rstime_t time_stamp ;
-
-		//std::list<MessageHeader> headers ;
-
-		std::string attach_title;
-		std::string attach_comment;
-		std::list<FileInfo> files;
-
-		int size;  /* total of files */
-		int count; /* file count     */
-};
-
 struct MessageInfo : RsSerializable
 {
 	MessageInfo(): msgflags(0), size(0), count(0), ts(0) {}
--- a/libretroshare/src/retroshare/rsmsgs.h
+++ b/libretroshare/src/retroshare/rsmsgs.h
@@ -236,19 +236,22 @@ struct MsgInfoSummary : RsSerializable
 {
 	MsgInfoSummary() : msgflags(0), count(0), ts(0) {}
 
-    RsMailMessageId msgId;
+	RsMailMessageId msgId;
 	RsPeerId srcId;
 
 	uint32_t msgflags;
-    std::list<uint32_t> msgtags;	// that leaves 25 bits for user-defined tags.
+	std::list<uint32_t> msgtags; /// that leaves 25 bits for user-defined tags.
 
 	std::string title;
-	int count; /* file count     */
+	int count; /** file count     */
 	rstime_t ts;
 
 
-	// RsSerializable interface
-	void serial_process(RsGenericSerializer::SerializeJob j, RsGenericSerializer::SerializeContext &ctx) {
+	/// @see RsSerializable
+	void serial_process(
+	        RsGenericSerializer::SerializeJob j,
+	        RsGenericSerializer::SerializeContext &ctx) override
+	{
 		RS_SERIAL_PROCESS(msgId);
 		RS_SERIAL_PROCESS(srcId);
 
--- a/libretroshare/src/retroshare/rsmsgs.h
+++ b/libretroshare/src/retroshare/rsmsgs.h
@@ -100,11 +100,11 @@ typedef std::string RsMailMessageId; // TODO: rebase on t_RsGenericIdType
 
 /**
  * Used to return a tracker id so the API user can keep track of sent mail
- * status
+ * status, it contains mail id, and recipient id
  */
-struct RsMailTrackId : RsSerializable
+struct RsMailIdRecipientIdPair : RsSerializable
 {
-	RsMailTrackId(RsMailMessageId mailId, RsGxsId recipientId):
+	RsMailIdRecipientIdPair(RsMailMessageId mailId, RsGxsId recipientId):
 	    mMailId(mailId), mRecipientId(recipientId) {}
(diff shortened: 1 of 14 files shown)


libretroshare/src/retroshare/rsmsgs.h
```diff
@@ -93,9 +95,31 @@ const ChatLobbyFlags RS_CHAT_LOBBY_FLAGS_PGP_SIGNED    ( 0x00000010 ) ; // requi
...
+ * Used to return a tracker id so the API user can keep track of sent mail
+ * status
+ */
+struct RsMailTrackId : RsSerializable
+{
+	RsMailTrackId(RsMailMessageId mailId, RsGxsId recipientId):
+	    mMailId(mailId), mRecipientId(recipientId) {}
+
+	RsMailMessageId mMailId;
+	RsGxsId mRecipientId;
```
csoler, line 111: seems that only mMailId is needed, since the rest of information (msg content, flags, and recipitent) should be available by using getMessage. It's inconsistent to add the recipient *inside* the ID structure.

libretroshare/src/retroshare/rsmsgs.h
```diff
@@ -93,9 +95,31 @@ const ChatLobbyFlags RS_CHAT_LOBBY_FLAGS_PGP_SIGNED    ( 0x00000010 ) ; // requi
...
 typedef std::string ChatLobbyNickName ;
-typedef std::string RsMailMessageId;			// should be uint32_t !!
-typedef uint64_t    MessageId ;
+typedef std::string RsMailMessageId; // TODO: rebase on t_RsGenericIdType
 
+/**
+ * Used to return a tracker id so the API user can keep track of sent mail
+ * status
+ */
+struct RsMailTrackId : RsSerializable
```
csoler: Given the next comment, I suppose that it would be even simpler to define that ID as a t_RSGenericId<some const>. You'd get serialization, comparisons, and type checking for free.

libretroshare/src/services/p3msgservice.cc
```diff
@@ -1192,14 +1211,125 @@ bool 	p3MsgService::MessageSend(MessageInfo &info)
...
+			}
+
+			const RsMailMessageId mailId = std::to_string(msgId);
+			pEvent->mChangedMsgIds.insert(mailId);
+			trackingIds.insert(RsMailTrackId(mailId, dst));
+			++ret;
+		}
+	};
+
+	pSend(to);
```
csoler, line 1314: It's surprising that the msg structure is not modified when you send in bcc. Normally the list of recipients should be cleared right?

libretroshare/src/services/p3msgservice.cc
```diff
@@ -1954,17 +2086,31 @@ void p3MsgService::notifyDataStatus( const GRouterMsgPropagationId& id,
...
+		it2->second->msgFlags &= ~RS_MSG_FLAGS_PENDING;
+		imsg[msg_id] = it2->second;
+		msgOutgoing.erase(it2);
+#endif
 
 		RsServer::notify()->notifyListChange( NOTIFY_LIST_MESSAGELIST,
 		                                      NOTIFY_TYPE_ADD );
 		IndicateConfigChanged();
 
+		using Evt_t = RsMailStatusEvent;
```
csoler, line 2092: I'm wondering if there is a way to put these 4 lines into one, maybe using a method like sendEvent(RsMailStatusEvent(...));.

libretroshare/src/services/p3msgservice.cc
```diff
@@ -2007,29 +2153,31 @@ bool p3MsgService::receiveGxsTransMail( const RsGxsId& authorId,
...
-			std::cerr << __PRETTY_FUNCTION__ << " (II) receiving "
+			RsInfo() << __PRETTY_FUNCTION__ << " (II) receiving "
 			          << "message of hash " << hash << " more than once. "
 			          << "Probably it has arrived  before by other means."
 			          << std::endl;
 			return true;
 		}
-		mRecentlyReceivedMessageHashes[hash] = time(NULL);
+		mRecentlyReceivedMessageHashes[hash] =
+		        static_cast<uint32_t>(time(nullptr));
```
csoler, line 2152: seems to me we switched to rs_time() to avoid the need to cast.

libretroshare/src/retroshare/rsmsgs.h
G10h4ck: The problem is that when you use getMail to check the status of a mail you sent, you get the mail with all the recipients, you have no way to check to which specific recipient that status apply. Instead if sendMail return both mail id and the specific recipient for each single mail sent the caller can keep track of the status for each single recipient, because she knows each mail id to which specific sender has been sent.

libretroshare/src/retroshare/rsmsgs.h
G10h4ck: see my reply upster

libretroshare/src/services/p3msgservice.cc
G10h4ck: I expected that too, but didn't see that in the old code, maybe retroshare-gui does that?

libretroshare/src/services/p3msgservice.cc
G10h4ck: A macro could be used for this, but I would keep that for another PR

libretroshare/src/services/p3msgservice.cc
G10h4ck: The problem is `std::map<Sha1CheckSum, uint32_t> mRecentlyReceivedMessageHashes;` are you suggesting we should change that uint32_t for an rstime_t ?
I would be happy to do that if you are sure that will not break retrocompatibility, I am quite sure that map is not sent over network, do you know if it is saved to disk?

libretroshare/src/retroshare/rsmsgs.h
csoler: Then in the sendMail function, you may return a std::map<MailId, Recipient ID>. It feels a bit inconsistent to mix the two in the Mail ID structure, especially if you only need this in one particular place in the code.

G10h4ck: be aware that It is not Mail id structure ;), mail id structure remained the same RsMailTrackId is a new structure I used just for the purpose. right now I don't remember why I didn't just used a pair of the two, probably to have the JSON nicer to read :p

libretroshare/src/services/p3msgservice.cc
G10h4ck: I have double checked the code and the BCC filtering is done at lines 1898-1904

PR #1645, opened 2019-09-27
Allow friend information  update from short invite

G10h4ck:


libretroshare/src/pqi/p3peermgr.cc
```diff
@@ -1071,42 +1071,53 @@ bool p3PeerMgrIMPL::addFriend(const RsPeerId& input_id, const RsPgpId& input_gpg
...
+	if(sslId == getOwnId())
+	{
+		RsErr() << __PRETTY_FUNCTION__ << " Cannot add yourself as SSL-only "
+		        << "friend (id=" << sslId << ")" << std::endl;
+		   return false;
+	}
+
+	bool alreadyFriend = false;
 	peerState pstate;
+	pstate.skip_pgp_signature_validation = true;
```
csoler: Moving that here seems wrong to me: because the flag will be set even if the short invite received is wrong because it has a known (friend) SSL id with the wrong PGP id.

libretroshare/src/pqi/p3peermgr.cc
```diff
@@ -1119,13 +1130,9 @@ bool p3PeerMgrIMPL::addSslOnlyFriend( const RsPeerId& sslId, const RsPgpId& pgp_
 	if(dt.hiddenType)        pstate.hiddenType = dt.hiddenType;
 	if(!dt.location.empty()) pstate.location = dt.location;
 
-    pstate.skip_pgp_signature_validation = true;
```
csoler, line 1122: This is the place where it should be: after all checks.

libretroshare/src/pqi/p3peermgr.cc
G10h4ck: In that case `pstate` is completely discarded, while if we already have it the `pstate` and thus `pstate.skip_pgp_signature_validation` is overwritten with the one we have already. That flag would apply only in case is a completely new ssl-only friend

libretroshare/src/pqi/p3peermgr.cc
G10h4ck: see my explaination upster ;)

csoler: I'm thinking in terms of code security. This flag is a bomb: if some bug or hack flips it, the software's security is down. That's why it's preferable to set it after all tests are done, so that any further change in the code is less likely to unintentionally cause such a problem. Maybe you can add that comment before setting the flag.

G10h4ck: Ok I agree in commenting this on the code

libretroshare/src/pqi/p3peermgr.cc
csoler: also, the way it is now, you can re-add with short invite some friend you already have and it will reset his PGP signature validation. This is not the expected behavior! Really, you dont set such a flag to true *unless* all checks have been passed.

G10h4ck: no it will not, look at line 1118 `pstate` and thus `pstate.skip_pgp_signature_validation` get overriden with the already known state if it is already friend. So if it has already PGP `pstate.skip_pgp_signature_validation` will be reset to `false`, if not it will remain `true`

sehraf: what is the motivation behind having it up here?

csoler: my point is: with such a dangerous flag, we need to be as clear as possible. Moving that =true up here requires lots of thinking in order to make sure that it does the correct thing in all cases. Whatever you gain with this is not worth it IMHO.

G10h4ck: Wait I found another risky and quircky case, I am writing the code to handle that too

G10h4ck: > what is the motivation behind having it up here?

The motivation was so it get overridden by already known value if it was an already known location, but now it is irrelevant I have just pushed a safer version ;)

I really appreciate @csoler security insight

G10h4ck:
Please take a look now

libretroshare/src/pqi/p3peermgr.cc
```diff
@@ -1071,45 +1071,81 @@ bool p3PeerMgrIMPL::addFriend(const RsPeerId& input_id, const RsPgpId& input_gpg
...
+	 * PGP id of an already known friend but the SSL-id of a location generated
+	 * by the attacker which doesn't have access to the legitimate PGP
+	 * certificate.
+	 * In that case being pstate.skip_pgp_signature_validation false on
+	 * connection attempt the PGP signaure verification would fail and the
+	 * connection closed.
+	 * Instead if pstate.skip_pgp_signature_validation would have been
+	 * superficially set to true the PGP signature verification would have been
+	 * skipped and the attacker connection would be accepted. */
+	if(!AuthGPG::getAuthGPG()->isGPGAccepted(pgp_id))
```
csoler: I think that the test should even be broader: it should be if(!AuthGPG::getauthGPG()->isGpgId()), because whenever you have the PGP public key, you shouldn't need to skp the PGP validation. It's only when the key doesn't exist (which is what isGpgId() is checking---the name is not so good) that you have to use the skip_validation flag

csoler: nice catch btw ;-)

libretroshare/src/pgp/pgphandler.cc
```diff
@@ -1667,6 +1668,9 @@ bool PGPHandler::getGPGFilteredList(std::list<RsPgpId>& list,bool (*filter)(cons
 	return true ;
 }
 
+bool PGPHandler::isPgpPubKeyAvailable(const RsPgpId &id)
+{ return _public_keyring_map.find(id) != _public_keyring_map.end(); }
+
 bool PGPHandler::isGPGId(const RsPgpId &id)
```
csoler, line 1674: you kept the old method?

csoler:
This PR seems ok for merging IMHO.

libretroshare/src/pgp/pgphandler.cc
G10h4ck: Yeah I kept the old method, but marked it as deprecated

PR #1648, opened 2019-09-27
Home view changes for short invites

defnax:
Added Label for display the Retroshare ID (short invite)
Added Button for view optional the certificate
some renaming changes from certificate to retrohare id

image
image
image


felisucoibi:
+1

csoler:
Just a few ideas:
- the button to show the full cert should have tooltip saying "show full certificate (old format)"
- when the full cert it shown I would hide the short one, so that it is clear that one can replace the other.
- the tooltip for the short one should not say "A profile key", since it doesnt contain it, contrarily to the full cert.

defnax:
> Just a few ideas:
>
> * when the full cert it shown I would hide the short one, so that it is clear that one can replace the other.

i dont like this idea, the new Retroshare id must be always visible, on jami/tox they dont hide when the QR code is visble too as expamle.
i added the hide feature for temporary hide the cert as you said in few month we dont need more the cert display there.

defnax:
> * the tooltip for the short one should not say "A profile key", since it doesnt contain it, contrarily to the full cert.

i removed the tooltip from label i dont know what it contains all?
there needs a new class for getIdDescription for Retroshare ID
`ConfCertDialog::getCertificateDescription`

PR #1651, opened 2019-09-28
Posted links logo header

defnax:
Added for Posted header frame Logo & Name
Removed unwanted icon from tree.
This was added on pr https://github.com/RetroShare/RetroShare/pull/1535


csoler:
Lots of users complained that our UI wastes space. In this PR you're adding a bar that is pretty large with some text that actually duplicates the one in the list at left.

PR #1652, opened 2019-09-28
Added Add button for friends list

defnax:
image


csoler:
The export/import is already available in the context menu of the friend list. It's better that functionalities are in a single place. That avoids people wondering if two buttons actually do the same thing or not.
Same for add Friend, it's already in the Home tab. 
@G10h4ck: what's your opinion on this last one?

defnax:
> The export/import is already available in the context menu of the friend list. It's better that functionalities are in a single place. That avoids people wondering if two buttons actually do the same thing or not.
> Same for add Friend, it's already in the Home tab.
> @G10h4ck: what's your opinion on this last one?

yes its hidden nobody see it. same for export pgp profile (only in new node) 
We has no syncing friendslist feature, but i think import friendslist is broken i found today not worked.
its faster when you are on friendslist to hava a add friend there too

other projects has buttons for easy/ import/export on ui and we must hide on context menues?

csoler:
Before I re-implemented this friend list from scratch, it was a mess there: there was menus and buttons in different places to do similar things. So I tried to make it less messy. It makes sense to put the export/import friend list in the context menu since:
- it doesn't need an additional button (which means more space, less visual complexity)
- you basically reach other information about nodes/friends from the context menu anyway

defnax:
> * you basically reach other information about nodes/friends from the context menu anyway

yes i know you are fan of context menus.

it was only one Button, not 3 buttons.

PR #1653, opened 2019-09-28
Removed default stylesheets from titlebar frames

defnax:
I removed the old stylesheets from the title bar frames, from the default Mainwindow dialogs
i think now looks better then before (better look & feel)

image


csoler:
in this PR the resource file that is referred to is ../images.qrc instead of just images.qrc. I think it's a problem. apart from that, it is probably a matter of taste. The previous settings used to show the title bars far better IMHO.

defnax:
> in this PR the resource file that is referred to is ../images.qrc instead of just images.qrc. I think it's a problem. apart from that, it is probably a matter of taste. The previous settings used to show the title bars far better IMHO.

i has thinked we need to say good bye to this title bar with stylesheet.
i can fix that soon with images.qrc was maybe mistake by me

csoler:
ok. If you can make it a bit more visible it would be nice. Now with the PR, the bar disapears on the background.

defnax:
> ok. If you can make it a bit more visible it would be nice. Now with the PR, the bar disapears on the background.

i removed the stylesheets the frame is always there.
i want say its not needed on default toolbar with color, we need make some new things ui.
else nobody doing or create new ui. .

i saw on unseen it looks more nice without color the bar

csoler:
ok. I trust you on that.

PR #1654, opened 2019-09-29
Fix implicitly-declared operator= warnings

PhenomRetroShare:
/libretroshare/src/gxs/rsgds.h:54: warning: implicitly-declared
'ContentValue& ContentValue::operator=(const ContentValue&)' is
deprecated [-Wdeprecated-copy]
   54 |     MsgLocMetaData(const MsgLocMetaData& meta){ msgId =
meta.msgId; val = meta.val;}

/libretroshare/src/gxs/rsgds.h:69: warning: implicitly-declared
'ContentValue& ContentValue::operator=(const ContentValue&)' is
deprecated [-Wdeprecated-copy]
   69 |     GrpLocMetaData(const GrpLocMetaData& meta){ grpId =
meta.grpId; val = meta.val;}

From 248 to 146 warnings.

### DIFF
--- a/libretroshare/src/gxs/rsgds.h
+++ b/libretroshare/src/gxs/rsgds.h
@@ -48,13 +48,12 @@ public:
  * This allows modification of local
  * meta data items of a message
  */
-class MsgLocMetaData {
+struct MsgLocMetaData {
+	MsgLocMetaData() = default;
+	MsgLocMetaData(const MsgLocMetaData& meta): msgId(meta.msgId), val(meta.val) {}
 
-public:
-    MsgLocMetaData(const MsgLocMetaData& meta){ msgId = meta.msgId; val = meta.val;}
-    MsgLocMetaData() {}
-    RsGxsGrpMsgIdPair msgId;
-    ContentValue val;
+	RsGxsGrpMsgIdPair msgId;
+	ContentValue val;
 };
 
 typedef std::map<RsGxsGroupId,RsGxsGrpMetaData*> RsGxsGrpMetaTemporaryMap;
@@ -63,14 +62,12 @@ typedef std::map<RsGxsGroupId,RsGxsGrpMetaData*> RsGxsGrpMetaTemporaryMap;
  * This allows modification of local
  * meta data items of a group
  */
-class GrpLocMetaData {
+struct GrpLocMetaData {
+	GrpLocMetaData() = default;
+	GrpLocMetaData(const GrpLocMetaData& meta): grpId(meta.grpId), val(meta.val) {}
 
-public:
-    GrpLocMetaData(const GrpLocMetaData& meta){ grpId = meta.grpId; val = meta.val;}
-    GrpLocMetaData(){}
-    RsGxsGroupId grpId;
-    ContentValue val;
-
+	RsGxsGroupId grpId;
+	ContentValue val;
 };
 
 /*!


libretroshare/src/gxs/rsgds.h
```diff
@@ -51,8 +51,8 @@ class RsGxsSearchModule  {
 class MsgLocMetaData {
 
 public:
-    MsgLocMetaData(const MsgLocMetaData& meta){ msgId = meta.msgId; val = meta.val;}
-    MsgLocMetaData() {}
+    MsgLocMetaData(const MsgLocMetaData& meta): msgId(meta.msgId), val(meta.val) {}
+    MsgLocMetaData() = default;
```
G10h4ck: Now that you touch it, you should made this class a struct, remove public keyword, and this line too (the class have no "special" members like pointers so the default costructor would be generated anyway AFAIR )

libretroshare/src/gxs/rsgds.h
```diff
@@ -66,8 +66,8 @@ typedef std::map<RsGxsGroupId,RsGxsGrpMetaData*> RsGxsGrpMetaTemporaryMap;
 class GrpLocMetaData {
 
 public:
-    GrpLocMetaData(const GrpLocMetaData& meta){ grpId = meta.grpId; val = meta.val;}
-    GrpLocMetaData(){}
+    GrpLocMetaData(const GrpLocMetaData& meta): grpId(meta.grpId), val(meta.val) {}
+    GrpLocMetaData() = default;
```
G10h4ck: Now that you touch it, you should made this class a struct, remove public keyword, and this line too (the class have no "special" members like pointers so the default constructor would be generated anyway AFAIR )

G10h4ck (changes requested):

libretroshare/src/gxs/rsgds.h
PhenomRetroShare: Done

libretroshare/src/gxs/rsgds.h
```diff
@@ -48,13 +48,12 @@ class RsGxsSearchModule  {
  * This allows modification of local
  * meta data items of a message
  */
-class MsgLocMetaData {
+struct MsgLocMetaData {
+	MsgLocMetaData() {}
```
G10h4ck: What I meant  you should remove this empty constructor, and probably the compiler will generate it automatically. If not please use `MsgLocMetaData() = default;` that tell explicitly to the compiler to generate it.

libretroshare/src/gxs/rsgds.h
```diff
@@ -63,14 +62,12 @@ typedef std::map<RsGxsGroupId,RsGxsGrpMetaData*> RsGxsGrpMetaTemporaryMap;
...
  */
-class GrpLocMetaData {
-
-public:
-    GrpLocMetaData(const GrpLocMetaData& meta){ grpId = meta.grpId; val = meta.val;}
-    GrpLocMetaData(){}
-    RsGxsGroupId grpId;
-    ContentValue val;
+struct GrpLocMetaData {
+	GrpLocMetaData() {}
```
G10h4ck: Same as upster

G10h4ck (changes requested):

libretroshare/src/gxs/rsgds.h
PhenomRetroShare: Done

PR #1659, opened 2019-09-29
Calm down p3netmgr at least when internet connection is not available

G10h4ck:
+ Solve compiler warnings, delete tons of cruft and deadcode


libretroshare/src/pqi/p3netmgr.cc
```diff
@@ -415,47 +408,37 @@ void p3NetMgrIMPL::netStartup()
...
 {
-	rstime_t now = time(NULL);
-	bool doSlowTick = false;
+	rstime_t now = time(nullptr);
+	rstime_t dontCheckNetUntil;
+	{ RS_STACK_MUTEX(mNetMtx); dontCheckNetUntil = mDoNotNetCheckUntilTs; }
+
+	if(now >= dontCheckNetUntil) netStatusTick();
+
+	uint32_t netStatus; { RS_STACK_MUTEX(mNetMtx); netStatus = mNetStatus; }
```
csoler, line 417: I dont think you need the mutex nor that second variable here.

PR #1661, opened 2019-10-01
Fix setting port when adding from short invite with dyndns

G10h4ck:


defnax:
1. its now working short invite with friend request? i has tested i doesnt get any notify.

2. the short invite contains no ip? in my tests i get on short only when i enable "include all your own ip" else the short has no ip included.

PR #1665, opened 2019-10-04
Fix implicitly-declared operator= warnings

PhenomRetroShare:
/retroshare-gui/src/TorControl/HiddenService.cpp:105: warning:
implicitly-declared 'CryptoKey& CryptoKey::operator=(const CryptoKey&)'
is deprecated [-Wdeprecated-copy]

### DIFF
--- a/retroshare-gui/src/TorControl/CryptoKey.cpp
+++ b/retroshare-gui/src/TorControl/CryptoKey.cpp
@@ -51,8 +51,13 @@ void RSA_get0_factors(const RSA *r, const BIGNUM **p, const BIGNUM **q)
 void base32_encode(char *dest, unsigned destlen, const char *src, unsigned srclen);
 bool base32_decode(char *dest, unsigned destlen, const char *src, unsigned srclen);
 
-CryptoKey::CryptoKey()
+CryptoKey& CryptoKey::operator = (const CryptoKey &t)
 {
+	// Check for self assignment
+	if(this != &t)
+		d = t.d;
+
+	return *this;
 }
 
 CryptoKey::~CryptoKey()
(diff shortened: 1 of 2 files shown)


csoler:
I dont understand what you're trying to do here. There's no need for an assigment operator since the default ssigment operator is going to do exactly what you explicitely implemented. You could also remove the copy constructor which does the default as well.
I have the feeling that you got a warning somewhere in the code that uses this class that may concern the way we use it instead of the class itself.

PhenomRetroShare:
It's here: [/retroshare-gui/src/TorControl/HiddenService.cpp:105](https://github.com/RetroShare/RetroShare/blob/502459a25a01e3136c0e8c1b05fb3b8dd72f8f4d/retroshare-gui/src/TorControl/HiddenService.cpp#L105)
```void HiddenService::setPrivateKey(const CryptoKey &key)
{
    if (m_privateKey.isLoaded()) {
        BUG() << "Cannot change the private key on an existing HiddenService";
        return;
    }

    if (!key.isPrivate()) {
        BUG() << "Cannot create a hidden service with a public key";
        return;
    }
//--- Here ---//
    m_privateKey = key;
    m_hostname = m_privateKey.torServiceID() + QStringLiteral(".onion");
    emit privateKeyChanged();
}```
How can I do it better?

PhenomRetroShare:
@csoler : >How can I do it better?

csoler:
> @csoler : >How can I do it better?

Answer is : no need to do it at all. The implicit copy constructor is going to copy d, which is a Shared pointer (managed by Qt). So there's no ambiguity about what happens in the copy operation.

PR #1667, opened 2019-10-05
Some tweaks for posted links

hunbernd:
- Display the full URL instead of just the host
- Jump back to the first page, when changing topic
- Removed inconsistent icon


felisucoibi:
thks

PR #1668, opened 2019-10-05
Some changes for the feed items

defnax:
* Changes for thumbnail display & subscribe buttons
* Display Poster thumbnails without wasted empty space
* Display the thumbnail as Poster, 1:1 or 16:9 format
* Moved the Subscribe buttons to the left
* Changed channel feed icons


felisucoibi:
ole

retroshare-gui/src/gui/feeds/GxsChannelPostItem.cpp
```diff
@@ -392,9 +392,21 @@ void GxsChannelPostItem::fill()
...
+			ui->logoLabel->setMaximumSize(82,108);
+		}
+		else if( thumbnail.width() < 109 ){
+			ui->logoLabel->setMinimumSize(108,108);
+			ui->logoLabel->setMaximumSize(108,108);
+		}
+		else{
+			ui->logoLabel->setMinimumSize(156,108);
+			ui->logoLabel->setMaximumSize(156,108);
+		}
```
csoler: This is likely to fail, because you give an absolute size everywhere. Try to multiply everything by font height/14.0

csoler: ....or compare the size to that of another widget.

csoler:
I like the new buttons. Some ideas:
- the expand button really looks like the "set as read" button. Could you make the "set as read" button green instead of blue?

retroshare-gui/src/gui/feeds/GxsChannelPostItem.cpp
defnax: without fixed not possible, how to display thumbnail as poster?
People using on channels poster format or 16:9 sometime 1:1
at the moment the thumbnail label uses fixed format on ui without my changes too.
i detect it with the width if its Poster else 16:9 (156,108)

when not set fixed then it will look ugly the label

defnax:
> I like the new buttons. Some ideas:
> 
> * the expand button really looks like the "set as read" button. Could you make the "set as read" button green instead of blue?

yes i saw that i thinked maybe make it green
i want change all the feeds icons on the buttons

retroshare-gui/src/gui/feeds/GxsChannelPostItem.cpp
defnax: i can remove that changes, and first change all the icons on this pr

csoler: ok. I'm pretty sure it's possible to do it in a way or another. If you measure how many unites of text height there is, then you can set the size of the image accordingly.

defnax: the problem people attaching different size/dimension of poster pics format, some is 75x108, 80x108, 82x108... that the reason i set minum and maximum then the empty little background will filled with color.
else you see in one channel lots of different sized qlabels 

Best was at posting you set your right format when attaching thumbnail to the media
and set your default fixed thumbnail format for your channel then we dont need fixed size label.

csoler: What Im saying is that, instead of typing 108, you should use 108/14.0*font().height(). This way the size adapts to the screen. This is what we do everywhere else.

defnax: thx i will look on it

defnax:
now done with other icons too, but on some feeds need to replace peer,info, frien request, message icons too... but doesnt have yet the icons for that, when get time in another pr.

retroshare-gui/src/gui/feeds/GxsChannelPostItem.cpp
```diff
@@ -389,12 +389,26 @@ void GxsChannelPostItem::fill()
...
 		// Wiping data - as its been passed to thumbnail.
+		if( thumbnail.width() < 90 ){
+			ui->logoLabel->setMaximumSize(82,108*f);
+		}
+		else if( thumbnail.width() < 109 ){
+			ui->logoLabel->setMinimumSize(108,108*f);
+			ui->logoLabel->setMaximumSize(108,108*f);
+		}
+		else{
+			ui->logoLabel->setMinimumSize(156,108*f);
```
csoler: It seems to me that you need to use f in both left and right, because the image should be scaled both ways. I mean:
setMinimumSize(108*f,108*f)

defnax: ok i will change

defnax: done

PR #1671, opened 2019-10-06
Fixed misleading comments in retroshare.pri

hunbernd:
Fixed misleading comments in retroshare.pri
Fixed no_retroshare_service had no effect

Note: the default state of the switches was not changed


csoler:
thx

PR #1674, opened 2019-10-09
Fix double free crash in ~pqissludp()

G10h4ck:
Use POD type member intitialization for sockaddr_storage
Cleanup a bunch of clutter

### DIFF
--- a/libretroshare/src/pqi/pqissludp.cc
+++ b/libretroshare/src/pqi/pqissludp.cc
@@ -3,7 +3,8 @@
  *                                                                             *
  * libretroshare: retroshare core library                                      *
  *                                                                             *
- * Copyright 2004-2006 by Robert Fernie <retroshare@lunamutt.com>              *
+ * Copyright (C) 2004-2006  Robert Fernie <retroshare@lunamutt.com>            *
+ * Copyright (C) 2015-2019  Gioacchino Mazzurco <gio@altermundi.net>           *
  *                                                                             *
  * This program is free software: you can redistribute it and/or modify        *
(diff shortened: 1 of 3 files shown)


libretroshare/src/pqi/pqissludp.cc
```diff
@@ -50,38 +51,21 @@ static const uint32_t PQI_SSLUDP_DEF_CONN_PERIOD = 300;  /* 5  minutes? */
 /********** PQI SSL UDP STUFF **************************************/
 
 pqissludp::pqissludp(PQInterface *parent, p3LinkMgr *lm) :
-	pqissl(NULL, parent, lm), tou_bio(NULL),// listen_checktime(0),
-	mConnectPeriod(PQI_SSLUDP_DEF_CONN_PERIOD), mConnectFlags(0),
-	mConnectBandwidth(0)
-{
-	RS_STACK_MUTEX(mSslMtx);
-
-	sockaddr_storage_clear(remote_addr);
```
csoler, line 59: I believe that this is used to zero the memory in these structs, that have no initialization otherwise.

libretroshare/src/pqi/pqissludp.cc
```diff
@@ -430,30 +414,6 @@ int pqissludp::net_internal_SSL_set_fd(SSL *ssl, int fd)
 	return 1;
 }
 
-int pqissludp::net_internal_fcntl_nonblock(int /*fd*/)
```
csoler, line 433: There's quite a hierarchy of classes above and below pqissludp. Sometimes you find functions that do nothing (like here) but prevent calling overloaded methods in subclasses. Did you check that?

PR #1675, opened 2019-10-09
Added comment button

defnax:
Added comment button to Comments dialog, which was always missed since long ago
image


felisucoibi:
+1

PR #1680, opened 2019-10-12
enabled buffering for distant chat, so that msgs are sent when the tu…

csoler:
…nnel is ok

### DIFF
--- a/libretroshare/src/chat/p3chatservice.cc
+++ b/libretroshare/src/chat/p3chatservice.cc
@@ -341,7 +341,7 @@ bool p3ChatService::sendChat(ChatId destination, std::string msg)
     message.incoming = false;
     message.online = true;
 
-	if(!isOnline(vpid))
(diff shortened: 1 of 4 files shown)


csoler:
Solves a long term problem with distant chat that was due to the first versions of GxsTunnelService not buffering the out data.

felisucoibi:
wiiiiii

defnax:
found some issues/problems:

importend:
1. You must open the Chat window then i get the offline Chat message, but you dont know more who you wrote
you must look manuelly, seems else you dont get the distant chat.


2. there is no notify like direct chat, that some one wrote you notify on people dialog needed.
3. i doesnt see the text which i wrote in the chat window when i re open the chat window again its empty. 
4. Storing/loading chat history for distant chats seems not works? not sure why but for direct chat and lobbies it works.

normally it must send the offline message without i need reopen the chat window.
the sender or receiver must start again a distant chat to get the offline chat message.
or its bug on my rs?

PR #1687, opened 2019-10-16
Set tooltip for the copy retroshare link button

defnax:
Set tooltip for the copy retroshare link button


defnax:
but i cant ignore retrochess for this pr?, its added to my master

PR #1688, opened 2019-10-17
Implement TOFU authentication for RetroShare

paquitaj:
This might be extremely dangerous in common RetroShare use case, but it
is extremely useful for chat-servers.
Of course it is disabled by default and one have to explicitly enable
it at compile time to have TOFU enabled RetroShare version.

### DIFF
--- a/libretroshare/src/pqi/authssl.cc
+++ b/libretroshare/src/pqi/authssl.cc
@@ -1159,7 +1159,13 @@ err: // TODO: this label is very short and might collide every easly
 /********************************************************************************/
 
 static int verify_x509_callback(int preverify_ok, X509_STORE_CTX *ctx)
-{ return AuthSSL::instance().VerifyX509Callback(preverify_ok, ctx); }
+{
+#ifdef RS_TOFU_AUTHENTICATION
+	return AuthSSL::instance().tofuVerifyX509Cb(preverify_ok, ctx);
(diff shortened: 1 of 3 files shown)


paquitaj:
@unsenep2pdev @csoler the continuous integration worked fine even if it is reported broken here, probably a github bug.
look at this links  https://ci.appveyor.com/project/RetroShare58622/retroshare/builds/28181575  https://travis-ci.org/RetroShare/RetroShare/builds/599173311

unsenep2pdev:
It's possible to have man-in-the-middle attack if your connection is over Tor hidden service? Can the attacker somehow intercept the tor connexion and accept as someone's friend?

csoler:
I don't think so. If used over Tor you should probably be safe since there is no exit nodes involved.

paquitaj:
Over Tor and I2P hidden services there is no risk of man in the middle even using TOFU.
With plain internet (no Tor, no I2P) the man in the middle risk is just at first connection attempt, once you got the PGP key of someone, there is no risk of man in the middle anymore as the PGP signature is verified at connection attempt, before accepting it.

PhenomRetroShare:
Hello,
Why don't you use https://github.com/RetroShare/ChatServer?

csoler:
I'm closing this branch. It is conflicted and offers a dangerous alternative that can be misused.

PR #1694, opened 2019-10-26
added notification when distant chat has undelivered messages. Update…

csoler:
…d the text to remove technical terms such as tunnels


felisucoibi:
step by step

PR #1696, opened 2019-10-30
Take advantage of modern C++ features

G10h4ck:
Harness some C++ modern features to make easier to write safer, cleaner and faster code, see commits messages for details

### DIFF
--- a/libretroshare/src/libretroshare.pro
+++ b/libretroshare/src/libretroshare.pro
@@ -503,6 +503,7 @@ HEADERS +=	util/folderiterator.h \
             util/stacktrace.h \
             util/rsdeprecate.h \
             util/cxx11retrocompat.h \
+    util/cxx17retrocompat.h \
             util/rsurl.h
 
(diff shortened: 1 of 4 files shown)


csoler:
This is very nice. Is that backward compatible? I mean can I change an existing flag definition into the new one and the code will still compile ?

G10h4ck:
It is completely retro-compatible in terms of serialization, in fact you can specify the underlining type, and even have uint8_t backed flags, but you will probably need a bit of refactor in term of usage because those enums are scoped (and then type safe) so you need to access the values like this EnumFlagTypeName::VALUE_NAME, and probably the compiler will complain about some type unsafe thing we do around like implicit conversion to uint and we will need to fix that too (which is good)

G10h4ck:
One of the features introduced by C++20 will make it even easier to port pre-existent code to the new flags type http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1099r5.html

PR #1697, opened 2019-10-30
Multiple improvements to global router

G10h4ck:
Verify messages hop by hop, this avoid the spreading of invalid messages through
      the network, as a side effect missing RsGxsId keys are propagated hop by hop
      from the sender toward the destination of the message.
    Temporarly store messages which validity could not be verified because of
      missing key in a queue, attempt request sender key and signature validation
      validity periodically until timeout, drop them after timeout, this should
      avoid messages sent by new users being dropped by the GRouter, which is very
      counterintuitive and frustrating for newcomers. The queue is not persistent
      accross RetroShare restarts, so messages not yet validated are lost when
      RetroShare is turned off.
    Do not drop forwarded messages that have unkown service id.
    When receiving a message with unkown service id inform the sender setting
      specific flag in receipt.
    Use type safe flags for GRouter items.
    RsTickEvent::prev_event_ago age is unsigned!
    p3IdService improve some internal implementations and API.
    Use change assert into static_assert for some checks that can be done at compile
      time.
    Solve a bunch of compiler warnings.
    Deprecate some cruft.

### DIFF
--- a/libretroshare/src/grouter/grouteritems.h
+++ b/libretroshare/src/grouter/grouteritems.h
@@ -27,7 +27,7 @@
 #include "serialiser/rstlvkeys.h"
 #include "rsitems/rsserviceids.h"
 #include "retroshare/rstypes.h"
-
+#include "retroshare/rsflags.h"
 #include "retroshare/rsgrouter.h"
 #include "groutermatrix.h"
 
@@ -83,40 +83,52 @@ class RsGRouterNonCopyableObject
 // and routing ID. Sub-items are responsible for providing the serialised data to be signed for
 // both signing and checking.
 
-class RsGRouterAbstractMsgItem: public RsGRouterItem
+enum class RsGrouterItemFlags : uint32_t
 {
-public:
-    explicit RsGRouterAbstractMsgItem(uint8_t pkt_subtype) : RsGRouterItem(pkt_subtype), flags(0) {}
-    virtual ~RsGRouterAbstractMsgItem() {}
+	NONE               = 0x0,
+	ENCRYPTED          = 0x1,
+	SERVICE_UNKNOWN    = 0x2
+};
+RS_REGISTER_ENUM_FLAGS_TYPE(RsGrouterItemFlags)
+
+struct RsGRouterAbstractMsgItem: RsGRouterItem
+{
+    explicit RsGRouterAbstractMsgItem(uint8_t pkt_subtype):
+        RsGRouterItem(pkt_subtype), flags(RsGrouterItemFlags::NONE) {}
 
     GRouterMsgPropagationId routing_id ;
     GRouterKeyId destination_key ;
     GRouterServiceId service_id ;
     RsTlvKeySignature signature ;		// signs mid+destination_key+state
-	uint32_t flags ; 					// packet was delivered, not delivered, bounced, etc
+
+	/// packet was delivered, not delivered, bounced, etc
+	RsGrouterItemFlags flags;
 };
 
-class RsGRouterGenericDataItem: public RsGRouterAbstractMsgItem, public RsGRouterNonCopyableObject
+struct RsGRouterGenericDataItem:
+        RsGRouterAbstractMsgItem, RsGRouterNonCopyableObject
 {
-    public:
-        RsGRouterGenericDataItem() : RsGRouterAbstractMsgItem(RS_PKT_SUBTYPE_GROUTER_DATA), data_size(0), data_bytes(NULL), duplication_factor(0) { setPriorityLevel(QOS_PRIORITY_RS_GROUTER) ; }
-        virtual ~RsGRouterGenericDataItem() { clear() ; }
+	RsGRouterGenericDataItem():
+	    RsGRouterAbstractMsgItem(RS_PKT_SUBTYPE_GROUTER_DATA),
+	    data_size(0), data_bytes(nullptr), duplication_factor(0)
+	{ setPriorityLevel(QOS_PRIORITY_RS_GROUTER); }
 
-        virtual void clear()
-        {
-            free(data_bytes);
-            data_bytes=NULL;
-        }
+	virtual ~RsGRouterGenericDataItem() { clear(); }
+	virtual void clear() { free(data_bytes); }
 
-		virtual void serial_process(RsGenericSerializer::SerializeJob j,RsGenericSerializer::SerializeContext& ctx);
+	virtual void serial_process(
+	        RsGenericSerializer::SerializeJob j,
+	        RsGenericSerializer::SerializeContext& ctx );
 
-        RsGRouterGenericDataItem *duplicate() const ;
+	RsGRouterGenericDataItem *duplicate() const;
 
-        // packet data
-        //
-        uint32_t data_size ;
-        uint8_t *data_bytes;
-        uint32_t duplication_factor ;	// number of duplicates allowed. Should be capped at each de-serialise operation!
+	/// packet data
+	uint32_t data_size;
+	uint8_t* data_bytes;
+
+	/** number of duplicates allowed. Should be capped at each de-serialise
+	 * operation! */
+	uint32_t duplication_factor;
 };
 
 class RsGRouterSignedReceiptItem: public RsGRouterAbstractMsgItem
--- a/libretroshare/src/grouter/grouteritems.h
+++ b/libretroshare/src/grouter/grouteritems.h
@@ -103,11 +103,14 @@ struct RsGRouterAbstractMsgItem: RsGRouterItem
 
 	/// packet was delivered, not delivered, bounced, etc
 	RsGrouterItemFlags flags;
+
+	~RsGRouterAbstractMsgItem();
 };
 
-struct RsGRouterGenericDataItem:
-        RsGRouterAbstractMsgItem, RsGRouterNonCopyableObject
+class RsGRouterGenericDataItem:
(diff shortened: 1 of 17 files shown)


libretroshare/src/grouter/grouteritems.h
```diff
@@ -83,40 +83,52 @@ class RsGRouterNonCopyableObject
...
     GRouterServiceId service_id ;
     RsTlvKeySignature signature ;		// signs mid+destination_key+state
-	uint32_t flags ; 					// packet was delivered, not delivered, bounced, etc
+
+	/// packet was delivered, not delivered, bounced, etc
+	RsGrouterItemFlags flags;
 };
 
-class RsGRouterGenericDataItem: public RsGRouterAbstractMsgItem, public RsGRouterNonCopyableObject
+struct RsGRouterGenericDataItem:
```
csoler: RsItems generally are classes, not structs. So we should keep the same type of data.

libretroshare/src/grouter/groutertypes.h
```diff
@@ -21,17 +21,22 @@
...
 #include "retroshare/rsgrouter.h"
 
-class RsGRouterGenericDataItem ;
+struct RsGRouterGenericDataItem;
 class RsGRouterSignedReceiptItem ;
 
-static const uint16_t GROUTER_CLIENT_ID_MESSAGES     = 0x1001 ;
+#ifndef V07_NON_BACKWARD_COMPATIBLE_CHANGE_GROUTER_SERVICE_ID
+/// Pay special attention fixing this, as may break retrocompatibility!
+RS_DEPRECATED_FOR(RsServiceType::MSG)
```
csoler, line 37: MSG? Probably a copy/paste error?

libretroshare/src/grouter/p3grouter.cc
```diff
@@ -1485,132 +1536,204 @@ bool p3GRouter::sliceDataItem(RsGRouterAbstractMsgItem *item,std::list<RsGRouter
...
+			Dbg2() << __PRETTY_FUNCTION__ << "  message is not at origin. "
+			       << "Setting message transmission to RECEIPT_OK" << std::endl;
 
-    // also note the incoming route in the routing matrix
+			it->second.data_status  = RS_GROUTER_DATA_STATUS_RECEIPT_OK;
+			it->second.receipt_item = receiptItem.release();
+			// no need of dluplicate() just release unique_ptr
+		}
 
-    if(!mTurtle->isTurtlePeer(receipt_item->PeerId()))
```
csoler, line 1604: Why is this removed?

libretroshare/src/grouter/groutertypes.h
G10h4ck: this doesn't mean that `RsServiceType::MSG` is deprecated, but that the near defined symbol `GROUTER_CLIENT_ID_MESSAGES` is deprecated and that `RsServiceType::MSG` should be used instead of that

libretroshare/src/grouter/p3grouter.cc
G10h4ck: It hasn't been removes it, has been moved at line 1649, because this method has been changed a plain diff won't help much

libretroshare/src/services/p3gxscircles.cc
```diff
@@ -1118,7 +1118,7 @@ bool p3GxsCircles::cache_request_load(const RsGxsCircleId &id)
 		return true;
 	}
 
-	int32_t age = 0;
+	uint32_t age = 0;
 	if (RsTickEvent::prev_event_ago(CIRCLE_EVENT_CACHELOAD, age))
```
csoler, line 1122: Technically it's not possible to ensure that age will not be negative. That is why this function uses a int32 instead of a uint32_t. So hanging to int32 is probably bad.

libretroshare/src/util/rstickevent.h
```diff
@@ -48,8 +49,8 @@ void    schedule_event(uint32_t event_type, rstime_t when, const std::string &el
 void    schedule_in(uint32_t event_type, uint32_t in_secs);
 void    schedule_in(uint32_t event_type, uint32_t in_secs, const std::string &elabel);
 
-int32_t event_count(uint32_t event_type);
-bool 	prev_event_ago(uint32_t event_type, int32_t &age);
+    int32_t event_count(uint32_t event_type);
+	bool prev_event_ago(uint32_t event_type, uint32_t& age);
```
csoler, line 53: probably not a good idea. See above comment.

retroshare.pri
```diff
@@ -262,6 +271,7 @@ rs_v07_changes {
     DEFINES += V07_NON_BACKWARD_COMPATIBLE_CHANGE_003
     DEFINES += V07_NON_BACKWARD_COMPATIBLE_CHANGE_004
     DEFINES += V07_NON_BACKWARD_COMPATIBLE_CHANGE_UNNAMED
+    DEFINES += V07_NON_BACKWARD_COMPATIBLE_CHANGE_GROUTER_SERVICE_ID
```
csoler, line 274: What is the need for that particular change?

libretroshare/src/retroshare/rsgrouter.h
```diff
@@ -19,19 +20,29 @@
...
+
+RS_DEPRECATED_FOR(RsGxsId)
+typedef RsGxsId GRouterKeyId;
+
+/**
+ * @note Some items use this type, when migrating to RsServiceType (2 bytes),
+ * special care should be taken to do it in a way which doesn't break
+ * retrocompatibility
+ */
+RS_DEPRECATED_FOR(RsServiceType)
```
csoler, line 39: I dont understand that. The GRouterServiceId is rather a client ID. Such a client is not necessarily a RSService. So merging the two limit possibilities and does not help much either.

libretroshare/src/rsitems/rsserviceids.h
```diff
@@ -40,7 +40,6 @@ enum class RsServiceType : uint16_t
 	FILE_DATABASE              = 0x0019,
 	SERVICEINFO                = 0x0020,
 	BANDWIDTH_CONTROL          = 0x0021,
-	MAIL                       = 0x0022,
```
csoler, line 43: If possible keep the line with "unused since XXX (formerly MAIL)". So when we use this number, we know that we dont have risks to collide with an existing service id that was only recently removed.

libretroshare/src/services/p3gxscircles.cc
csoler: I'm telling this because, even if the logic ensures that it behaves in some way, a bug can turn this number into a < 0 number and therefore make it way too large. If you can avoid that when it is computed, then it's fine.

libretroshare/src/services/p3msgservice.cc
```diff
@@ -1988,13 +1997,38 @@ void p3MsgService::manageDistantPeers()
...
-                mGRouter->registerKey(*it,GROUTER_CLIENT_ID_MESSAGES,"Messaging contact") ;
-            else
-                mGRouter->unregisterKey(*it,GROUTER_CLIENT_ID_MESSAGES) ;
-        }
+		for( auto it = std::as_const(own_id_list).begin();
+		     it != own_id_list.end(); ++it )
+		{
+			if(mShouldEnableDistantMessaging)
+			{
+#ifndef V07_NON_BACKWARD_COMPATIBLE_CHANGE_GROUTER_SERVICE_ID
```
csoler, line 2005: clearly at this point we see that this change is problematic because it adds unnecessary ifdefs and retrocompatibility issues.

libretroshare/src/services/p3msgservice.cc
```diff
@@ -1988,13 +1997,38 @@ void p3MsgService::manageDistantPeers()
...
+				// Keep for RetroShare < 0.6.6 retro-compatibility
+				mGRouter->registerKey(
```
csoler, line 2007: Same here! You get to register 2 GRouter clients in order to keep backward compat!

libretroshare/src/grouter/groutertypes.h
csoler: That's an error. See my other comments about GRouterId type.

libretroshare/src/grouter/p3grouter.cc
```diff
@@ -3,7 +3,8 @@
  *                                                                             *
  * libretroshare: retroshare core library                                      *
  *                                                                             *
- * Copyright 2013 by Cyril Soler <csoler@users.sourceforge.net>                *
+ * Copyright (C) 2013  Cyril Soler <csoler@users.sourceforge.net>              *
+ * Copyright (C) 2019  Gioacchino Mazzurco <gio@eigenlab.org>                  *
```
csoler, line 7: Personnally I dont add my own name everytime I modify a file, unless for a full rewrite, which means the file can be considered as new. For small changes, individual commits hold your name already and git blame give the exact spectrum of your contributions.

csoler:
also I would really prefer not to see simple memory management turned into a forest of std::unique_ptr. These things are nice are useful, except when not using them is much simpler. A typical example is in the handleIncoming method(). All methods called below take const RsItems* and do not delete them. They are all deleted in the parent. Turning all this into std::unique_ptr makes it totally impossible to follow/read.

csoler:
... and it makes the important changes (mostly the queuing of unhandled items) totally impossible to spot and check in the PR.

G10h4ck:
Superseded by other smaller pull requests

PR #1700, opened 2019-11-10
Code to embed webUI server in the main executable

csoler:


libretroshare/src/jsonapi/jsonapi.cpp
```diff
@@ -432,31 +432,71 @@ bool JsonApiServer::requestNewTokenAutorization(const std::string& token)
...
+            return false;
+    return true;
+}
+
+static bool parseToken(const std::string& clear_token,std::string& user,std::string& passwd)
+{
+	uint32_t last_index=0;
+	uint32_t nb_colons=0;
+
+	for(uint32_t i=0;i<clear_token.length();++i)
```
G10h4ck: you can use std::string::find

libretroshare/src/jsonapi/jsonapi.cpp
```diff
@@ -473,33 +513,57 @@ bool JsonApiServer::revokeAuthToken(const std::string& token)
...
+    if(p != passwd)
+	{
+        p = passwd;
+		IndicateConfigChanged();
+	}
+	return true;
+}
+
+
 bool JsonApiServer::authorizeToken(const std::string& token)
```
G10h4ck: Now that we have JsonApiServer::authorizeUser, we should probably delete JsonApiServer::authorizeToken or at least mark it as RS_DEPRECATED_FOR(authorizeUser) in the header

libretroshare/src/jsonapi/jsonapi.h
```diff
@@ -120,40 +120,48 @@ struct JsonApiServer : RsSingleJobThread, p3Config
...
 
 	/**
 	 * @brief Revoke given auth token
 	 * @jsonapi{development}
-	 * @param[in] token decoded
+	 * @param[in] user par of the decoded token
 	 * @return true if the token has been revoked, false otherwise
 	 */
-	bool revokeAuthToken(const std::string& token);
+	bool revokeAuthToken(const std::string& user);
```
G10h4ck: did you changed the implementation too to revoke user? in that case the method name should be changed too to revokeAuthUser

libretroshare/src/jsonapi/jsonapi.h
```diff
@@ -120,40 +120,48 @@ struct JsonApiServer : RsSingleJobThread, p3Config
...
+
 	/**
 	 * @brief Get decoded version of the given encoded token
 	 * @jsonapi{development,unauthenticated}
-	 * @param[in] token encoded
+	 * @param[in] radix64_token encoded
 	 * @return token decoded
 	 */
-	static std::string decodeToken(const std::string& token);
+	static std::string decodeToken(const std::string& radix64_token);
```
G10h4ck, line 98: please keep using camelCase instead of snake_case

csoler:
So I did the changes to have the webUI be served by the same restbed server than jsonAPI. It seems to work. I'm not yet tested all the restart options though.

libretroshare/src/jsonapi/jsonapi.cpp
```diff
@@ -115,11 +117,40 @@ JsonApiServer::corsOptionsHeaders =
 	return false;
 }
 
-JsonApiServer::JsonApiServer(uint16_t port, const std::string& bindAddress,
-        const std::function<bool(const std::string&)> newAccessRequestCallback ) :
-    mPort(port), mBindAddress(bindAddress),
-    mNewAccessRequestCallback(newAccessRequestCallback),
-    configMutex("JsonApiServer config")
+static bool is_alphanumeric(char c) { return (c>='0' && c<'9') || (c>='a' && c<='z') || (c>='A' && c<='Z') ;}
+static bool is_alphanumeric(const std::string& s)
```
G10h4ck: I think this two functions should be moveved to rsutils, we have something like rsstring or the like header for this kind of stuff

libretroshare/src/jsonapi/jsonapi.cpp
```diff
@@ -416,47 +424,57 @@ void JsonApiServer::registerHandler(
 			else session->close(rb::UNAUTHORIZED);
 		} );
 
-	mService.publish(resource);
+    _resources.push_back(resource);
```
G10h4ck: Why do we keep resource in an owned variable, before mService was doing that on it's own, moreover please keep the convention of using mName for member variables

libretroshare/src/jsonapi/jsonapi.cpp
```diff
@@ -416,47 +424,57 @@ void JsonApiServer::registerHandler(
...
+{
+    mNewAccessRequestCallback = callback;
+}
 
-bool JsonApiServer::requestNewTokenAutorization(const std::string& token)
+bool JsonApiServer::requestNewTokenAutorization(const std::string& user)
 {
-	if(rsLoginHelper->isLoggedIn() && mNewAccessRequestCallback(token))
-		return authorizeToken(token);
+    std::string passwd;
```
G10h4ck, line 428: This can't work, password need to be a parameter too

libretroshare/src/jsonapi/jsonapi.cpp
```diff
@@ -473,33 +491,65 @@ bool JsonApiServer::revokeAuthToken(const std::string& token)
...
+    cfgmgr->addConfiguration("jsonapi.cfg",this);
+
+    RsFileHash hash;
+    loadConfiguration(hash);
+}
+
+bool JsonApiServer::authorizeUser(const std::string& user,const std::string& passwd)
+{
+	if(!is_alphanumeric(user) || !is_alphanumeric(passwd))
+        return false;
```
G10h4ck, line 498: This should at least print an RsErr() when happens

libretroshare/src/jsonapi/jsonapi.cpp
```diff
@@ -473,33 +491,65 @@ bool JsonApiServer::revokeAuthToken(const std::string& token)
...
 		IndicateConfigChanged();
-		return true;
 	}
-	return false;
+	return true;
 }
 
-/*static*/ std::string JsonApiServer::decodeToken(const std::string& token)
+
+// bool JsonApiServer::authorizeToken(const std::string& token)
```
G10h4ck: please delete deadcode

libretroshare/src/jsonapi/jsonapi.cpp
```diff
@@ -551,3 +601,34 @@ void JsonApiServer::handleCorsOptions(
         const std::shared_ptr<restbed::Session> session )
 { session->close(rb::NO_CONTENT, corsOptionsHeaders); }
 
+int JsonApiServer::status()
```
G10h4ck: this method should be bool and just return `RestbedService::isRunning()` or be deleted if `RestbedService::isRunning()` is public

libretroshare/src/jsonapi/jsonapi.h
```diff
@@ -50,110 +55,55 @@ extern JsonApiServer* jsonApiServer;
...
+	bool requestNewTokenAutorization(const std::string& user) override;
 
-	/**
-	 * @brief Revoke given auth token
-	 * @jsonapi{development}
-	 * @param[in] token decoded
-	 * @return true if the token has been revoked, false otherwise
-	 */
-	bool revokeAuthToken(const std::string& token);
+    void registerResourceProvider(const JsonApiResourceProvider *);
```
G10h4ck, line 86: Why you get a pointer and not a reference? What would be the point to call this method on a nullptr ? Same comments apply to the other similar methods

libretroshare/src/jsonapi/jsonapi.h
```diff
@@ -201,16 +174,18 @@ struct JsonApiServer : RsSingleJobThread, p3Config
...
 	        RsGenericSerializer::SerializeContext& ctx,
-	        const std::shared_ptr<restbed::Session> session )
+	        const std::shared_ptr<rb::Session> session )
 	{
-		return checkRsServicePtrReady(
-		            serviceInstance.get(), serviceName, ctx, session );
+		return checkRsServicePtrReady( serviceInstance.get(), serviceName, ctx, session );
 	}
+
+	std::vector<std::shared_ptr<rb::Resource> > _resources;
```
G10h4ck: Why do we need to keep track on ourself of resource, if possible we should leave that work to restbed service

libretroshare/src/jsonapi/jsonapi.h
```diff
@@ -201,16 +174,18 @@ struct JsonApiServer : RsSingleJobThread, p3Config
...
+	std::set<const JsonApiResourceProvider *> _resource_providers;
```
G10h4ck: please keep using the mName convention for member variables

libretroshare/src/jsonapi/restbedservice.cc
```diff
@@ -0,0 +1,93 @@
...
+ * You should have received a copy of the GNU Lesser General Public License    *
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.       *
+ *                                                                             *
+ *******************************************************************************/
+
+#include "util/rsthreads.h"
+#include "util/rsdebug.h"
+#include "restbedservice.h"
+
+RestbedService::RestbedService()
```
G10h4ck, line 28: Please use proper initializers instead of explicitily assign variables in constructor body

libretroshare/src/jsonapi/restbedservice.cc
```diff
@@ -0,0 +1,93 @@
...
+        return true;
+
+	mService->stop();
+
+	RsThread::ask_for_stop();
+
+	while(isRunning())
+	{
+		std::cerr << "(II) shutting down restbed service." << std::endl;
+		rstime::rs_usleep(1000*1000);
```
G10h4ck, line 55: this function is deprecated, please use the one suggested by deprecation notice

libretroshare/src/jsonapi/restbedservice.cc
```diff
@@ -0,0 +1,93 @@
...
+
+bool RestbedService::restart()
+{
+    stop();
+    start("Restbed Service");
+
+    return true;
+}
+
+bool RestbedService::stop()
```
G10h4ck: Because this function won't return before stop if completed, better calling this fullStop to be consistent with other RsThread derived classes

libretroshare/src/jsonapi/restbedservice.cc
```diff
@@ -0,0 +1,93 @@
...
+void RestbedService::runloop()
+{
+	auto settings = std::make_shared< restbed::Settings >( );
+	settings->set_port( mListeningPort );
+	settings->set_bind_address( mBindingAddress );
+	settings->set_default_header( "Connection", "close" );
+
+	if(mService->is_up())
+	{
+		std::cerr << "(II) WebUI is already running. Killing it." << std::endl;
```
G10h4ck: Please use proper RsWarn() output classes instead of std::cerr and also fox the message

libretroshare/src/jsonapi/restbedservice.h
```diff
@@ -0,0 +1,56 @@
+/*******************************************************************************
+ * libretroshare/src/jsonapi/: restbedservice.h                                *
+ *                                                                             *
+ * libretroshare: retroshare core library                                      *
+ *                                                                             *
+ * Copyright 2019-2019 Cyril Soler                                             *
+ *                                                                             *
+ * This program is free software: you can redistribute it and/or modify        *
+ * it under the terms of the GNU Lesser General Public License as              *
```
G10h4ck: We should use AGPLv3 at least for the optional components that involve web interactions tecnologies, moreover restbed is relead under AGPL so also RetroShare must comply with AGPL when compiling with this options

libretroshare/src/retroshare/rsjsonapi.h
```diff
@@ -0,0 +1,118 @@
+/*******************************************************************************
+ * libretroshare/src/retroshare: rsjsonapi.h                                   *
+ *                                                                             *
+ * Copyright (C) 2018-2019  Gioacchino Mazzurco <gio.eigenlab.org>             *
+ * Copyright (C) 2019-2019  Cyril Soler <csoler@users.sourceforge.net>         *
+ *                                                                             *
+ * This program is free software: you can redistribute it and/or modify        *
+ * it under the terms of the GNU Lesser General Public License as              *
```
G10h4ck: This should be AGPL see previous comment about JSON API licensing

libretroshare/src/retroshare/rsjsonapi.h
```diff
@@ -0,0 +1,118 @@
...
+
+#include <string>
+
+class p3ConfigMgr;
+class JsonApiResourceProvider;
+
+class RsJsonAPI
+{
+public:
+    enum {
```
G10h4ck: This enum and its values is not useful so it should be removed

libretroshare/src/retroshare/rsjsonapi.h
```diff
@@ -0,0 +1,118 @@
...
+    virtual void registerResourceProvider(const JsonApiResourceProvider *)=0;
+    virtual void unregisterResourceProvider(const JsonApiResourceProvider *)=0;
+    virtual bool hasResourceProvider(const JsonApiResourceProvider *)=0;
+
+	/**
+	 * @brief Get status of the json api server
+	 * @jsonapi{development}
+	 * @return the status picked in the enum JSONAPI_STATUS_UNKNOWN/RUNNING/NOT_RUNNING
+	 */
+	virtual int status() =0;
```
G10h4ck: Should be deleted like the enum upster, bool isRunning() or similar is more then enough

libretroshare/src/retroshare/rsjsonapi.h
```diff
@@ -0,0 +1,118 @@
...
+	virtual bool restart() =0;
+	virtual bool stop()  =0;
+
+	virtual void setBindingAddress(const std::string& address) =0;
+	virtual void setListeningPort(uint16_t port) =0;
+	virtual uint16_t listeningPort() const =0;
+
+    virtual void connectToConfigManager(p3ConfigMgr *cfgmgr)=0;
+
+    virtual void registerResourceProvider(const JsonApiResourceProvider *)=0;
```
G10h4ck, line 75: Please add doxygen documentation at least for those methods intended to make this service extensible

libretroshare/src/retroshare/rsjsonapi.h
```diff
@@ -0,0 +1,118 @@
...
+
+    //////////////// @Gio: The methods below should not be accessible from the API server !
+    ///
+	/**
+	 * @brief Add new auth (user,passwd) token to the authorized set, creating the token user:passwd internally.
+	 * @param[in] alphanumeric_user username to autorize decoded
+	 * @param[in] alphanumeric_passwd passwd to autorize decoded
+	 * @return true if the token has been added to authorized, false if error occurred
+	 */
+	virtual bool authorizeUser(const std::string& alphanumeric_user,const std::string& alphanumeric_passwd)=0;
```
G10h4ck, line 111: this should be accessible from JSON API, and they where authenticated too, think of Android the only way you have to interact with the library is through JSON API

libretroshare/src/retroshare/rsjsonapi.h
```diff
@@ -0,0 +1,118 @@
...
+
+	/**
+	 * @brief Revoke given auth token
+	 * @param[in] user par of the decoded token
+	 * @return true if the token has been revoked, false otherwise
+	 */
+	virtual bool revokeAuthToken(const std::string& user)=0;
```
G10h4ck, line 118: this should be accessible from JSON API, and they where authenticated too, think of Android the only way you have to interact with the library is through JSON API

libretroshare/src/retroshare/rsjsonapi.h
```diff
@@ -0,0 +1,118 @@
...
+
+	/**
+	 * @brief Get authorized tokens
+	 * @return the set of authorized encoded tokens
+	 */
+    virtual std::map<std::string,std::string> getAuthorizedTokens() =0;
```
G10h4ck, line 124: this should be accessible from JSON API, and they where authenticated too, think of Android the only way you have to interact with the library is through JSON API

libretroshare/src/retroshare/rsjsonapi.h
```diff
@@ -0,0 +1,118 @@
...
+
+	/**
+	 * @brief Check if given JSON API auth token is authorized
+	 * @param[in] token decoded
+	 * @return tru if authorized, false otherwise
+	 */
+	virtual bool isAuthTokenValid(const std::string& token)=0;
```
G10h4ck, line 131: this should be accessible from JSON API, and they where authenticated too, think of Android the only way you have to interact with the library is through JSON API

libretroshare/src/rsserver/p3webui.cc
```diff
@@ -0,0 +1,206 @@
+/*******************************************************************************
+ * libretroshare/src/rsserver: p3webui.cc                                      *
+ *                                                                             *
+ * libretroshare: retroshare core library                                      *
+ *                                                                             *
+ * Copyright 2019-2019 Cyril Soler                                             *
+ *                                                                             *
+ * This program is free software: you can redistribute it and/or modify        *
+ * it under the terms of the GNU Lesser General Public License as              *
```
G10h4ck, line 9: AGPLv3

libretroshare/src/rsserver/p3webui.cc
```diff
@@ -0,0 +1,206 @@
...
+	"image/svg+xml",
+	"font/ttf",
+	"font/woff",
+	"application/octet-stream",
+};
+
+#ifdef WINDOWS_SYS
+const std::string RsWebUI::DEFAULT_BASE_DIRECTORY = "data/webui";
+#else
+const std::string RsWebUI::DEFAULT_BASE_DIRECTORY = "/usr/share/retroshare/webui/";
```
G10h4ck: I think we have an OS aware method to retrieve data directory that work well in all setups, android, windows, mac, linux, appimage etc etc this won't work on all setup for example on android

libretroshare/src/rsserver/p3webui.cc
```diff
@@ -0,0 +1,206 @@
...
+		{
+			const auto request = session->get_request( );
+			const std::string filename = request->get_path_parameter( "filename" );
+			std::string directory = request->get_path_parameter( "dir" );
+
+            if(!directory.empty())
+                directory += "/";
+
+			std::string resource_filename = _base_directory + "/" + directory + filename;
+			std::cerr << "Reading file: \"" << resource_filename << "\"" << std::endl;
```
G10h4ck: Use RsDbg() instead of raw std::cerr

libretroshare/src/rsserver/p3webui.cc
```diff
@@ -0,0 +1,206 @@
...
+			std::ifstream stream( resource_filename, std::ifstream::in );
+
+			if ( stream.is_open( ) )
+			{
+				const std::string body = std::string( std::istreambuf_iterator< char >( stream ), std::istreambuf_iterator< char >( ) );
+
+				std::cerr << "  body length=" << body.length() << std::endl;
```
G10h4ck: RsDbg()

libretroshare/src/rsserver/p3webui.cc
```diff
@@ -0,0 +1,206 @@
...
+    return rtab;
+}
+
+
+void p3WebUI::setHtmlFilesDirectory(const std::string& html_dir)
+{
+    _base_directory = html_dir;
+}
+
+int p3WebUI::status() const
```
G10h4ck, line 170: burn this method :p

libretroshare/src/rsserver/p3webui.cc
```diff
@@ -0,0 +1,206 @@
...
+{
+    if(rsJsonAPI->status()==RsJsonAPI::JSONAPI_STATUS_RUNNING && rsJsonAPI->hasResourceProvider(this))
+        return WEBUI_STATUS_RUNNING;
+    else
+        return WEBUI_STATUS_NOT_RUNNING;
+}
+
+void p3WebUI::setUserPassword(const std::string& passwd)
+{
+#ifdef RS_JSONAPI
```
G10h4ck, line 180: Can webui be enabled without JSON API being enabled??

libretroshare/src/rsserver/p3webui.h
```diff
@@ -0,0 +1,50 @@
+/*******************************************************************************
+ * libretroshare/src/rsserver: p3webui.h                                       *
+ *                                                                             *
+ * libretroshare: retroshare core library                                      *
+ *                                                                             *
+ * Copyright 2019-2019 Cyril Soler                                             *
+ *                                                                             *
+ * This program is free software: you can redistribute it and/or modify        *
+ * it under the terms of the GNU Lesser General Public License as              *
```
G10h4ck, line 9: AGPLv3

libretroshare/src/rsserver/p3webui.h
```diff
@@ -0,0 +1,50 @@
...
+#include <vector>
+#include <memory>
+#include "retroshare/rswebui.h"
+#include "jsonapi/jsonapi.h"
+
+class p3WebUI: public RsWebUI, public JsonApiResourceProvider
+{
+public:
+    p3WebUI(){}
+    virtual ~p3WebUI(){}
```
G10h4ck: use = default;

libretroshare/src/rsserver/p3webui.h
```diff
@@ -0,0 +1,50 @@
...
+
+    // implements RsWebUI
+
+    virtual void setHtmlFilesDirectory(const std::string& html_dir) override;
```
G10h4ck, line 37: override imply virtual so no need to specify it

libretroshare/src/rsserver/rsinit.cc
```diff
@@ -391,6 +392,22 @@ int RsInit::InitRetroShare(const RsConfigOptions& conf)
 	if(!RsAccounts::init(rsInitConfig->optBaseDir,error_code))
 		return error_code ;
 
+#ifdef RS_JSONAPI
+    // We create the JsonApiServer this early, because it is needed *before* login
+
+    RsInfo() << "Allocating jsonAPI server (not launched yet) " << std::endl;
+    JsonApiServer *jas = new JsonApiServer();
+    jas->setListeningPort(conf.jsonApiPort);
+    jas->setBindAddress(conf.jsonApiBindAddress);
```
G10h4ck, line 401: Is it still possible to set json api port and binding address from command line options?

retroshare-gui/src/gui/settings/WebuiPage.cpp
```diff
@@ -181,15 +191,21 @@ void WebuiPage::onAllIPCBClicked(bool /*checked*/)
...
         return;
     }
-    QDesktopServices::openUrl(QUrl(QString("http://localhost:")+QString::number(ui.port_SB->value())));
+
+    emit passwordChanged();
 }
+
+void WebuiPage::onStartWebBrowserClicked()
+{
+    QDesktopServices::openUrl(QUrl(QString("http://localhost:")+QString::number(rsJsonAPI->listeningPort())));
```
G10h4ck, line 205: localhost may be traslateb both to `127.0.0.1` and to `::1` in our case restbed either bind on one or the other so depending on host configuration opening that url may fail, so please change `localhost` with the correct bind address, which moreover could be configured to something else

retroshare-gui/src/gui/settings/rsharesettings.cpp
```diff
@@ -30,6 +30,9 @@
 #include "rsharesettings.h"
 #include "gui/MainWindow.h"
 
+// for this one, we'd rather use a file in retroshare/*h list.
+#include <jsonapi/jsonapi.h>
```
G10h4ck, line 34: is the comment still correct?

retroshare-gui/src/gui/settings/rsharesettings.cpp
```diff
@@ -1150,9 +1153,23 @@ void RshareSettings::setWebinterfaceEnabled(bool enabled)
     setValueToGroup("Webinterface", "enabled", enabled);
 }
 
+QString RshareSettings::getWebinterfaceFilesDirectory()
+{
+#ifdef WINDOWS_SYS
+    return valueFromGroup("Webinterface","directory","data/webui/").toString().toStdString();
+#endif
+    return valueFromGroup("Webinterface","directory","/usr/share/retroshare/webui/").toString();
```
G10h4ck, line 1161: again those hardcoded values won't work on all platforms/distro, use the method we have somewhere to get data directory

retroshare-gui/src/gui/settings/rsharesettings.cpp
```diff
@@ -1150,9 +1153,23 @@ void RshareSettings::setWebinterfaceEnabled(bool enabled)
...
+void RshareSettings::setWebinterfaceFilesDirectory(const QString& s)
+{
+    setValueToGroup("Webinterface","directory",s);
+}
+
+
 uint16_t RshareSettings::getWebinterfacePort()
 {
-    return valueFromGroup("Webinterface", "port", 9090).toUInt();
+    return valueFromGroup("Webinterface", "port", 1984).toUInt();
```
G10h4ck: we shouldn't have this configuration at weui level but at jsoapi server level

retroshare.pri
```diff
@@ -653,6 +653,10 @@ android-* {
     }
     CONFIG *= no_libresapihttpserver
     CONFIG -= libresapihttpserver
+
+	CONFIG += no_rs_service_terminal_login
+	CONFIG += no_rs_service_webui_terminal_password
```
G10h4ck, line 658: this 2 line must be removed

G10h4ck (changes requested):

libretroshare/src/jsonapi/jsonapi.cpp
csoler: ok

retroshare.pri
csoler: Android is the only platform where terminal login is not needed. So it's preferable to have terminal logins available by default and remove them only in android build, which is what I'm doing here.

libretroshare/src/jsonapi/jsonapi.cpp
csoler: that is because I used to keep an array of all resources (including the one from webui) then I changed the code and forgot to remove this.

libretroshare/src/jsonapi/jsonapi.h
csoler: because it's stored in a vector, and I have an abstract class here.

libretroshare/src/jsonapi/jsonapi.h
csoler: everywhere else for parameters we use this case.

libretroshare/src/jsonapi/jsonapi.h
csoler: because some of them are created in the constructor, whereas the resources get requested later by the parent RestbedService class when it starts.

libretroshare/src/jsonapi/restbedservice.cc
csoler: we'd better replace the implementation of rs_usleep to call sleep_for. I cannot find in the .h of sleep_for whether it is microseconds or miliseconds.

libretroshare/src/jsonapi/restbedservice.h
csoler: There's no web interaction is restbedservice. But given that it is in jsonapi/ directory, I agree to use AGPLv3. It's simpler for writing the licence file.

libretroshare/src/rsserver/p3webui.cc
csoler: nope. This one is in rsserver.

libretroshare/src/rsserver/p3webui.h
csoler: I dont think so: override means that the parent class contains a virtual method that the current is replacing. It does not imply that the current method is virtual.

libretroshare/src/rsserver/rsinit.cc
csoler: yes I think so. It's the responsibility of main() to set the port and bind address parameters, so that rsinit sees them in conf.

libretroshare/src/jsonapi/jsonapi.h
G10h4ck: In that case take a `const JsonApiResourceProvider&` as paramether, and then store an std::reference_wrapper<JsonApiResourceProvider> in the container see https://en.cppreference.com/w/cpp/utility/functional/reference_wrapper it has same performances then pointer but guarantee it is not null and make clear that the object is not owned

libretroshare/src/jsonapi/restbedservice.cc
G10h4ck: You can use it like that `std::this_thread::sleep_for(std::chrono::seconds(1));` substitute seconds with whatever unit is more comfortable to your case and that's it ;)

libretroshare/src/rsserver/p3webui.h
G10h4ck: that is implied by the fact that it is declared virtual in the parent class ;)

libretroshare/src/jsonapi/jsonapi.h
csoler: come on! Why do such complicated things when a pointer does exactly what we need? If somebody supplied NULL, the code will crash and the bug will be fixed very easily. I agree with the general need to put safeguards in the code and make it bug-free by design, but here the compromise clearly let the pointer option win.

PR #1707, opened 2019-11-18
Added message length check for channels

defnax:
Added for Channel messages to check the length of the Message and picture attachments same like on forums method.

It will help better to know whats my limit and helps when screenshots or pics attached
without the length check you will never know if the Channel message goes out with picture attachments.


felisucoibi:
+1

PR #1716, opened 2019-11-24
Fixed feedreader compile and changed icons for the feed item

defnax:
Fixed Feedreader compile and changed icons for the feed items to up to date with the rs gui feeds


plugins/FeedReader/FeedReaderPlugin.cpp
```diff
@@ -55,7 +55,7 @@ extern "C" {
 #ifdef WIN32
 	__declspec(dllexport)
 #endif
-	uint32_t RETROSHARE_PLUGIN_revision = 0;
+	uint32_t RETROSHARE_PLUGIN_revision = abs(atoi(RS_EXTRA_VERSION)) ;
```
csoler: abs() is here to replace a cast to uint32_t? Given that RS_EXTRA_VERSION has no reason to be negative, I would just do 
uint32_t RETROSHARE_PLUGIN_revision  = (uint32_t)atoi(RS_EXTRA_VERSION);

defnax: changed it now

PR #1724, opened 2019-11-27
Fix Continuous Integration

G10h4ck:


rex4539 (approved):
ACK.

This patch fixes my local macOS build when built with `qmake  CONFIG+=debug CONFIG+=rs_macos10.15 CONFIG+=no_direct_chat`.

It still fails when built with `qmake  CONFIG+=debug CONFIG+=rs_macos10.15 CONFIG+=no_direct_chat CONFIG+=rs_jsonapi CONFIG+=rs_webui` but perhaps this could be fixed with a separate patch.

PR #1734, opened 2019-12-01
aded PB PetaByte to filelists.

felisucoibi:
To be tested.


felisucoibi:
Tested, needs more love, I think there are kind of defaults, i'll try to do it.

felisucoibi:
I think the logic for TB is harcoded in libretroshare/src/file_sharing/hash_cache.cc if somebodyu can have a look.

csoler:
The loop in misc.cpp should go to 6 instead of 5.

felisucoibi:
Finished, it works.

PR #1735, opened 2019-12-01
Switched security items to new notification system

csoler:


libretroshare/src/pqi/pqissllistener.cc
```diff
@@ -486,6 +486,20 @@ int	pqissllistenbase::continueSSL(IncomingSSLInfo& incoming_connexion_info, bool
...
 		}
 
+		if(rsEvents)
+		{
+			auto ev = std::unique_ptr<RsAuthSslConnectionAutenticationEvent>(new RsAuthSslConnectionAutenticationEvent);
+
+			ev->mSslId = incoming_connexion_info.sslid;
+			ev->mPgpId = incoming_connexion_info.gpgid;
+			ev->mSslCn = incoming_connexion_info.sslcn;
+			ev->mLocator = RsUrl(sockaddr_storage_iptostring(incoming_connexion_info.addr));
```
G10h4ck: You can pass sockaddr_storage directly to RsUrl no need to convert it to a string before ;)

libretroshare/src/retroshare/rsevents.h
```diff
@@ -170,3 +171,51 @@ class RsEvents
...
+    enum ConnextionErrorCode: uint8_t {
+        UNKNOWN_ERROR                   = 0x00,
+        MISSING_AUTHENTICATION_INFO     = 0x01,
+        PGP_SIGNATURE_VALIDATION_FAILED = 0x02,
+        MISMATCHED_PGP_ID               = 0x03,
+        NO_CERTIFICATE_SUPPLIED         = 0x04,
+        NOT_A_FRIEND                    = 0x05,
+        MISSING_CERTIFICATE             = 0x06,
+        IP_IS_BLACKLISTED               = 0x07,
+        NO_ERROR                        = 0x08,
```
G10h4ck: Please use 0 to indicate no error to be forward compatible with std::error_condition, also better you use an enum class outside the event struct for similar reasons ;)

libretroshare/src/pqi/authssl.cc
```diff
@@ -1218,7 +1220,8 @@ int AuthSSLimpl::VerifyX509Callback(int /*preverify_ok*/, X509_STORE_CTX* ctx)
 		{
 			ev->mSslCn = sslCn;
 			ev->mPgpId = pgpId;
-			ev->mErrorMsg = errMsg;
+			ev->mErrorCode = RsAuthSslConnectionAutenticationEvent::MISSING_AUTHENTICATION_INFO;
```
G10h4ck: why did you remove `errMsg` propagation here?

retroshare-gui/src/gui/NewsFeed.cpp
```diff
@@ -176,6 +181,88 @@ void NewsFeed::sortChanged(int index)
...
+    {
+        RsAuthSslConnectionAutenticationEvent e = *pssl_e;	// make a copy because we lose memory ownership here
+
+		RsQThreadUtils::postToObject( [=]()
+		{
+			/* Here it goes any code you want to be executed on the Qt Gui
+			 * thread, for example to update the data model with new information
+			 * after a blocking call to RetroShare API complete, note that
+			 * Qt::QueuedConnection is important!
+			 */
```
G10h4ck: You need a copy of the `shared_ptr` in this context otherwise the event is probably deleted before this function could be executed. So you should reference `event` directly inside the lambda body so the lambda will automatically `[=]` make a copy for you, so you should move also the cast inside the lambda and upstair instead of checking for `nullptr` in the cast check for `event->mType == RsEventType::AUTHSSL_CONNECTION_AUTENTICATION` which is also cheaper

retroshare-gui/src/gui/notifyqt.cpp
```diff
@@ -89,6 +89,10 @@ void NotifyQt::SetDisableAll(bool bValue)
 	}
 }
 
+NotifyQt::~NotifyQt()
```
G10h4ck: use `= default;` for empty destructors as the compiler can optimize better those

G10h4ck: Or did you mean to unregister the event handler here?

retroshare-gui/src/gui/NewsFeed.cpp
```diff
@@ -74,6 +75,8 @@ NewsFeed::NewsFeed(QWidget *parent) :
     RsAutoUpdatePage(1000,parent),
     ui(new Ui::NewsFeed)
 {
+	rsEvents->registerEventsHandler( [this](std::shared_ptr<const RsEvent> event) { handleEvent(event); }, mEventHandlerId );
```
G10h4ck, line 76: Is mEventHandlerId initialized somewhere?

G10h4ck (changes requested):

retroshare-gui/src/gui/notifyqt.cpp
csoler: actually it's a remaining of the previous use of the handler in notifyQt but I moved it to Feed

retroshare-gui/src/gui/NewsFeed.cpp
csoler: actually I just make the copy line 194. The original shared_ptr is not used in the function.

retroshare-gui/src/gui/NewsFeed.cpp
csoler: I thought that registerEventsHandler() initializes the ID for you. It would actually make sense!

libretroshare/src/retroshare/rsevents.h
csoler: ok

felisucoibi:
+1

retroshare-gui/src/gui/NewsFeed.cpp
G10h4ck: It initialize it if 0 is passed, otherwise it assumes the handler id was carrectly pregenerated, this is neccessary for certain cases, what I can do is creating a type with deleted costructor so one cannot do it by error

retroshare-gui/src/gui/NewsFeed.cpp
G10h4ck: the problem is that if you don' refer to a `shared_ptr` in that context the object will not survive, and when the function is executed in the Qt thread `e` will point to an already deleted memory location, so you need to refer to the original `shared_ptr` or a properly `dynamic_pointer_cast` copy of it and not to a raw pointer in the place you will finally use it

G10h4ck (changes requested):

libretroshare/src/pqi/authssl.cc
```diff
@@ -1222,7 +1222,7 @@ int AuthSSLimpl::VerifyX509Callback(int /*preverify_ok*/, X509_STORE_CTX* ctx)
 			ev->mSslId = sslId;
 			ev->mPgpId = pgpId;
 			ev->mErrorMsg = errMsg;
-			ev->mErrorCode = RsAuthSslConnectionAutenticationEvent::MISSING_AUTHENTICATION_INFO;
+			ev->mErrorCode = RsAuthSslConnectionAutenticationEvent::AuthenticationCode::MISSING_AUTHENTICATION_INFO;
```
G10h4ck, line 1225: please declare `AuthenticationCode` outside of `RsAuthSslConnectionAutenticationEvent`, otherwise it will be almost unusable via JSON API

G10h4ck (changes requested):

libretroshare/src/chat/p3chatservice.cc
```diff
@@ -855,12 +855,8 @@ bool p3ChatService::handleRecvChatMsgItem(RsChatMsgItem *& ci)
         {
 #ifdef RS_DIRECT_CHAT
 			/* notify public chat message */
-			RsServer::notify()->AddPopupMessage(
-			            RS_POPUP_GROUPCHAT,
-			            ci->PeerId().toStdString(), "", message );
-			RsServer::notify()->AddFeedItem(
-			            RS_FEED_ITEM_CHAT_NEW,
-			            ci->PeerId().toStdString(), message, "" );
+			RsServer::notify()->AddPopupMessage( RS_POPUP_GROUPCHAT, ci->PeerId().toStdString(), "", message );
```
G10h4ck, line 858: is this still necessary?

libretroshare/src/pqi/pqissl.cc
```diff
@@ -1253,8 +1262,10 @@ int pqissl::accept_locked( SSL *ssl, int fd,
...
 
-	if(rsBanList && !rsBanList->isAddressAccepted(
-	            foreign_addr, checking_flags, check_result ))
+    if( RsX509Cert::getCertSslId(*SSL_get_peer_certificate(ssl)) != PeerId())
+        std::cerr << "(EE) pqissl::accept_locked(): PeerId() is " << PeerId() << " but certificate ID is " << RsX509Cert::getCertSslId(*SSL_get_peer_certificate(ssl)) << std::endl;
+
+	if(rsBanList && !rsBanList->isAddressAccepted( foreign_addr, checking_flags, check_result ))
 	{
 		RsErr() << __PRETTY_FUNCTION__
 		        << " Refusing incoming SSL connection from blacklisted "
```
G10h4ck, line 1271: This should be RsInfo() instead of RsErr()

libretroshare/src/chat/p3chatservice.cc
csoler: I removed all addFeedItem(), but I did not take care of the popup messages in this PR

libretroshare/src/pqi/pqissl.cc
csoler: done

PR #1740, opened 2019-12-19
Added new rich text editor & redesigned Posted Composer

defnax:
* Added new Rich Text Editor
* Redesigned the Create Post Composer for Posted Links
* When only link is added display different thumbnail & some small fixes.

The new Rich Text Editor can be soon easyly embeded for Channels & Forums Composer too.

image

this composer can be more improved some ideas:
* selection box for subscribed topics. (for easy switch where want to post)
image

when someone want improve it please make a pr after this Pr is merged.


felisucoibi:
ole y ole!!! you are the best!

retroshare-gui/src/util/MRichTextEdit.h
```diff
@@ -0,0 +1,101 @@
+/*
```
csoler: There's a license problem here. retroshare-gui is AGPLv3+. Depending on where you found this file, it may be possible to re-license it with AGPLv3+ as well,

retroshare-gui/src/util/MRichTextEdit.cpp
```diff
@@ -0,0 +1,620 @@
+/*
```
csoler: same here about license.

retroshare-gui/src/util/MRichTextEdit.h
defnax: from here 
https://github.com/Anchakor/MRichTextEditor
https://github.com/Anchakor/MRichTextEditor/blob/master/LICENSE.LGPL

retroshare-gui/src/util/MRichTextEdit.cpp
defnax: https://github.com/Anchakor/MRichTextEditor

retroshare-gui/src/util/MRichTextEdit.h
defnax: @csoler what now?

csoler: Can you ask the author if you can relicense this file to AGPLv3+ ? If not, we cannot add this file. If it comes from Qt examples, it shouldn't be a problem if I remember correctly.

PR #1743, opened 2019-12-23
removed @jsonapi from RsFiles#turtleSearchRequest at rsfiles.h

zeners:
removed jsonapi annotation, because doxygen throws an error and a callback-function can't be called in a JSON-API Call

imho the required feature has to be wrapped in another function, which will collect the callbacked answers to be async queried in JSON-API

### DIFF
--- a/libretroshare/src/retroshare/rsfiles.h
+++ b/libretroshare/src/retroshare/rsfiles.h
@@ -401,11 +401,11 @@ public:
 
 	/**
 	 * @brief Request remote files search
-	 * @jsonapi{development}
+     * //jsonapi{development} // how to define a function callback in json?
 	 * @param[in] matchString string to look for in the search. If files deep
 	 *	indexing is enabled at compile time support advanced features described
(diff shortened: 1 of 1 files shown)


csoler:
Adding the [in] wasn't sufficient to fix doxygen??

zeners:
@csoler no, insert [in] doesn't help
it is less doxygen, it is more json-generator
make[1]: *** [Makefile.libretroshare:2585: ../../jsonapi-generator/src/jsonapi-includes.inl] Error 234

G10h4ck:
Do you have doxygen 1.8.16 ? That version is known to be broken see https://github.com/doxygen/doxygen/issues/7236 in that case install instead doxygen 1.8.15 or report the bug to your distribution so they downgrade to 1.8.15 which is not affected by that bug

Methods that take callbacks are exposed in JSON API via event stream, so each time the callback would be called you get the parameters passed to the callback streamed as JSON

RS code is correct so I close this PR

zeners:
@G10h4ck thx 4 reply

PR #1746, opened 2019-12-27
[Need Fixes] Added more colors for quoted syntax highlighting

defnax:
* Added 8 Color Levels

its not backward compatible on RSTextbrowser "> " coloring with space not worked there only with 1 color.
Future quoted posts get be colorized up to 8 colors.
Old Quoted text will still stay with green color.

image
image


retroshare-gui/src/util/RsSyntaxHighlighter.cpp
```diff
@@ -32,23 +32,86 @@ void RsSyntaxHighlighter::highlightBlock(const QString &text)
 	int index = 0;
 	QStringList lines = text.split(endl);
 	foreach (const QString &line, lines) {
-		if(line.trimmed().startsWith('>')) {
-			setFormat(index, line.length(), quotationFormat);
+		if(line.trimmed().startsWith(">>>>>>>>") || line.trimmed().startsWith("> > > > > > > >")) {
```
csoler, line 35: This code is not reliable enough: what happens if due to several editing, the message starts with "> > >>>> > >" ? Some spaces have been removed and some not.
I suggest that you use a regexp to search for the string, and test in various situations.

csoler: This might be the reason beyond the PR not working on existing quoted text.

defnax: The PR works perfect on MimeTExtEdit (with & without space) and on RSTextBrowser without space,
but im not sure where is the bug on RSTEXTBROWSER
possible somewhere bug in read mode? normaly it must work backward compatible
but not sure why it doesnt works there.



**MimeTextEdit**
image


**RSTextBrowser**
image

defnax: > This code is not reliable enough: what happens if due to several editing, the message starts with "> > >>>> > >" ? Some spaces have been removed and some not.
> I suggest that you use a regexp to search for the string, and test in various situations.

but i think this will not happen or some one editing the text manually & adding extra spacing.
when i reply now old quoted text it formats for the new quoted text then it get be highlighted in RSTextBrowser.
It was nice when some one find the issue why in RSTextbrowser doesnt works coloring with 1 space

image

defnax: when use "> > > > " some time the text gets be hightlighted but not on every post
image

it looks like some quoted text has only 1 line? then it will stays always green? when im right
but some quoted text have own seperated line then it gets colored .

defnax:
this needs more work to get be useable

defnax:
if some one have time, can fix this pr?
with spacing the highlighting not works on RSTextBrowser

PR #1748, opened 2019-12-29
V0.6 improved gui 2

csoler:
This PR adds a parameter to registerHandler() in RsEvents so that events are only send to the handlers that expect a particular type.

I also improved the consistency in RsGossipDiscoveryEvent and RsBroadcastDiscoveryEvent.

(This PR is 2 commits above PR 1735. It should be examined after 1740 is merged)


libretroshare/src/services/rseventsservice.cc
```diff
@@ -176,30 +176,25 @@ void RsEventsService::handleEvent(std::shared_ptr<const RsEvent> event)
...
+		}
+
+        // Call all clients that registered a callback for this event type
+
+        for(auto cbit: mHandlerMaps[event_type_index])
+            cbit.second(event);
+
+        // Also call all clients that registered with NONE, meaning that they expect all events
+
+        for(auto cbit: mHandlerMaps[static_cast<uint32_t>(RsEventType::NONE)])
```
G10h4ck, line 197: AFAIR RsEventType can be used as key type in the map directly no need to use plain integers + casting

libretroshare/src/services/rseventsservice.cc
```diff
@@ -176,30 +176,25 @@ void RsEventsService::handleEvent(std::shared_ptr<const RsEvent> event)
...
+		RS_STACK_MUTEX(mHandlerMapMtx); /*  LOCKED AREA */
 
-	if(mCallback)
-	{
-		mCallback(event); // It is relevant that this happens outside mutex
-		mCallback = std::function<void(std::shared_ptr<const RsEvent>)>(nullptr);
-		goto getHandlerFromMapLock;
+		if(event_type_index >= mHandlerMaps.size() || event_type_index < 1)
+		{
+			RsErr() << "Cannot handle an event of type " << event_type_index << ": out of scope!" << std::endl;
```
G10h4ck, line 186: Please add `__PRETTY_FUNCTION__` at the begin and change "scope!" with "range!"

G10h4ck (changes requested):

libretroshare/src/services/rseventsservice.cc
csoler: it's a vector, not a map!

PR #1749, opened 2019-12-29
Forum composer small ui changes

defnax:
* Added PlaceHolder text
* Set default font family to "MS Sans Serif"
* Fixed to get work Place Holder text on QTextEdit, seems there was a second unwanted line

image


felisucoibi:
+1

PR #1762, opened 2020-01-16
V0.6 bug fixing

csoler:


csoler:
This pull request should fix the crashes in news feed. To be tested.
I also fixed somehow the logic in deleting items. Before that, the FeedItem class used to send a signal from within its destructor. That is definitely not a good idea.
The next step should be to get rid of the TokenQueue system in this part of the code too.

hunbernd:
No crash.

But I don't get notifications about new mail, posts etc.

csoler:
Thx. I'll check that. I assume that the corresponding checkboxes are activates in your RS?

csoler:
I fixed it.

PR #1771, opened 2020-01-26
[Experimental] Video player for downloaded mp4 videos

defnax:
i has forked from my other work with gif/webp pr -> https://github.com/RetroShare/RetroShare/pull/1767

* Added QVideoWidget for SubFileItem to play downloaded Videos

at the moment i enabled only for mp4 extensions the player, it needs to be tested on other platforms Linux & Mac.

image

I wanted demonstrate that is possible to play videos with on the SubFileItem.

This idea can be used for Share short videos & play on Channels & comment.
Or on Posted links would be usefull too when it supports attachments feature.
Its at the moment basic needs to improved.

There is no file limit yet or compress feature for videos.

other idea with own Video Player Page https://github.com/RetroShare/RetroShare/pull/1773

Todo:
* Only add player widget when needed to SubFileItem

PR #1773, opened 2020-01-27
[Experimental] Video player page for main window

defnax:
* Added Video player page for main window to play downloaded channel videos.

it needs to be tested on other platforms Linux & Mac.
im not sure which video formats are supported mp4 works normally on all platforms with Qt media player.
I has not set yet extension limitations for the Player.
This player can be improved, i wanted share my work before its lost.

Other idea Video Player for SubFileItem: https://github.com/RetroShare/RetroShare/pull/1771

Todo:
* Need to add a second Play button or action for play on the buildin Player Page (SubFileItem)

image


csoler:
This is very nice, but I do not see the advantage of having a video player inside RS, since your system plays videos when you click on the files. We've been fighting against RS becoming a bloatware, and this IMHO is clearly related to this question. I'd be glad to hear other users thoughts about this.

defnax:
its only a demonstration, it needs to be more improved.
i want show that is possible here.

At moment nobody do this type of things on rs
its more usefull to play for short videos & comment & like the video....

Often to open the system player takes not fast, you avoid to lose time to open the external player.

This player will be more useful for Channels or Posted or the embeded player in SubFileItem.
Tribler has similiar player for the downaloded videos.

felisucoibi:
is usefull for example for a youtube like plugin. +1 def

PR #1774, opened 2020-01-27
JSON API improvements

G10h4ck:


libretroshare/src/jsonapi/jsonapi.cpp
```diff
@@ -127,18 +129,16 @@ JsonApiServer::corsOptionsHeaders =
...
 {
-	uint32_t colonIndex = 0;
-	uint32_t colonCounter = 0;
+	auto colonIndex = std::string::npos;
 	const auto tkLen = clear_token.length();
 
-	for(uint32_t i=0; i < tkLen && colonCounter < 2; ++i)
-		if(clear_token[i] == ':') { ++colonCounter; colonIndex = i; }
-		else if(!librs::util::is_alphanumeric(clear_token[i])) return false;
+	for(uint32_t i=0; i < tkLen; ++i)
```
csoler, line 144: This is ambiguous. What happens if the token is "user:passwd:passwd" ? then the user will be "user:passwd". That means a passwd cannot contain ":".  In summary, I think that the passwd should be required to be alphanumeric. Otherwise you need to send it in base64 in which case the passwd can be anything and the parsing can be made non ambiguous by ensuiring that the token contains exactly one ":" and that the encoded passwd is what remains after ":".

libretroshare/src/serialiser/rstypeserializer.h
```diff
@@ -909,6 +910,27 @@ struct RsTypeSerializer
 	                       t_RsTlvList<TLV_CLASS,TLV_TYPE>& member,
 	                       RsJson& jDoc );
 
+	[[noreturn]] static void fatalUnknownSerialJob(int j)
```
csoler, line 913: what's this????

libretroshare/src/util/rsstring.h
```diff
@@ -30,10 +30,10 @@ namespace librs {
 
 		bool ConvertUtf8ToUtf16(const std::string& source, std::wstring& dest);
 		bool ConvertUtf16ToUtf8(const std::wstring& source, std::string& dest);
-
+#if 0
 		bool is_alphanumeric(char c) ;
```
csoler, line 34: has these two been moved somewhere else?

libretroshare/src/serialiser/rstypeserializer.h
G10h4ck: https://en.cppreference.com/w/cpp/language/attributes/noreturn  it tells that the function won't return, aka if the program get there it will be terminated like when you call exit(...)

libretroshare/src/jsonapi/jsonapi.cpp
G10h4ck: If the token is like "somethingA:somethinchB:somethingC" then user is "somethingA" and password is "somethinchB:somethingC"

G10h4ck: In fact the token is transferred as a base64 encoded string, RFC7617 has specific rules, IMHO the most relevant parts are:

- For the user-id, recipients MUST support all characters defined in the "UsernameCasePreserved" profile defined in Section 3.3 of [RFC7613], with the exception of the colon (":") character.

- For the password, recipients MUST support all characters defined in the "OpaqueString" profile defined in Section 4.2 of [RFC7613].

- Constructs the user-pass by concatenating the user-id, a single colon (":") character, and the password

Basically in the token what is before the first ':' is to be considered username what is before is to be considered password, and both user and password con be non alphanumeric and not even ASCII

libretroshare/src/util/rsstring.h
G10h4ck: They are not used anymore so I kept them under `#ifdef 0` if we need them in the future.

libretroshare/src/jsonapi/jsonapi.cpp
G10h4ck: The line is now `if(clear_token[i] == ':') { colonIndex = i; break; }`

PR #1782, opened 2020-02-09
Reenable UDP stunner and add documentation

sehraf:
Retroshare implements the STUN protocol (old RFC 3489) to determine the NAT type it is running behind.
While the actual determination logic resides in the class `pqiNetStateBox`, the UDP stunner is responsible for running the actual protocol.

The UDP stunner was disabled a while ago because it seemed only be used to determine the external IP address (using STUN). While this is true as well it's only a byproduct.
As @drbob pointed out it should be re-enabled to allow RS to determine the NAT type again - and based on that knowledge - build more UDP connections (there are multiple types and currently only the one (*direct*) is working).

The patch further adds some (minor) documentation to the UDP stunner, makes the implementation more RFC conform as well as document the RFCs responsible and allows (optionally) to utilize the it for external IP determination. (While this is generally problematical it might be useful for corner cases)

### DIFF
--- a/libretroshare/src/tcponudp/udpstunner.cc
+++ b/libretroshare/src/tcponudp/udpstunner.cc
@@ -47,6 +47,12 @@ const uint32_t TOU_STUN_MAX_RECV_RATE = 25; /* every 25 seconds */
 const int32_t TOU_STUN_DEFAULT_TARGET_RATE  = 15; /* 20 secs is minimum to keep a NAT UDP port open */
 const double  TOU_SUCCESS_LPF_FACTOR = 0.90;
 
+/*
+ * based on RFC 3489
+ */
+const uint16_t STUN_BINDING_REQUEST  = 0x0001;
+const uint16_t STUN_BINDING_RESPONSE = 0x0101;
+
 #define EXCLUSIVE_MODE_TIMEOUT	300
 
 UdpStunner::UdpStunner(UdpPublisher *pub)
@@ -187,7 +193,6 @@ int	UdpStunner::releaseExclusiveMode(std::string holder, bool forceStun)
 		return 0;
 	}
 
-	rstime_t now = time(NULL);
 	mExclusiveMode = false;
 	if (forceStun)
 	{
@@ -217,6 +222,7 @@ int	UdpStunner::releaseExclusiveMode(std::string holder, bool forceStun)
 	}
 
 #ifdef DEBUG_UDP_STUNNER_FILTER
+	rstime_t now = time(NULL);
 	std::cerr << "UdpStunner::cancelExclusiveMode() Canceled. Was in ExclusiveMode for: " << now - mExclusiveModeTS;
 	std::cerr << " secs";
 	std::cerr << std::endl;
@@ -326,7 +332,7 @@ bool UdpStunner::locked_handleStunPkt(void *data, int size, struct sockaddr_in &
 #endif
 		/* generate a response */
 		int len;
-		void *pkt = UdpStun_generate_stun_reply(&from, &len);
+		void *pkt = UdpStun_generate_stun_reply(&from, &len, data);
 		if (!pkt)
 			return false;
 
@@ -485,7 +491,7 @@ bool    UdpStun_response(void *stun_pkt, int size, struct sockaddr_in &addr)
 		return false;
 	}
 
-	if (htons(((uint16_t *) stun_pkt)[0]) != 0x0101)
+	if (htons(((uint16_t *) stun_pkt)[0]) != STUN_BINDING_RESPONSE)
 	{
 		/* not a response */
 		return false;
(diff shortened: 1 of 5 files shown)


libretroshare/src/tcponudp/udpstunner.cc
```diff
@@ -47,6 +47,12 @@ const uint32_t TOU_STUN_MAX_RECV_RATE = 25; /* every 25 seconds */
 const int32_t TOU_STUN_DEFAULT_TARGET_RATE  = 15; /* 20 secs is minimum to keep a NAT UDP port open */
 const double  TOU_SUCCESS_LPF_FACTOR = 0.90;
 
+/*
+ * based on RFC 3489
+ */
+const uint16_t STUN_BINDING_REQUEST  = 0x0001;
+const uint16_t STUN_BINDING_RESPONSE = 0x0101;
```
G10h4ck: Wouldn't it be more appropriate to have those as static constxepr memebers of UdpStunner ?

libretroshare/src/tcponudp/udpstunner.cc
```diff
@@ -517,33 +523,41 @@ bool UdpStun_generate_stun_pkt(void *stun_pkt, int *len)
...
-	/* transaction id - should be random */
-	((uint32_t *) stun_pkt)[1] = (uint32_t) htonl(0x0020); 
-	((uint32_t *) stun_pkt)[2] = (uint32_t) htonl(0x0121); 
-	((uint32_t *) stun_pkt)[3] = (uint32_t) htonl(0x0111); 
-	((uint32_t *) stun_pkt)[4] = (uint32_t) htonl(0x1010); 
+	/* RFC 3489
+	 *	The transaction ID is used to correlate requests and responses.
+	 *
+	 * RFC 5389 introduces a mmgic cokie at the location where preciously the transaction ID was located:
+	 *	In RFC 3489 [RFC3489], this field was part of
```
G10h4ck: why is RFC3489 repeated twice?

libretroshare/src/tcponudp/udpstunner.cc
```diff
@@ -517,33 +523,41 @@ bool UdpStun_generate_stun_pkt(void *stun_pkt, int *len)
...
+	 *	the transaction ID; placing the magic cookie in this location allows
+	 *	a server to detect if the client will understand certain attributes
+	 *	that were added in this revised specification.
+	 */
+	((uint32_t *) stun_pkt)[1] = htonl(RsRandom::random_u32());
+	((uint32_t *) stun_pkt)[2] = htonl(RsRandom::random_u32());
+	((uint32_t *) stun_pkt)[3] = htonl(RsRandom::random_u32());
+	((uint32_t *) stun_pkt)[4] = htonl(RsRandom::random_u32());
```
G10h4ck: This 4 lines would e probably easier substituted by a call of `RsRandom::random_bytes`

libretroshare/src/tcponudp/udpstunner.h
```diff
@@ -53,10 +53,15 @@ class TouStunPeer
 		return; 
 	}
 	
+	/// id for identification
 	std::string id;
+	/// remote address
 	struct sockaddr_in remote, eaddr;
```
G10h4ck: Please document both members

G10h4ck (changes requested):

libretroshare/src/tcponudp/udpstunner.cc
sehraf: Sure, will do.

libretroshare/src/tcponudp/udpstunner.cc
sehraf: It's a plain copy+paste from the RFC.
Can remove the reference in square brackets.

libretroshare/src/tcponudp/udpstunner.cc
sehraf: Can do

libretroshare/src/tcponudp/udpstunner.h
sehraf: Ah, oversaw that! Thanks for the review!

sehraf:
I've rebased and squashed the fixups.

G10h4ck (approved):

G10h4ck:
Now it looks ok to me, I'd wait a few days for @csoler and @drbob to say something, and then merge it

csoler:
What is the priority of the external IP information using STUN w.r.t. other---more reliable---methods? Normally the expected behavior should be that external IP given by friends (that is 100% reliable) should erase previous estimates given by STUN. Is that the case already?

sehraf:
When you enable this mechanic (it's off by default) it counts as one (like all other sources, I think)
This happens as soon as **two** dht peers have reported the same external address.

Edit:
If desired I can change it to only use it as a fallback.

sehraf:
For the record:
* you must use `CONFIG+=use_dht_stunner_ext_ip` to use the dht stunner as a source of external IP
* dht stunner will only return an external IP when it was reported by two peers
* the switch turns on the old behavior 
* it counts as one like all other sources do, see https://github.com/RetroShare/RetroShare/blob/6f9d875fc69fab52b20e5f886fc614949d937250/libretroshare/src/pqi/p3netmgr.cc#L666

csoler:
> When you enable this mechanic (it's off by default) it counts as one (like all other sources, I think)
> This happens as soon as **two** dht peers have reported the same external address.
> 
> Edit:
> If desired I can change it to only use it as a fallback.

Well, given that DHT is poisoned by peers who impersnnate your friends with a wrong IP, I am a bit skeptical about the effect on the DHT stunner and the reliability of the IP it will give. So it's important that other more reliable sources (such as discovery) wipe out the result of DHT stunner when you use it.

sehraf:
I've moved the block that gets the external IP from DHT stunner to the end and added a `if (address_votes.empty()) ` to only use it as fallback when all others haven't provided an IP address.

PR #1784, opened 2020-02-10
Update code to get Wiki / Wire and PhotoShare compiling again

drbob:
Patch to enable unfinished services to be compiled.
configurations on .pro files are disabled by default


defnax:
great thank you :)

felisucoibi:
nice! rs is alive, nothing work atm but will and with lot of new things.

defnax:
need to test when get time , if wiki or photoshare works with the latest rs .

PR #1792, opened 2020-02-14
Set new default column width for author & date

defnax:
* Set new default column width for author & date


defnax:
this wont help for the resizable problem

PR #1796, opened 2020-02-16
Switch PhotoService to use GxsComments

drbob:
WIP improvements to PhotoService:
 - switch p3photoservice to use gxcommentservice
 - clean out old RsPhotoComment classes
 - redesign PhotoDialog to use standard GxsCommentService classes
 - remove unneeded old Comment classes from GUI
 - fix indenting for PhotoDialog and p3photoservice.h
 - fix layout issue in Extra Widgets for GxsGroupDialog


defnax:
you removing the nice commenting for photos?
i never liked yet the treewidget commenting method where we using for channels/posted.
Hope some one develope a nice new comment dialog.

i wish for commenting like how on youtube or fb/instagram like commenting for  channel/posted/photos service.

comment

drbob:
hi defnax, 

I'm switching comments to the standard GxsComment widgets. These are used by both Channels and Posted and so are well tested and should be fully functional. The current PhotoComments don't appear to work anyway.

If we redesign the Comment Widgets, then it needs to be in a general manner, that can be used as an alternative for any of the services. This means it needs to provide the same interface as GxsCommentDialog. I expect that we could re-skin the GxsCommentTreeWidget to look like what you are showing in you images, but I'm not going to work on that until I've got these services into a better state.

PR #1797, opened 2020-02-16
Removed unfinished checkbox

defnax:
removed checkbox which was not yet finished to use.


defnax:
this was added mistakly in a previos pr from me , it was not finished seems i forget to remove that checkbox.

PR #1805, opened 2020-02-22
Revert "[Experimental] Video player page for main window"

defnax:
Reverts RetroShare/RetroShare#1773


defnax:
seems now i found the revert button here

PR #1810, opened 2020-02-23
Messages improvements

defnax:
* Moved the buttons to a better place (like thunderbird)
* removed the top frame
* changed to use the right icons for the context menu
* Added to display the Box name on the Main Tab when switching box
* Fixed to clear the Message Widgets Labels when switching the Boxes (Inbox to Trash)
* Added a Splitter for the Boxes List
* Added a "More" Menu Button for other actions
* Cleanup & Removed old icons with new one for Composer

image


csoler:
The new layout is better IMHO because it is more compact.
* It took me a while to find the Compose button. Maybe you should move it next to "Search" 
* Another issue (already here before your PR): the toolbar icons in the compose window are super small.
* the icons for Reply, forward, etc are quite old-fashioned. Would be better to have flat ones in the new style. Or maybe just remove the text? The icons alone may be sufficient. What do you think?

defnax:
> The new layout is better IMHO because it is more compact.
> 
> * It took me a while to find the Compose button. Maybe you should move it next to "Search"

i wanted like same how on email clients the postion is often top of near the inbox
easier to find for the email client users

image



> * Another issue (already here before your PR): the toolbar icons in the compose window are super small.

you mean this?
image
that needs to be replaced too some mail icons needs better ones

> * the icons for Reply, forward, etc are quite old-fashioned. Would be better to have flat ones in the new style. Or maybe just remove the text? The icons alone may be sufficient. What do you think?

i wanted like thunderbird with text, mostly on email cllients by default they keep the text.
yes its planned i wanted change the compose/tag/reply/reply all, foward/delete icons too but later when this is merged,
i need to look for nice flat ones 
maybe thinking add the "More" Button later for use as menu like in Thunderbird

example thunderbird
image
Opera Mail
image

defnax:
updated with new icons
image

PR #1814, opened 2020-02-28
Better chat history loading

cyisfor:
Yeah... just take these as suggestions, no guarantee I didn't break everything and just haven't noticed yet...


cyisfor:
I just noticed that I was always loading the remote name and never the local name, so I added code to check for historyIt->incoming.

Also the force push is just because I forgot to set the author name or something.

defnax:
@csoler hope you dont forget this tp check

retroshare-gui/src/gui/chat/ChatWidget.cpp
```diff
@@ -393,30 +430,99 @@ void ChatWidget::init(const ChatId &chat_id, const QString &title)
...
-					RsIdentityDetails details;
-					time_t start = time(nullptr);
-					while (!rsIdentity->getIdDetails(RsGxsId(historyIt->peerName), details))
-					{
-						std::this_thread::sleep_for(std::chrono::milliseconds(10));
-						if (time(nullptr)>start+2)
-						{
-							std::cerr << "ChatWidget History haven't found Id Details and have wait 1 sec for it." << std::endl;
-							break;
+				const std::string
```
csoler: This is bad because std::string will truncate the data to fir first 0. You'd better use this:
std::string str = historyIt->chatPeerId.toStdString();

retroshare-gui/src/gui/chat/ChatWidget.cpp
```diff
@@ -393,30 +430,99 @@ void ChatWidget::init(const ChatId &chat_id, const QString &title)
...
+					notbeinghashableisagoodidea((const char*)historyIt->chatPeerId.toByteArray());
+				struct lookup_info& state = statecache[notbeinghashableisagoodidea];
+				if(!state.hasName) {
+					int tries;
+					for(tries=0;tries<4;++tries) {
+						assert(!state.hasName);
```
csoler: generally we dont use assert() except when testing the code, but not to be merged in master. assert crashes the program without warning. Here it is preferable to gracefully quit the function. Use exceptions if you prefer, or a proper return after printing an error message.

retroshare-gui/src/gui/chat/ChatWidget.cpp
```diff
@@ -393,30 +430,99 @@ void ChatWidget::init(const ChatId &chat_id, const QString &title)
...
-					{
-						std::this_thread::sleep_for(std::chrono::milliseconds(10));
-						if (time(nullptr)>start+2)
-						{
-							std::cerr << "ChatWidget History haven't found Id Details and have wait 1 sec for it." << std::endl;
-							break;
+				const std::string
+					notbeinghashableisagoodidea((const char*)historyIt->chatPeerId.toByteArray());
+				struct lookup_info& state = statecache[notbeinghashableisagoodidea];
+				if(!state.hasName) {
```
csoler: uninitialized memory read. Consequently, this outcome of this test is random.

retroshare-gui/src/gui/chat/ChatWidget.cpp
```diff
@@ -393,30 +430,99 @@ void ChatWidget::init(const ChatId &chat_id, const QString &title)
...
+					int tries;
+					for(tries=0;tries<4;++tries) {
+						assert(!state.hasName);
+						if(state.isDirect) {
+						GOT_DIRECT:
```
csoler: these labels are a bit misleading: we're not in a switch and there's no goto going there. If you want to mark some par of the code with some kind of milestone or so, keep this but comment out the line so we know it's not actually used.

retroshare-gui/src/gui/chat/ChatWidget.cpp
```diff
@@ -393,30 +430,99 @@ void ChatWidget::init(const ChatId &chat_id, const QString &title)
...
+								RsIdentityDetails details;
+								if(rsIdentity->getIdDetails(state.gxs, details)) {
+									state.name = details.mNickname;
+									state.hasName = true;
+									break;
+								}
+							}
+						} else if(chatId.isLobbyId()) {
+							// XXX:
+							// no joke, this is how libretroshare/src/pqi/p3historymgr.c ACTUALLY
```
csoler: you're right. It's quite bad because it supposes that the endian will be kept between two calls. Since history is loaded and saved on the same machine, it is the case. however, that conversion is inelegant for the least.

retroshare-gui/src/gui/chat/ChatWidget.cpp
csoler: ah forget this. I didn't see the goto later on.

csoler:
Tx Cy. Sorry for taking so long to review this. There's just 3 things to fix: proper initialization of state_cache members to avoid uninitialized memory read, remove assert(), and proper conversion of Id to string using toStdString();

cyisfor:
Thanks for reviewing it. I definitely shouldn't have left that assert in there. As for uninitialized memory, turns out C++ makes that pretty difficult. If you read here:
https://en.cppreference.com/w/cpp/container/map/operator_at
it says that if the element does not exist, then it is inserted, and it says that inserted elements are either value-initialized, or zero-initialized. Long story short, state.hasName is always going to be false initially.

I ended up avoiding using the chat lobby ID at all, since it was supposed to be the one posting the message, not the chat lobby that got reported. So I had to mess with the code for saving chat history, so it would save that message poster, instead of telling us that every message was written by some guy named "General discussion chat (EN)" or something. So I'll push a few more commits to show what I've fixed. It's still got some issues, but... it works.

...I probably need to remove some sass from my code.

PR #1785, opened 2020-02-10
[NEED REVIEW] Convert #defines to enum class

sehraf:
rsconfig.h contains a large amount of `#define` for enum-like values. Convert these to `enum class` to get type safety.
This uncovered a few missing cases in switch statements and one bogus comparison `p3linkmgr.cc:1066` which is **currently not fixed yet**!

```c++
	uint32_t ownNetMode = mNetMgr->getNetworkMode();
```
is supposed to be 
```c++
	RsNetworkMode ownNetMode = mNetMgr->getNetworkMode();
```
which is later bitwise compared with `RS_NET_MODE_xyz`.

As soon as this is figured out the PR is merge-able.

PR #1815, opened 2020-02-29
Get the basics of TheWire service working

drbob:
Added REF fields to Wire Msgs to allow replys to work well.
Added description of how the Message fields should be used.
Added WireGroupDialog for creation of WireGroups.
Updated PulseAddDialog to support publishing Pulses and Replies.
Added classes to display Pulses / Groups.
Added single selection point of WireGroup for publishing.
Added basic "refresh" to reload new GXS data.

### DIFF
--- a/libretroshare/src/retroshare/rswire.h
+++ b/libretroshare/src/retroshare/rswire.h
@@ -3,7 +3,7 @@
  *                                                                             *
  * libretroshare: retroshare core library                                      *
  *                                                                             *
- * Copyright 2012-2012 by Robert Fernie <retroshare@lunamutt.com>              *
+ * Copyright 2012-2020 by Robert Fernie <retroshare@lunamutt.com>              *
  *                                                                             *
  * This program is free software: you can redistribute it and/or modify        *
  * it under the terms of the GNU Lesser General Public License as              *
@@ -45,11 +45,7 @@ class RsWireGroup
 
 
 /***********************************************************************
- * So pulses operate in the following modes.
- *
- * => Standard, a post to your own group.
- * => @User, gets duplicated on each user's group.
- * => RT, duplicated as child of original post.
+ * RsWire - is intended to be a Twitter clone - but fully decentralised.
  *
  * From Twitter: 
  *  twitter can be: embedded, replied to, favourited, unfavourited, 
@@ -73,6 +69,36 @@ class RsWirePlace
 
 };
 
+/************************************************************************
+ * Pulse comes in three flavours.
+ *
+ *
+ * Original Msg Pulse
+ * - Spontaneous msg, on your own group.
+ * - mPulseType = WIRE_PULSE_TYPE_ORIGINAL_MSG
+ * - Ref fields are empty.
+ *
+ * Reply to a Pulse (i.e Retweet), has two parts.
+ * as we want the retweet to reference the original, and the original to know about reply.
+ * This info will be duplicated in two msgs - but allow data to spread easier.
+ *
+ * Reply Msg Pulse, will be Top-Level Msg on Publisher's Group.
+ * - mPulseMode = WIRE_PULSE_TYPE_REPLY_MSG
+ * - Ref fields refer to Parent (InReplyTo) Msg.
+ *
+ * Reply Reference, is Child Msg of Parent Msg, on Parent Publisher's Group.
+ * - mPulseMode = WIRE_PULSE_TYPE_REPLY_REFERENCE
+ * - Ref fields refer to Reply Msg.
+ * - NB: This Msg requires Parent Msg for complete info, while other two are self-contained.
+ ***********************************************************************/
+
+#define  WIRE_PULSE_TYPE_ORIGINAL_MSG        (0x0001)
+#define  WIRE_PULSE_TYPE_REPLY_MSG           (0x0002)
+#define  WIRE_PULSE_TYPE_REPLY_REFERENCE     (0x0004)
+
+#define  WIRE_PULSE_TYPE_SENTIMENT_POSITIVE  (0x0010)
+#define  WIRE_PULSE_TYPE_SENTIMENT_NEUTRAL   (0x0020)
+#define  WIRE_PULSE_TYPE_SENTIMENT_NEGATIVE  (0x0040)
 
 class RsWirePulse
 {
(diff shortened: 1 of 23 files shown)


csoler:
Can you fix compilation? 
* These two are missing in FORMS: gui/TheWire/PulseDetails.ui, gui/TheWire/WireGroupItem.ui
* lots of undefined references to WireGroupDialog. I dont know if that's because a file is missing. If so, it's surprising that you can compile yourself.

The proper patch to compile is:
`
       HEADERS += \
               gui/TheWire/PulseItem.h 
                gui/TheWire/WireDialog.h 
                gui/TheWire/PulseAddDialog.h 
               gui/TheWire/PulseDetails.h 
               gui/TheWire/WireGroupDialog.h 
               gui/TheWire/WireGroupItem.h 
        FORMS += gui/TheWire/PulseItem.ui 
                gui/TheWire/WireDialog.ui 
                gui/TheWire/PulseAddDialog.ui 
               gui/TheWire/PulseDetails.ui 
               gui/TheWire/WireGroupItem.ui 
        SOURCES += gui/TheWire/PulseItem.cpp 
                gui/TheWire/WireDialog.cpp 
                gui/TheWire/PulseAddDialog.cpp 
               gui/TheWire/PulseDetails.cpp 
               gui/TheWire/WireGroupDialog.cpp 
               gui/TheWire/WireGroupItem.cpp 
`

libretroshare/src/retroshare/rswire.h
```diff
@@ -73,26 +69,63 @@ class RsWirePlace
...
-
-//	RsWirePlace mPlace;
+	// Store actual Pulse here.
+	std::string mPulseText;
+
+	uint32_t mPulseType;
+
+	// These Ref to the related (parent or reply) if reply (MODE_REPLY_MSG set)
+	// Mode                            REPLY_MSG only    REPLY_REFERENCE
+	RsGxsGroupId   mRefGroupId;   //   PARENT_GrpId      REPLY_GrpId
```
csoler, line 116: Does that mean that when you receive a  response to a tweet, the parent group of that tweet (meaning the twitter account?) needs to be subscribed to in order to see the msg. Consequently, all tweets in this new group that respond to other groups will in turn require to sync the other referred groups. I have the feeling that when you subscribe to a single group you'll recursively subscribe to everyone's.

libretroshare/src/retroshare/rswire.h
```diff
@@ -73,26 +69,63 @@ class RsWirePlace
...
+ * - NB: This Msg requires Parent Msg for complete info, while other two are self-contained.
+ ***********************************************************************/
+
+#define  WIRE_PULSE_TYPE_ORIGINAL_MSG        (0x0001)
+#define  WIRE_PULSE_TYPE_REPLY_MSG           (0x0002)
+#define  WIRE_PULSE_TYPE_REPLY_REFERENCE     (0x0004)
+
+#define  WIRE_PULSE_TYPE_SENTIMENT_POSITIVE  (0x0010)
+#define  WIRE_PULSE_TYPE_SENTIMENT_NEUTRAL   (0x0020)
+#define  WIRE_PULSE_TYPE_SENTIMENT_NEGATIVE  (0x0040)
```
sehraf, line 101: Did you consider to use classed enums instead of yet another bunch of ints? see #1785

csoler:
I see that the UI allows to display posts in a time range. This makes me think that we need some mechanism in GXS data access to only request posts in a given time range. This is not possible at this time.

unsenep2pdev:
We have implement gxs pagination on our branch. You can integrate into your branch manually. It won't be compatible for auto merge.

https://github.com/unsenep2pdev/RetroShare/commit/eedeb181a103d6cb0ff2fbc02ef07f88711c942a

drbob:
compilation should be fixed... sorry, I didn't commit my retroshare-gui.pro changes, but forgot I'd added files.

libretroshare/src/retroshare/rswire.h
drbob: No - I've tried to design it to avoid this scenario.

A reply tweet, includes the information about the parent tweet. so you don't need to subscribe (or even have access to parent group/msgs) to understand the context of the tweet. You'll likely need to be subscribed to a group to reply though - as this is a gxs limitation.

Additionally, when replying a 2nd REF child msg is posted to the parent group/msg so that the original poster can track replies without having to subscribe to them either.

If you want to traverse further up or down the reply tree, then you'll need to subscribe to additional groups... But I've tried to design it so that users won't have to subscribe to thousands of groups.

I don't know how its going to work in practice, we'll have to run a trial and see!

libretroshare/src/retroshare/rswire.h
drbob: I can change this. Do Enums work well for bitwise flags?

drbob:
> I see that the UI allows to display posts in a time range. This makes me think that we need some mechanism in GXS data access to only request posts in a given time range. This is not possible at this time.

The TokenReqOpts Interface include time ranges in them, though I don't think they are implemented in GXS at the moment. This would be a good use-case to use them. Currently that part of TheWire UI is not used.

drbob:
> We have implement gxs pagination on our branch. You can integrate into your branch manually. It won't be compatible for auto merge.
> 
> [unsenep2pdev@eedeb18](https://github.com/unsenep2pdev/RetroShare/commit/eedeb181a103d6cb0ff2fbc02ef07f88711c942a)

Interesting to see pagination in GXS Data requests. Its a good idea.

I can see you sorted by timestamp. This is useful in its own right, and should be added. Are they any other sorting orders would be useful?

If you are adding page#, then you should probably also add  "items_per_page" parameter too, so we are not limited to 25 per page. 

If you want to make this generic: One thing you might not have thought about is the additional filtering that happens after the database request. You request 25 items, and these could be filtered down to 5, or even Zero, before the data is returned.  This might confuse the user of the interface, as often you read pages until zero is returned.  You might need to return a flag to indicate there is more data.

libretroshare/src/retroshare/rswire.h
sehraf: Yes, there is the `RS_REGISTER_ENUM_FLAGS_TYPE()` macro that was made for exactly that use case. 
Beside having type safety, you additionally need `!!` for bool conversion when doing something like `if (!!(relayMode & RsDhtRelayMode::ENABLED))`

For reference see https://github.com/RetroShare/RetroShare/blob/master/libretroshare/src/retroshare/rsflags.h#L34 or 1b29e81e899506b4a8a87e2325e35b7a14899a82 or acda7b8cc4ce66c339b0e6f933dff67e29fba252

PR #1817, opened 2020-03-01
Add option to disable DH init check for faster init

G10h4ck:
The check is not necessart and on Android it takes around 1 minute to
  complete.

### DIFF
--- a/libretroshare/src/pqi/authssl.cc
+++ b/libretroshare/src/pqi/authssl.cc
@@ -404,10 +404,16 @@ int AuthSSLimpl::InitAuth(
 
 		std::cout.flush() ;
 
+#ifndef RS_DISABLE_DIFFIE_HELLMAN_INIT_CHECK
 		if(DH_check(dh, &codes) && codes == 0)
-			SSL_CTX_set_tmp_dh(sslctx, dh);	
+			SSL_CTX_set_tmp_dh(sslctx, dh);
 		else
-			pfs_enabled = false ;
+			pfs_enabled = false;
+#else // ndef RS_DISABLE_DIFFIE_HELLMAN_INIT_CHECK
+		/* DH_check(...) is not strictly necessary and on Android devices it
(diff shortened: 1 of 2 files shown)


libretroshare/src/pqi/authssl.cc
```diff
@@ -404,10 +404,16 @@ int AuthSSLimpl::InitAuth(
 
 		std::cout.flush() ;
 
+#ifdef RS_DIFFIE_HELLMAN_INIT_CHECK
```
csoler: i would prefer that the define line is #ifdef SKIP_DH_INIT_CHECK, so that the default is to check when no options are enabled. Then on android only, you add the option to have this define enabled. This way, unless we explicitly want it, the check is always performed.

G10h4ck:
Look at `retroshare.pri` that is enabled by default on all platform also on Android, I just made it configurable at build time ;)

csoler:
> Look at `retroshare.pri` that is enabled by default on all platform also on Android, I just made it configurable at build time ;)

Yes, and that's what I'm saying: for now, in order to disable this check, you need to explicitely include rs_dh_init_check. If you dont, then neither rs_dh_init_check nor no_rs_dh_init_check will be here, and the check will not happen. So I do not call this the default behavior. 
The default behavior should be that when nothing's specified (could be for many reasons, including a qmake mixup, or a commit error that somehow disables this rs_dh_init_check flag that is normally here) the check happens. 

With the flag system that is here, the way to do it is to do:
no_rs_dh_init_check {
    DISABLE_DH_INIT_CHECK = 1
}

then in the code:

#ifdef DISABLE_DH_INIT_CHECK
   enable_pfs without checking;
#else
   check_pfs and enable it if ok
#endif

G10h4ck:
If nothing specified in qmake command line the check was performed, anyway I edited it so now it is more explicit as you suggested ;)

PR #1820, opened 2020-03-03
Fixing settings pages

defnax:
* Set back show the right tabs by default
* Added for Files Tabs it taked to much space before


csoler:
Having two tabs is good. 
I think you should keep the text area that explains about share flags, in the first tab, since it speaks about directories

defnax:
> Having two tabs is good.
> I think you should keep the text area that explains about share flags, in the first tab, since it speaks about directories

yes it taked to much space now settings takes less space then before on every settings page looks now better, 

ok i will look maybe i doesnt checked if it was for the shared directories

PR #1826, opened 2020-03-07
attempt to solve GXS data access problems in new blocking API

csoler:


libretroshare/src/services/p3gxscircles.cc
```diff
@@ -278,8 +278,9 @@ bool p3GxsCircles::getCirclesSummaries(std::list<RsGroupMetaData>& circles)
 	uint32_t token;
 	RsTokReqOptions opts;
 	opts.mReqType = GXS_REQUEST_TYPE_GROUP_META;
-	if( !requestGroupInfo(token, opts)
-	        || waitToken(token) != RsTokenService::COMPLETE ) return false;
+	if( !requestGroupInfo(token, opts) || waitToken(token) != RsTokenService::COMPLETE )
+        cancelRequest(token);
```
G10h4ck: AFAIU due to the change in rsgxsifacehelper.h:304 this is not needed anymore

libretroshare/src/services/p3gxscircles.cc
```diff
@@ -289,8 +290,9 @@ bool p3GxsCircles::getCirclesInfo( const std::list<RsGxsGroupId>& circlesIds,
 	uint32_t token;
 	RsTokReqOptions opts;
 	opts.mReqType = GXS_REQUEST_TYPE_GROUP_DATA;
-	if( !requestGroupInfo(token, opts, circlesIds)
-	        || waitToken(token) != RsTokenService::COMPLETE ) return false;
+	if( !requestGroupInfo(token, opts, circlesIds) || waitToken(token) != RsTokenService::COMPLETE )
```
G10h4ck, line 296: And this seems to only be white space changes

G10h4ck (changes requested):

libretroshare/src/services/p3gxscircles.cc
csoler: I fixed it.

PR #1829, opened 2020-03-12
RsFiles links support in libretroshare + a bunch of fixes

G10h4ck:


felisucoibi:
+1

libretroshare/src/util/radix64.h
```diff
@@ -26,7 +26,10 @@
 #include <vector>
 #include <stdint.h>
 
-class Radix64
+#include "util/rsdeprecate.h"
```
csoler, line 29: So, you're copying all the code in radix64.h, do just a few modifications and change the authors to you and Altermundi? Honnestly it's not acceptable. Please keep the original authors.

libretroshare/src/util/rsmemory.h
```diff
@@ -90,6 +89,16 @@ bool rs_unique_cast(
 	return false;
 }
 
+/** Mark a pointer as non-owned aka you are not in charge of deleting it and
+ * must not delete it.
+ * @see http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1408r0.pdf */
+template<typename T> using rs_view_ptr = T*;
```
csoler, line 98: There's "const" for that. Isn't it sufficient?

csoler: So plz do not use this unless extremely totally really necessary. It makes the code much less readable. See rsurl for instance.

libretroshare/src/gossipdiscovery/gossipdiscoveryitems.h
```diff
@@ -84,17 +84,23 @@ class RsDiscPgpKeyItem: public RsDiscItem
...
 
-	RsDiscPgpKeyItem() : RsDiscItem(RsGossipDiscoveryItemType::PGP_CERT_BINARY)
+	RsDiscPgpKeyItem() :
+	    RsDiscItem(RsGossipDiscoveryItemType::PGP_CERT_BINARY),
+	    bin_data(nullptr), bin_len(0)
 	{ setPriorityLevel(QOS_PRIORITY_RS_DISC_PGP_CERT); }
 
-    virtual ~RsDiscPgpKeyItem() { delete[](bin_data);bin_data=nullptr;bin_len=0;}
-
+	~RsDiscPgpKeyItem() override { free(bin_data); }
```
csoler, line 92: you're sure about the free() / delete[] ?

libretroshare/src/services/rseventsservice.cc
```diff
@@ -100,40 +100,45 @@ RsEventsHandlerId_t RsEventsService::generateUniqueHandlerId_unlocked()
...
 
-    if( (int)eventType > mHandlerMaps.size() + 10)
-    {
-        RsErr() << "Cannot register an event handler for an event type larger than 10 plus the max pre-defined event (value passed was " << (int)eventType << " whereas max is " << (int)RsEventType::MAX << ")" << std::endl;
-        return false;
-    }
-
-    if( (int)eventType >= mHandlerMaps.size())
-        mHandlerMaps.resize( (int)eventType +1 );
+	if( eventType >= RsEventType::__MAX)
```
csoler: Originally I left 10 slots to allow people to register event types e.g. in plugins. If you make the comparison strict like this, plugins cannot use their own events anymore. So unless you have something special in mind to fix this, it's better to not do this particular change.

libretroshare/src/services/rseventsservice.cc
```diff
@@ -100,40 +100,45 @@ RsEventsHandlerId_t RsEventsService::generateUniqueHandlerId_unlocked()
...
+		RsErr() << __PRETTY_FUNCTION__ << " Invalid event type: "
+		        << static_cast<uint32_t>(eventType) << " >= RsEventType::__MAX:"
+		        << static_cast<uint32_t>(RsEventType::__MAX) << std::endl;
+		print_stacktrace();
+		return false;
+	}
 
-	if(!hId)
-        hId = generateUniqueHandlerId_unlocked();
+	if(!hId) hId = generateUniqueHandlerId_unlocked();
```
csoler, line 112: same here.

libretroshare/src/services/rseventsservice.cc
```diff
@@ -174,27 +179,29 @@ void RsEventsService::threadTick()
...
-			return;
-		}
-
-        // Call all clients that registered a callback for this event type
-
-        for(auto cbit: mHandlerMaps[event_type_index])
-            cbit.second(event);
-
-        // Also call all clients that registered with NONE, meaning that they expect all events
+		RsErr() << __PRETTY_FUNCTION__ << " Invalid event type: "
```
csoler: Same here. Use mHandlerMaps.size() instead of RsEventType::__MAX !!

libretroshare/src/pgp/pgphandler.cc
```diff
@@ -630,50 +630,61 @@ std::string PGPHandler::SaveCertificateToString(const RsPgpId& id,bool include_s
 	return makeRadixEncodedPGPKey(key,include_signatures) ;
 }
 
-bool PGPHandler::exportPublicKey(const RsPgpId& id,unsigned char *& mem_block,size_t& mem_size,bool armoured,bool include_signatures) const
+bool PGPHandler::exportPublicKey(
+        const RsPgpId& id,
+        rs_view_ptr<unsigned char>& mem_block, size_t& mem_size,
```
csoler: So, before, it was clear that the caller has memory ownership of the pointer since it is passed as variable.
Now, I have no idea at all. I would prefer that we keep *not using* shared_ptr and all sorts of C++11 goodies in low level code parts. We need to be very explicit of who deletes the memory and who allocates it. This was the case before this change.

libretroshare/src/pgp/pgphandler.h
```diff
@@ -107,7 +107,9 @@ class PGPHandler
 		bool LoadCertificateFromBinaryData(const unsigned char *bin_data,uint32_t bin_data_len, RsPgpId& gpg_id, std::string& error_string);
 
 		std::string SaveCertificateToString(const RsPgpId& id,bool include_signatures) const ;
-		bool exportPublicKey(const RsPgpId& id,unsigned char *& mem,size_t& mem_size,bool armoured,bool include_signatures) const ;
+		bool exportPublicKey( const RsPgpId& id,
```
csoler, line 112: See my comment in the .cc: low level code => no fancy pointer things. You have no idea how these things behave on different systems, in particular on windows. Debugging memory issues with this would be a total nightmare.

libretroshare/src/retroshare/rsgxschannels.h
```diff
@@ -504,8 +465,62 @@ class RsGxsChannels: public RsGxsIfaceHelper, public RsGxsCommentService
...
+	/**
+	 * @brief Get auto-download option value for given channel
+	 * @jsonapi{development}
+	 * @deprecated This feature rely on very buggy code, the returned value is
+	 *	not reliable @see setChannelAutoDownload().
+	 * @param[in] channelId channel id
+	 * @param[out] enabled storage for the auto-download option value
+	 * @return false if something failed, true otherwhise
+	 */
+	RS_DEPRECATED
```
csoler, line 503: I dont understand why you'd deprecate this. It's used and useful at the same time.

libretroshare/src/serialiser/rsserializer.cc
```diff
@@ -121,58 +125,52 @@ RsItem *RsConfigSerializer::deserialise(void *data, uint32_t *size)
...
+		RsErr() << fName << " " << ec << std::endl;
+		print_stacktrace();
+		return false;
+	};
 
-	if(tlvsize > *size)
-		throw std::runtime_error("Cannot serialise: not enough room.") ;
+	if(tlvsize > *size) return failure(std::errc::no_buffer_space);
 
+	SerializeContext ctx(static_cast<uint8_t*>(data), tlvsize, mFlags);
```
csoler, line 143: This works but only because you changed the constructor of SerializeContext. Took me a while to figure out.

libretroshare/src/serialiser/rsserializer.h
```diff
@@ -200,63 +233,28 @@ struct RsGenericSerializer : RsSerialType
 		SIZE_ESTIMATE = 0x01,
 		SERIALIZE     = 0x02,
 		DESERIALIZE   = 0x03,
-		PRINT         = 0x04,
+		PRINT         = 0x04, /// @deprecated use rsdebug.h << operator instead
 		TO_JSON,
 		FROM_JSON
 	} SerializeJob;
 
-
```
csoler, line 208: I dont understand the logic behind removing the serialize format. Up to me it's really needed: it tells the seriaizer what you need. How do you do that now??

libretroshare/src/util/rsdir.cc
```diff
@@ -516,6 +528,23 @@ bool	RsDirUtil::checkCreateDirectory(const std::string& dir)
 	return true;
 }
 
+#if __cplusplus < 201703L
+bool std::filesystem::create_directories(const std::string& path)
+{
+	for( std::string::size_type lastIndex = 0; lastIndex < std::string::npos;
```
csoler, line 534: This loop works because you increase lastIndex inside the loop. Otherwise after path[lastIndex]=='/', the call to path.find('/',lastIndex) will always return lastIndex.

On windws if path is "C:/users/someguy" for instance, then it will try to create directory C:/ in the current directory. Does windows interpret that correctly?
On linux however, if path is "some/path/" then it will try to create directory "some/" in current dir. Which is not what you want. You can to create directory "some". Maybe checkCreateDirectory takes care of it, but it would be safer to drop the "/" of curDir when it's the last character.

retroshare-gui/src/gui/common/RsCollection.cpp
```diff
@@ -273,14 +273,14 @@ void RsCollection::recursAddElements(QDomDocument& doc,const ColFileInfo& colFil
...
 }
 
-void RsCollection::recursAddElements(QDomDocument& doc,const FileTree& ft,uint32_t index,QDomElement& e) const
+void RsCollection::recursAddElements(
+        QDomDocument& doc, const RsFileTree& ft, uint32_t index,
+        QDomElement& e ) const
 {
-	std::vector<uint32_t> subdirs ;
-	std::vector<FileTree::FileData> subfiles ;
+	std::vector<std::uintptr_t> subdirs;
```
csoler: normally subdirs is an array of indexes. So it has no chance to be a uintptr_t. It should be a uint32_t.

libretroshare/src/file_sharing/file_tree.cc
```diff
@@ -97,7 +159,9 @@ void FileTreeImpl::recurs_buildFileTree(FileTreeImpl& ft,uint32_t index,const Di
 			std::cerr << "(EE) Cannot request dir details for pointer " << dd.children[i].ref << std::endl;
 }
 
-bool FileTreeImpl::getDirectoryContent(uint32_t index,std::string& name,std::vector<uint32_t>& subdirs,std::vector<FileData>& subfiles) const
+bool RsFileTree::getDirectoryContent(
+        std::string& name, std::vector<std::uintptr_t>& subdirs,
+        std::vector<FileData>& subfiles, std::uintptr_t index ) const
 {
 	if(index >= mDirs.size())
```
csoler: that is where you see that using a std::uintptr_t is inconsistent. index should be a uint32_t.

libretroshare/src/ft/ftserver.cc
```diff
@@ -1872,15 +1975,17 @@ void ftServer::ftReceiveSearchResult(RsTurtleFTSearchResultItem *item)
 			hasCallback = true;
 
 			std::vector<TurtleFileInfoV2> cRes;
-			for( const auto& tfiold : item->result)
-				cRes.push_back(tfiold);
+			std::transform( item->result.begin(), item->result.end(),
```
csoler: That would be acceptable if the code was simpler. But no, it really is not. Dont you think?

libretroshare/src/ft/ftserver.cc
```diff
@@ -2101,3 +2206,144 @@ RsFileItem::RsFileItem(RsFileItemType subtype) :
...
+	switch(static_cast<RsFilesErrorNum>(ev))
+	{
+	case RsFilesErrorNum::FILES_HANDLE_NOT_FOUND:
+		return std::errc::invalid_argument;
+	default:
+		return std::error_condition(ev, *this);
+	}
+}
+
+std::error_condition ftServer::dirDetailsToLink(
```
csoler, line 2222: That kind of transform is highly breakable. What happens if dirDetails.name is "fileData=something". I can think of 100 different ways to make the parsing fail, whatever you might have made it bullet proof already.
The question is: do you actually need to parse urls that represent dirDetails somewhere else than inside RS? I would be surprised. If you do, then you need to encode the data properly so that it doesn't fail with whatever utf8 we might encounter, etc.

libretroshare/src/retroshare/rsgxschannels.h
```diff
@@ -282,26 +283,6 @@ class RsGxsChannels: public RsGxsIfaceHelper, public RsGxsCommentService
 	 */
 	virtual bool ExtraFileRemove(const RsFileHash& hash) = 0;
 
-	/**
-	 * @brief Get auto-download option value for given channel
-	 * @jsonapi{development}
```
csoler, line 288: Why do you remove these? Aren't they used in GUI ? I believe they are.

libretroshare/src/serialiser/rstypeserializer.h
```diff
@@ -163,19 +198,24 @@ struct RsTypeSerializer
 			ctx.mOk = ctx.mOk && to_JSON(member_name, member, ctx.mJson);
 			break;
 		case RsGenericSerializer::FROM_JSON:
-			ctx.mOk &= (ctx.mOk || ctx.mFlags & RsGenericSerializer::SERIALIZATION_FLAG_YIELDING)
+			ctx.mOk &= ( ctx.mOk ||
```
csoler, line 201: nice catch. The previous version was ambiguous. I'm not even sure it was correct.

libretroshare/src/serialiser/rstypeserializer.h
```diff
@@ -439,176 +454,222 @@ struct RsTypeSerializer
...
-				          << memberName << "\"" << std::endl;
+			uint32_t elCount = 0;
+			RS_SERIAL_PROCESS(elCount);
+			if(!ctx.mOk) break;
+
+			/* This check is not perfect but will catch most pathological cases.
+			 * Avoid multiplying by sizeof(el_t) as it is not a good exitimation
+			 * of the actual serialized size, depending on the elements
+			 * structure and on the so it would raises many false positives. */
+			if(elCount > ctx.mSize - ctx.mOffset)
```
csoler: since all these things are uint32_t, it's much safer to do 
   if(elCount + ctx.mOffset > ctx.mSize)
debugging bugs caused by overloading of uint32_t is pretty painful.

libretroshare/src/serialiser/rstypeserializer.h
```diff
@@ -439,176 +454,222 @@ struct RsTypeSerializer
...
-	/// std::vector<T>
+//============================================================================//
+//                       Sequence containers                                  //
+//============================================================================//
+
+	/** std::list is supported */ template <typename... Args>
+	struct is_sequence_container<std::list<Args...>>: std::true_type {};
+
+	/** std::set is supported */ template <typename... Args>
+	struct is_sequence_container<std::set<Args...>>: std::true_type {};
```
csoler, line 465: When you deserialize using push_back, it's not the same than using std::set::insert. It only works if the items are already sorted, right? It's easy to make a rogue peer that will send unsorted items and create bugs at its friends, or even later on if , std::set is used in some GXS items.

libretroshare/src/serialiser/rstypeserializer.h
```diff
@@ -439,176 +454,222 @@ struct RsTypeSerializer
...
+//                               Strings                                      //
+//============================================================================//
+
+	/// Strings
 	template<typename T>
-	static void serial_process( RsGenericSerializer::SerializeJob j,
-	                            RsGenericSerializer::SerializeContext& ctx,
-	                            std::list<T>& v,
-	                            const std::string& memberName )
+	typename std::enable_if<is_string<T>::value>::type
```
csoler, line 606: Which other types than std::string are you targetting here?

libretroshare/src/serialiser/rstypeserializer.h
```diff
@@ -439,176 +454,222 @@ struct RsTypeSerializer
...
 		}
-		case RsGenericSerializer::DESERIALIZE:
+		case RsGenericSerializer::SERIALIZE:
 		{
-			uint32_t n=0;
-			serial_process(j,ctx,n,"temporary size");
-			for(uint32_t i=0;i<n;++i)
+			uint32_t len = static_cast<uint32_t>(member.length());
+			RS_SERIAL_PROCESS(len);
+			if(len > ctx.mSize - ctx.mOffset)
```
csoler: same remark here: use if(len + ctx.mOffset > ctx.mSize)

libretroshare/src/serialiser/rstypeserializer.h
```diff
@@ -439,176 +454,222 @@ struct RsTypeSerializer
...
-			if(v.empty())
-				std::cerr << "  Empty list \"" << memberName << "\""
-				          << std::endl;
-			else
-				std::cerr << "  List of " << v.size() << " elements: \""
-				          << memberName << "\"" << std::endl;
+			uint32_t len;
+			RS_SERIAL_PROCESS(len);
+			if(!ctx.mOk) break;
+			if(len > ctx.mSize - ctx.mOffset)
```
csoler: same here

libretroshare/src/serialiser/rstypeserializer.h
```diff
@@ -910,13 +945,190 @@ struct RsTypeSerializer
 	                       t_RsTlvList<TLV_CLASS,TLV_TYPE>& member,
 	                       RsJson& jDoc );
 
-	[[noreturn]] static void fatalUnknownSerialJob(int j)
+//============================================================================//
```
csoler, line 963: There was some bullet proof code in FileListIO named Read125Size and Write125Size. I remember fixing nasty bugs in it, so we need to be careful using some new code here. It would be safer to copy the code in FileListIO, put it here, and call it from FileListIO so we have one instance of that code only.

csoler (changes requested):
Pfff. Was a long long read. It would be very helpful if you could keep the PR for technical aspects (here serialization in VLQ mode) and make another PR for cosmetic changes. It took me 4 days to read all this in details.
Anyway, most of it is great. There's a few problems and things to fix (see my 25 remarks). Wont take you long.
Thx

libretroshare/src/serialiser/rsserializer.h
G10h4ck: That enum was never used, the serialization format is controlled by serial job, and in some cases changes depending on the flag (some members ar not serializaed if some flag is passed etc)

retroshare-gui/src/gui/common/RsCollection.cpp
G10h4ck: we used this to wrap the void* that is used internally in DirDetails, in another PR I am converting this stuff to uint64_t, AFAIR that void* was used for some reason tomake it easier for rs-gui is it sitll the case? If not I would get rid of all this trick types and use uint64_t only see https://github.com/RetroShare/RetroShare/pull/1844

libretroshare/src/file_sharing/file_tree.cc
G10h4ck: I agree to not use std::uintptr_t due to it's lenght changing depending on architecture, see my previous comment

libretroshare/src/ft/ftserver.cc
G10h4ck: the other version doesn't compile on all platforms (aka Android), because of that I changed it, moreover std::trasform suggest to the compile to use SIMD instructions when available AFAIR

libretroshare/src/ft/ftserver.cc
G10h4ck: It encode fields properly following the RFC standard, so special characters won't break it. Also in case parsing fails it just return an error and no more fuss ;)
BTW the relevant data is encoded in base64url format, the URL is just an disguise to encrust the data into in a way that suggests other application to make it clickable, it is important to do it at libretroshare level so all application using libretroshare have compatible links formats, without becoming crazy on making different URL libraries to behave similarly (they usually don't except for http[s] links that are the most common and tested). We already do this for channels, forums etc

libretroshare/src/retroshare/rsgxschannels.h
G10h4ck: they are, I have just deprecated them because they use the "service string" internally in a  terrible manner, and when you enable those features retroshare start acting weird, flooding the log with error messages etc, a few months ago I also investigated if it was possible to fix that functionality, but I discovered the terrible truth...

libretroshare/src/serialiser/rstypeserializer.h
G10h4ck: I read that code, the idea is similar but this implementation is more generic, the 125 version can handle integer up to uint32, while this can handle any integer size, also it is efficient handling signed integers

libretroshare/src/util/radix64.h
G10h4ck: Sure I'll add original authors too!

libretroshare/src/util/rsmemory.h
G10h4ck: nope, it is different concept, const imply the pointer or the pointed data (depending on where const is put) cannot be modified, while this "decoration" just make explicit that owner of the pointer is not transferred, having no implication on constness, also this doesn't use any C++11 feature it just make explicit ownership is passed to the reader, but for the compiler is the exactly the same then *

libretroshare/src/gossipdiscovery/gossipdiscoveryitems.h
G10h4ck: Yeah I made the change coherent, we use *alloc/free for raw memory chunks while new/delete for C++ classes, and in fact this one was an incoherent case that caused problems becuase all the rest of the code assumes *alloc/free is used for raw memory chunks

libretroshare/src/pgp/pgphandler.cc
G10h4ck: Nope before the pointer ownership wasn't clear, the way it was passed made explicit that `exportPublicKey` could modify the pointer address but not who should free that memory, in theory that pointer could have been a pointer to some internal structure that was passed just for reading modifying but not freeing to the caller, `rs_view_ptr<T>` for the compiler is the same as `T*` as you can see where it is declared, it doesn't use any C++11 smart pointer feature, it is just a "decoration" to make it explicit to the reader that the ownership of the pointer is not transferred, just by reading the function signature without need of reading the implementation.

libretroshare/src/pgp/pgphandler.h
G10h4ck: those are not smart pointers they are just reader "decorations" to make explicit if ownership is passed or not

libretroshare/src/retroshare/rsgxschannels.h
G10h4ck: see my comment up

libretroshare/src/services/rseventsservice.cc
G10h4ck: ok got the point, I'll also add a comment explaining why it make sense to have those 10 extra slots, BTW plugins should have their event type declared in the enum in the long term, but it make sense to have that extra slots so a new plugin may be registered on latest release before the plugin was available

libretroshare/src/serialiser/rstypeserializer.h
G10h4ck: ATM we have only std::string but the idea is that this code is generic enough to support other kind of strings implementation, think of std::basic_string<char, Our_custom_allocator> which is very similar to std::string except for using a specialized allocator

libretroshare/src/serialiser/rstypeserializer.h
G10h4ck: The elements are inserted with this `member.insert(member.end(), elem);` which on `std::vector`, `std::list` etc. is equivalent to `push_back(elem)` while on `std::set` which keep it's own sorting internally and check if the element is not already present it is equivalent to `instert(elem)`. So no problem with ~friends~ which send things out of order ;)

libretroshare/src/serialiser/rstypeserializer.h
G10h4ck: While we control both `ctx.mSize` and `ctx.mOffset` and it i always true that `ctx.mOffset <= ctx.mSize` so  `ctx.mSize - ctx.mOffset >= 0` and then it is safe to do `ctx.mSize - ctx.mOffset`.  Instead we don't have control over `elCount` which is sent by the peer which could be bugged and send us a big value that would cause `(elCount + ctx.mOffset) > std::numeric_limits<uint32_t>::max()` that would cause an undetectable overflow

G10h4ck: eventually we can put in the code some runtime check `assert(ctx.mOffset <= ctx.mSize)` but that should be not necessary as we control those values, in a way that that condition is always true.

libretroshare/src/util/rsmemory.h
csoler: So you're telling me only difference is for the readerwhen he reads the code, to know who has the memory ownership?

libretroshare/src/pgp/pgphandler.cc
csoler: Now it is clear to you. But it is not clear to anybody else including me. Using a proper comment above the method prototype would be much more useful IMHO.

libretroshare/src/ft/ftserver.cc
csoler: It doesn't compile because of C++11? There's nothing fancy there except the for loop that could be written with an explicit iterator

libretroshare/src/serialiser/rstypeserializer.h
csoler: First if you put an assert you offer the possibility to crash friend's nodes and even distant nodes by sending crafted items.
Then, everywhere else in the code, we pay attention to not computing the difference between unsigned ints because some compilers do not handle this correctly in all cases.
Finally, moving one term the the left part is not a terrible change to do as compared to what it can prevent. 
So I really really do not understand your reaction here.

libretroshare/src/serialiser/rstypeserializer.h
csoler: ok. Because 125 code is used in crypto parts too, I would not replace it by this new code that is too high level, and keep your new code where it is.

libretroshare/src/serialiser/rstypeserializer.h
csoler: also when you say "we have control on...", it's not a valid argument: bugs happen. And their effect should be limited. That is why we make methods bullet-proof on an individual basis. This is what makes a low level lib such as libretroshare stable.

libretroshare/src/util/rsmemory.h
G10h4ck: Exactly, right now is like that. But since this is not a /* comment */ but something that the compiler can parse we could eventually enforce more checks at compile time in future.

libretroshare/src/ft/ftserver.cc
G10h4ck: No it is not a C++11 issue, It's because of an implicit conversion that the Android compiler doesn't like, so as I had to rewrite it I have done the way the compiler error message suggested to do.

libretroshare/src/serialiser/rstypeserializer.h
G10h4ck: The peer have no way to craft `ctx.mSize` and `ctx.mOffset` while it can craft `elCount` because of that it is safer doing arithmetic operations on `ctx.mSize` and `ctx.mOffset` while it is not safe to do arithmetic operations on `elCount`. The runtime handle unsigned integers perfectly provided that the result of operation is >= 0 and <= std::numeric_limits<uint32_t>::max() it make no difference for it if we use - or + provided that the result is in that range. In this case we can guarantee that condition when doing `ctx.mSize - ctx.mOffset` (unless we have a bug) while we cannot guarantee it when calculating `elCount + ctx.mOffset` because we have no control over what value of `elCount` the peer sent to us. So in this case `ctx.mSize - ctx.mOffset` is safer then `elCount + ctx.mOffset`. I was trying to argument this, no reaction ;)

libretroshare/src/services/rseventsservice.cc
G10h4ck: done

libretroshare/src/jsonapi/jsonapi.cpp
```diff
@@ -395,7 +395,7 @@ JsonApiServer::JsonApiServer(): configMutex("JsonApiServer config"),
 				} );
 			};
 
-			bool retval = rsEvents->registerEventsHandler(
+			auto retval = rsEvents->registerEventsHandler(
```
csoler: This is typically a place where auto is not good: it's just as long as the type you're hiding, but it keeps you from seeing what the type should be. That's what I'm talking about when I say that sometimes you use C++11 for the sake of using C++11. Here it's really a weird idea.

G10h4ck: I changed it because it returns std::error_condition now, but I can make it explicit if you like

libretroshare/src/util/rsdir.cc
G10h4ck: AFAIU `RsDirUtil::checkCreateDirectory` operate correctly on absolute paths so I expect that when it is passed "C:/" it does nothing because it already exists like it does when it is passed "/" on Linux, anyway it would be interesting to have it tested on Windows but I haven't a machine with that, but sure if doesn't work on windows we will get plenty of rants once merged :p . About the case of "some/path/" if it is passed like that I do expect that "$PWD/some/" is created if not already existing and then the same with "$PWD/some/path/"

G10h4ck: `create_directories` function is basically the equivalent of `mkdir -p` on your shell

libretroshare/src/pgp/pgphandler.cc
csoler: Even though there is a problem here: you cannot make theh caller of that funciton not responsible from deleting that pointer. Who will delete it otherwise? Doesn't make sense.

So mem_block should be allocated using malloc (as you do) and not with new[], but deleted by the client. 
Besides, the pointer being passed as a reference, it is clear even without comment that it is to be created in the function and therefore left to the client to delete. But anyway, a simple comment above the function prototype could recall that.

csoler: clearly here the use of fancy decoration shadowed where the real memory management should take place, even for you. That's proof that it's not for the best here at least.

libretroshare/src/util/rsdir.cc
csoler: you're probably right because "/" is forbidden in directory names.

retroshare-gui/src/gui/common/RsCollection.cpp
csoler: no really this is both  wrong and misleading. Should be uint32_t. We're talking about indexes in an array here, not pointers.

libretroshare/src/file_sharing/file_tree.cc
csoler: apparently it's still std::uintptr_t in the pr. Maybe you forgot to revert this whole set of changes.

libretroshare/src/ft/ftserver.cc
csoler: come on! This should work:
for(auto& tfiold: item->result)
    cRes.push_back(TurtleFileInfoV2(tfiold));
Implicit conversion is not accepted because the constructor is explicit.

libretroshare/src/retroshare/rsgxschannels.h
csoler: well then it's better to put a comment about why it's deprecated. Because now it looks like there's a replacement but there isn't.

libretroshare/src/serialiser/rstypeserializer.h
csoler: thx for the change.

csoler:
Thx for all the changes. There's still a problem: you used rs_view_ptr in a place where the pointer's ownership belongs to the calling function, in pgphandler.cc:exportPublicKey(). 
As I said already, it would be much more clear to 
- use unsigned char*&   as parameter. At least we all know what it means, and
- comment in the .h that the pointer needs to be deleted using malloc.
Replacing that comment with a fancy type that is defined elsewhere is of no benefit here. IMHO it's just wasting time of the devs who will have to find out every time what that name actually means. And even though you didn't comment, passing the pointer as variable without const implicitly means that it's an out parameter that will be allocated in the function and that should be deleted outside.

PR #1830, opened 2020-03-12
Removing some default stylesheet froms Messages & ChatWidget & icon changes

defnax:
* Removed some default stylesheets from Messages & ChatWidget
* removed unnecessary stylesheets which have issues on some systems with dark styles


felisucoibi:
+1

PR #1831, opened 2020-03-13
TheWire fixup up display of messages

drbob:
Process messages to group replies correctly and sort by age.
Expand message display to include replies.
Add coloring to distinguish between replies and actual msgs.
Enable subscribe / unsubscribe buttons.
Add Selection of which set of groups are visible.
Add missing setNetworkExchangeService calls for photo + wire services.


defnax:
hi i get compile error on master not sure there is something missed to fix?
i has updated my branch with master then the photos compile broken

image

drbob:
This has now been updated to compile and run with master

PR #1832, opened 2020-03-13
i2pbob: fix deadlock on shutdown

sehraf:
I've always wondered why the thread **somtimes** doesn't exit. gdb told me today...
Add a break to exit the loop when it should stop.


sehraf:
Thanks for merging, when i find some spare time i'll add a proper timeout. 
Nevertheless the thread should stop ASAPish when requested.

PR #1840, opened 2020-04-03
Fix json API enabled builds

sehraf:
Current master fails to build when enabling the json API.


G10h4ck:
I fixed this too in https://github.com/RetroShare/RetroShare/pull/1829

sehraf:
Feel free to ignore this pr then. Latest Master doesn't build anyway as cyril messed something else up with the latest merge.

PR #1844, opened 2020-04-04
Solve non compatible 32/64bit file handle in JSON API

G10h4ck:
Discovered testing on Android as armv7 is quite common and 32bit


G10h4ck:
Included in https://github.com/RetroShare/RetroShare/pull/1829

PR #1845, opened 2020-04-04
Removed TokenQueue from GxsTransportStatistics

csoler:


csoler:
I removed the WIP tag since this PR is stable. I originally wanted to continue on it but it can be merged already, to the benefit of 1847

PR #1859, opened 2020-04-15
Making channels faster to load

csoler:
[x] step one: avoid creating complex widgets that we do not see.
[  ] step two: re-enable the search (widgets not loaded won't show up)


felisucoibi:
good, as i said 2 years ago channel loading is crap, for only 100 posts it gets stucked, or loading.... it should handle 10.000 posts on the fly.

csoler:
well, it you could report how this PR feels, it would be useful.

felisucoibi:
I will this week i will test everything.

felisucoibi:
I have to admin things are g etting better, if you cand do the same magic to forums and posts loading i will move from CRAP to USABLE  again FORUMS and CHANNELS, btw, is fixed the bug of channels loading that mix betqween channels? I wii see if this happens or is fixed as as side effect. Bests

PR #1865, opened 2020-04-18
Add Ability to undock a chat lobby to a window.

PhenomRetroShare:


csoler:
Sorry defax, I merged it and then I saw that you where in the review requests. For me it's all good. Of course there might be some additional design fixes that you'd like to apply here?

PR #1866, opened 2020-04-19
Fix SIGSEGV on RsThread::start

PhenomRetroShare:
mTid could be null while pError is null too.

This happens to me always with "gxs integrity".

### DIFF
--- a/libretroshare/src/util/rsthreads.cc
+++ b/libretroshare/src/util/rsthreads.cc
@@ -194,6 +194,15 @@ bool RsThread::start(const std::string& threadName)
 			print_stacktrace();
 			return false;
 		}
+		if(!mTid)
+		{
+			RsErr() << __PRETTY_FUNCTION__ << " pthread_create could not create"
+			        << " new thread: " << threadName << " mTid: " << mTid
+			        << std::endl;
+			mHasStopped = true;
+			print_stacktrace();
+			return false;
+		}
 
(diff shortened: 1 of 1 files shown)


libretroshare/src/util/rsthreads.cc
```diff
@@ -194,6 +194,15 @@ bool RsThread::start(const std::string& threadName)
 			print_stacktrace();
 			return false;
 		}
+		if(!mTid)
```
csoler, line 197: Is this compatible with windows? mTid is of class pthread_t and while it's a uint64_t on linux, it is a struct on windows.

G10h4ck: Appveyor build on windows successfully

PhenomRetroShare: In win10 using MSys2:
@\libretroshare\src\util\rsthreads.h:276
	pthread_t mTid;
@C:\msys64\mingw32\i686-w64-mingw32\include\pthread.h:196
typedef uintptr_t pthread_t;
@C:\msys64\mingw32\i686-w64-mingw32\include\corecrt.h:82
typedef unsigned int uintptr_t;

So it is uint too...

csoler: ok. My fault. I mixed it up with something else.

PR #1869, opened 2020-04-22
Fix OpenSSL fail to create Cert.

PhenomRetroShare:


libretroshare/src/pqi/authssl.cc
```diff
@@ -804,10 +804,18 @@ X509 *AuthSSLimpl::SignX509ReqWithGPG(X509_REQ *req, long /*days*/)
         }
         X509_NAME_free(issuer_name);
 
-        // NEW code, set validity time between null and null
-        // (does not leak the key creation date to the outside anymore. for more privacy)
-        ASN1_TIME_set(X509_get_notBefore(x509), 0);
-        ASN1_TIME_set(X509_get_notAfter(x509), 0);
+	// NEW code, set validity time between 2010 and 2110 (remember to change it when...)
+	// (does not leak the key creation date to the outside anymore. for more privacy)
+	if (!ASN1_TIME_set_string(X509_getm_notBefore(x509), "20000101000000Z"))
```
zapek: I don't understand the comment. You say it sets the time between 2010 and 2110 but then set it to 1st of January 2000? And why remember to change it? RS never checks the validity of it.

PhenomRetroShare: Oups, my git under QtCreator under MSys2 failed to stash so I have to rewrite this. Too quickly...
The comment is a joke. ;)

PR #1878, opened 2020-04-26
[WIP] fixing some GUI bugs

csoler:


G10h4ck:
The problem is not in RsTypeSerializer::RawMemoryWrapper::serial_process, it crashes there because we have invalid or uninitialized pointers around! Changing that just hide the bugs that are elsewhere.

PR #1879, opened 2020-04-27
RawMemoryWrapper::serial_process report error

G10h4ck:


libretroshare/src/serialiser/rstypeserializer.cc
```diff
@@ -543,51 +543,56 @@ void RsTypeSerializer::RawMemoryWrapper::serial_process(
...
+			 * must be null and size 0 at this point */
+
+			RsErr() << __PRETTY_FUNCTION__ << " DESERIALIZE got uninitialized "
+			        << " or pre-allocated buffer! Buffer pointer: " << first
+			        << " must be null and size: " << second << " must be 0 at "
+			        << "this point. Does your item costructor initialize them "
+			        << "properly?" << std::endl;
+			print_stacktrace();
+			ctx.mOk = false;
+			break;
```
csoler: actually we can afford to remove the break here. It's not a problem if that happens for what we do in this function. It's just a warning that will help us fixing badly initialized memory.

G10h4ck: done

PR #1880, opened 2020-04-27
clean rewrite of RsServer::threadTick

jolavillette:
- switch all TS to double, no more int
- remove the useless "if" at the beginning of threadTick, that was always true because of the usleep!
- stabilize the tick rate by properly calculating the sleeping time
- adjust the tick rate depending on the load
- clean the code for conditionnal run every second, 5s, 60s and 1h

### DIFF
--- a/libretroshare/src/rsserver/p3face-server.cc
+++ b/libretroshare/src/rsserver/p3face-server.cc
@@ -73,14 +73,13 @@ static double getCurrentTS()
 // In some cases (VOIP) it's likely that we will need to set them temporarily to a very low 
 // value, in order to favor a fast feedback
 
-const double RsServer::minTimeDelta = 0.05; // 25;
-const double RsServer::maxTimeDelta = 0.2;
-const double RsServer::kickLimit = 0.15;
+const double RsServer::minTickInterval = 0.05;
+const double RsServer::maxTickInterval = 0.2;
 
 
 RsServer::RsServer() :
-    coreMutex("RsServer"), mShutdownCallback([](int){}),
-    coreReady(false)
+	coreMutex("RsServer"), mShutdownCallback([](int){}),
+	coreReady(false)
 {
 	{
 		RsEventsService* tmpRsEvtPtr = new RsEventsService();
@@ -108,21 +107,16 @@ RsServer::RsServer() :
 	msgSrv = NULL;
 	chatSrv = NULL;
 	mStatusSrv = NULL;
-    mGxsTunnels = NULL;
+	mGxsTunnels = NULL;
 
-    mMin = 0;
-    mLoop = 0;
+	mLastts = getCurrentTS();
+	mTickInterval = maxTickInterval ;
+	mAvgRunDuration = 0;
+	mLastRunDuration = 0;
 
+	/* caches (that need ticking) */
 
-    mLastts = getCurrentTS();
-    mLastSec = 0; /* for the slower ticked stuff */
-    mTimeDelta = 0.25 ;
-
-    mAvgTickRate = mTimeDelta;
-
-    /* caches (that need ticking) */
-
-	/* Config */
+	/* config */
 	mConfigMgr = NULL;
 	mGeneralConfig = NULL;
 }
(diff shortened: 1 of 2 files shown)


libretroshare/src/rsserver/p3face-server.cc
```diff
@@ -108,21 +107,16 @@ RsServer::RsServer() :
...
 	mStatusSrv = NULL;
-    mGxsTunnels = NULL;
+	mGxsTunnels = NULL;
 
-    mMin = 0;
-    mLoop = 0;
+	mLastts = getCurrentTS();
+	mTickInterval = maxTickInterval ;
+	mAvgRunDuration = 0;
+	mLastRunDuration = 0;
```
G10h4ck, line 116: At first glance I'd say it is better to initialize those in initializers list like `coreMutex` and `coreReady`

libretroshare/src/rsserver/p3face-server.cc
```diff
@@ -132,143 +126,101 @@ RsServer::~RsServer()
...
-    {
-#ifdef	DEBUG_TICK
-        std::cerr << "Delta: " << delta << std::endl;
-        std::cerr << "Time Delta: " << mTimeDelta << std::endl;
-        std::cerr << "Avg Tick Rate: " << mAvgTickRate << std::endl;
-#endif
-
-        mLastts = ts;
+RsDbg() << "DEBUG_TICK" << std::endl;
+RsDbg() << "DEBUG_TICK ticking interval "<< mTickInterval << std::endl;
```
G10h4ck: To not indent those lines is intentional? If so why?

libretroshare/src/rsserver/p3face.h
```diff
@@ -188,16 +188,13 @@ class RsServer: public RsControl, public RsTickingThread
...
 		// Worker Data.....
 
-    int mMin ;
-    int mLoop ;
-    int mLastts ;
-    long mLastSec ;
-    double mAvgTickRate ;
-    double mTimeDelta ;
+        double mLastts;
+        double mTickInterval;
```
G10h4ck: should use <TAB> for indentation instead of space, like the following two lines did you read this  https://retroshare.readthedocs.io/en/latest/developer/coding/  ?

libretroshare/src/rsserver/p3face.h
```diff
@@ -188,16 +188,13 @@ class RsServer: public RsControl, public RsTickingThread
...
+	double mLastRunDuration;
+	double mAvgRunDuration;
 
-    static const double minTimeDelta; // 25;
-    static const double maxTimeDelta;
-    static const double kickLimit;
+	static const double minTickInterval;
+	static const double maxTickInterval;
```
G10h4ck, line 198: you want probably use `constexpr` instead of `const` here

G10h4ck (commented):
Haven't tested it yet, but looks good also I have left a few minor comments inline

libretroshare/src/rsserver/p3face-server.cc
```diff
@@ -108,21 +106,20 @@ RsServer::RsServer() :
...
-    mLastSec = 0; /* for the slower ticked stuff */
-    mTimeDelta = 0.25 ;
+	mGxsTunnels = NULL;
 
-    mAvgTickRate = mTimeDelta;
+	mLastts = getCurrentTS();
+	mTickInterval = maxTickInterval ;
+	mAvgRunDuration = 0;
+	mLastRunDuration = 0;
+	mCycle1 = 0;
```
csoler, line 117: If these are doubles, it's better to initialize them with 0.0
It wont change anything except that when you read you know it's a double.

libretroshare/src/rsserver/p3face-server.cc
```diff
@@ -132,143 +129,132 @@ RsServer::~RsServer()
...
 
-    } // end of only once a second.
+#ifdef TICK_DEBUG
+	RsDbg() << "TICK_DEBUG new mLastRunDuration " << mLastRunDuration << " mAvgRunDuration " << mAvgRunDuration << std::endl;
+	if (mLastRunDuration > WARN_BIG_CYCLE_TIME)
+		RsDbg() << "TICK_DEBUG excessively long lycle time " << mLastRunDuration << std::endl;
+#endif
+	
+// if the core has returned that there is more to tick we decrease the ticking interval, else we increse it
+// this should be studied closer as I dont think that the core ever returns 1
```
csoler, line 241: it probably does only on heavy load.

PR #1881, opened 2020-04-27
Add Votes when calling getPostData.

PhenomRetroShare:
Else, get "Not a GxsChannelPostItem neither a  RsGxsCommentItem" error.


defnax:
Hi, i hope we get a real Comments Notify for the News Feeds, like how for Forums (ForumMsgItem).
I doesnt like the “New Comment“ blank dummy Notify

PR #1882, opened 2020-04-27
Remove unmanaged compiler warnings.

PhenomRetroShare:


retroshare.pri
```diff
@@ -501,8 +501,16 @@ no_rs_deprecatedwarning {
 }
 
 no_rs_cppwarning {
+    QMAKE_CXXFLAGS += -Wno-unknown-warning-option
```
G10h4ck, line 504: Ok with this, while the others are concerning, I'd prefer we fix those other warnings and personally i take of them in the code i edit

G10h4ck (commented):
See inline comment

retroshare.pri
PhenomRetroShare: Me too but some not...
And now with QtCreator you have direct warning with clang while editing.
 This is only to show "important" warning not to lose in the crowd...
And we still can use "CONFIG+=rs_cppwarning".

retroshare.pri
```diff
@@ -501,8 +501,16 @@ no_rs_deprecatedwarning {
...
     QMAKE_CXXFLAGS += -Wno-cpp
     QMAKE_CXXFLAGS += -Wno-inconsistent-missing-override
+    QMAKE_CXXFLAGS += -Wno-misleading-indentation
+    QMAKE_CXXFLAGS += -Wno-unused-parameter
```
csoler, line 508: this one is very useful. Sometimes helps you detect some bugs.

csoler: Actually all warnings below should stay. They *are* useful.

PhenomRetroShare: So why you don't merge PR fixing it? ;)
For now all of them pollute compilation output.

PR #1884, opened 2020-04-27
Fix Channel Post Comments Number

PhenomRetroShare:


libretroshare/src/services/p3gxschannels.cc
```diff
@@ -1137,11 +1137,30 @@ bool p3GxsChannels::getChannelContent( const RsGxsGroupId& channelId,
...
 
 	return getPostData(token, posts, comments);
 }
 
+bool p3GxsChannels::getChannelComments(const RsGxsGroupId &channelId,
+                                       const std::set<RsGxsMessageId> &contentIds,
+                                       std::vector<RsGxsComment> &comments)
+{
+	std::vector<RsGxsGrpMsgIdPair> msgIds;
+	for (auto msg:contentIds)
```
csoler: seems to me that for(auto& msg:contentIds) would avoid a copy, right?

PhenomRetroShare: You're right.
I did not think about it.
Done

PR #1885, opened 2020-04-28
pqissl: document state machine and function hierarchy

sehraf:
While working on the code i got lost (sic!) and drew this map.
It's useful to keep track of things when overloading `pqissl`.
(Looks best with 4 space per tab)


csoler:
too bad we dont use the same codes to make arrows (See e.g. rsgxsnetservice.cc). Anyway, thx a lot for doing this!

sehraf:
I can change it to follow `rsgxsnetservice.cc`.
Did you use a tool for that?

PR #1894, opened 2020-04-30
[Mail] Set a default item height for the mail tree item to avoid shrinking

defnax:
* Set a default item height for the mail tree item to avoid shrinking under 20 pixel

Always the items on mail shrinks when From column contains no Avatar to avoid this i set a default height like how on the Friendslist.
Shrinking is gone not sure if this helps with the shrinking "Date" column.

Disabled in Messages Dialog
`    //msgwheader->resizeSection (RsMessageModel::COLUMN_THREAD_AUTHOR,     fm.width("[Retroshare]")*1.5);
    //msgwheader->resizeSection (RsMessageModel::COLUMN_THREAD_DATE,       fm.width("01/01/1970")*1.5);

this get conflicts with the defined values from RsMessageModel::sizeHintRole

### DIFF
--- a/retroshare-gui/src/gui/msgs/MessageModel.cpp
+++ b/retroshare-gui/src/gui/msgs/MessageModel.cpp
@@ -408,14 +408,17 @@ QVariant RsMessageModel::backgroundRole(const Rs::Msgs::MsgInfoSummary &fmpe, in
 
 QVariant RsMessageModel::sizeHintRole(int col) const
 {
-	float factor = QFontMetricsF(QApplication::font()).height()/14.0f ;
+	float x_factor = QFontMetricsF(QApplication::font()).height()/14.0f ;
+	float y_factor = QFontMetricsF(QApplication::font()).height()/14.0f ;
+	
+	y_factor = std::max(y_factor, 18.0f / 14.0f ); // allows to have a minimum 21 pixels default height
 
 	switch(col)
 	{
 	default:
-	case COLUMN_THREAD_SUBJECT:      return QVariant( QSize(factor * 170, factor*14 ));
-	case COLUMN_THREAD_DATE:         return QVariant( QSize(factor * 75 , factor*14 ));
-	case COLUMN_THREAD_AUTHOR:       return QVariant( QSize(factor * 75 , factor*14 ));
+	case COLUMN_THREAD_SUBJECT:      return QVariant( QSize(x_factor * 170 , y_factor*14*1.1f ));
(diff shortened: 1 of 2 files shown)


csoler:
yes, and this is exactly what you need to modify: sizeHintRole, not set a minimum size.

csoler:
btw, the shrink is not so bad. It allows to show the full avatar when there is very little room. Otherwise the avatar will be partly chopped off.

defnax:
> btw, the shrink is not so bad. It allows to show the full avatar when there is very little room. Otherwise the avatar will be partly chopped off.

shrinking looks so ugly you like that? ;) i has thinked you will fix this issue
i never like shrinking items every one look different size 
on email clients every item have same size and not shrinking

like the mail tree doesnt looks good at the moment since new Message model from 06/2019
image

defnax:
here the example code from FriendsListModel


> QVariant RsFriendListModel::sizeHintRole(const EntryIndex& e,int col) const
> {
> 	float x_factor = QFontMetricsF(QApplication::font()).height()/14.0f ;
> 	float y_factor = QFontMetricsF(QApplication::font()).height()/14.0f ;
> 
>     if(e.type == ENTRY_TYPE_NODE)
>         y_factor *= 3.0;
> 
>     if((e.type == ENTRY_TYPE_PROFILE) && !isProfileExpanded(e))
>         y_factor *= 3.0;
> 
>     if(e.type == ENTRY_TYPE_GROUP)
>         y_factor = std::max(y_factor, 24.0f / 14.0f ); // allows to fit the 24 pixels icon for groups in the line
> 
> 	switch(col)
> 	{
> 	default:
> 	case COLUMN_THREAD_NAME:               return QVariant( QSize(x_factor * 170, y_factor*14*1.1f ));
> 	case COLUMN_THREAD_IP:                 return QVariant( QSize(x_factor * 75 , y_factor*14*1.1f ));
> 	case COLUMN_THREAD_ID:                 return QVariant( QSize(x_factor * 75 , y_factor*14*1.1f ));
> 	case COLUMN_THREAD_LAST_CONTACT:       return QVariant( QSize(x_factor * 75 , y_factor*14*1.1f ));
> 	}
> }

csoler:
well whatever you do, please avoid absolute sizes. If you do setMinimumWidth(20); for instance, it will look ok on your screen, but not on everyone's screen.

defnax:
alternative option to fix this issue the GXS ID Treewidget item must always contain a icon else the empty one lets shrink the mail items

PR #1898, opened 2020-05-02
try to isolate pqistreamer from pqissl mutex, step 1: pqistreamer::tick_recv()

jolavillette:
int     pqistreamer::tick_recv(uint32_t timeout)
{
//      Only our thread manipulates mIncoming queue and related counters.
//      The lock of pqistreamer mutex is thus not needed here.
//      For the moment and for safety reasons it is still kept in 3 places:
//      - in pqistreamer::tick_recv before calling free_pend_locked, as I dont know what this method actually does
//      - in pqistreamer::handleincomingitem_locked, as it modifies mIncoming queue
//      - in pqistreamer::inReadBytes_locked, as it modifies related counters
//
//      RsStackMutex stack(mStreamerMtx); /**** LOCKED MUTEX ****/
(..)


csoler:
I fully support your investigations here. Make sure to not keep the locked_* names for methods that should not be called while locked! Otherwise we will miss locking in some places. So change the names when you cleanup your PR.

PR #1906, opened 2020-05-04
Fix unix_fcntl_nonblock() to not overwrite any existing flags.

sehraf:
Currently `unix_fcntl_nonblock` removes all flags and sets `O_NONBLOCK` only. This patch fixes this for Linux.

### DIFF
--- a/libretroshare/src/pqi/pqinetwork.cc
+++ b/libretroshare/src/pqi/pqinetwork.cc
@@ -413,9 +413,10 @@ int unix_fcntl_nonblock(int fd)
 {
         int ret;
 
-/******************* WINDOWS SPECIFIC PART ******************/
+/******************* OS SPECIFIC PART ******************/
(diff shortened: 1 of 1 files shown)


csoler:
what was the side effect of overwriting flags?

sehraf:
> what was the side effect of overwriting flags?

not sure to be honest. 
I've stumbled upon this while debugging my wip i2p sam integration.
I've fixed/change it as i would expected such function to only set/clear the flag in question and do nothing more. If clearing is actually intended i would rename it or split it up into two.

On a side note, there are actually many places where `fcntl(fd, F_SETFL, O_NONBLOCK);` is called directly.

Anyway i'm fine if you want to leave it as it is.

G10h4ck:
I think this fix is important at least for IPv6, so if we clear the flags dual stack flag is lost. Also I agree with sehraf that given the name the function should not clear all flags...

PR #1914, opened 2020-05-06
Pqihandler optim

jolavillette:
This PR includes the changes already in PR/1898 (pqistreamer, p3ServiceServer and p3FastService mutex optimization). 

I am now focusing on pqihandler mutex, which is a mandatory checkpoint for most network activity in RS.

Our problem is that for some processing rsmain needs to lock pqihandler and then, while keeping the mutex locked, it loops on all friends nodes locking/unlocking their individual pqiperson pqithreadstreamer and pqissl mutexes. But we know that pqithreadstreamer and pqissl mutexes are both locked 90 pct of the time when there is SSL activity... 

The result is that sometimes the loop during which pqihandler stays locked can last 100 or 200 ms, which is not acceptable.

pqihandler::tick locks pqihandler, then loops on every friend:
{
pqihandler::tick calls pqiperson::tick
pqiperson::tick locks pqiperson then calls pqithreadstreamer::tick
pqithreadstreamer::tick locks pqithreadstreamer then calls pqistreamer::tick_bio
pqistreamer::tick_bio locks pqistreamer then calls pqissl::tick
pqissl::tick locks pqissl then calls ConnectAttempt
}

To correct that I made 2 small changes in pqithreadstreamer::tick and in pqissl::tick

There is a side effect: it looks that the CPU usage is significantly higher, probably because RS does not spend its time waiting for those mutexes anymore.

Now the remaining long locks I notice are mostly in RsDataService (1 common mutex) and in RsGxsNetService (1 mutex per service I think).


libretroshare/src/pqi/pqiservice.cc
```diff
@@ -204,40 +193,27 @@ bool p3ServiceServer::sendItem(RsRawItem *item)
...
-	pqioutput(PQL_DEBUG_ALL, pqiservicezone, 
-		"p3ServiceServer::tick()");
-#endif
-
-	std::map<uint32_t, pqiService *>::iterator it;
+	// make a copy of the service map
+	std::map<uint32_t,pqiService *> local_map;
+	{	
+		RS_STACK_MUTEX(srvMtx);
+		local_map=services;
```
csoler, line 208: I remember discussing this, but now I have a doubt whether we actually need a mutex here. This map is not supposed to change accross time. I mean services are always the same.

libretroshare/src/pqi/pqistreamer.cc
```diff
@@ -199,18 +201,30 @@ RsItem *pqistreamer::GetItem()
 	pqioutput(PQL_DEBUG_ALL, pqistreamerzone, "pqistreamer::GetItem()");
 #endif
 
-	RsStackMutex stack(mStreamerMtx); /**** LOCKED MUTEX ****/
```
csoler, line 202: Removing that lock seems very risky to me, and also not needed anyway: the time spent in taking one element from the queue is negligible. So I suppose you want to avoid waiting for the lock.

libretroshare/src/pqi/pqistreamer.cc
```diff
@@ -263,10 +279,11 @@ void pqistreamer::updateRates()
 			setRate(false, 0);
 		}
 
+		mAvgLastUpdate = t;
+		mAvgReadCount = 0;
+
 		{
```
csoler, line 266: I really dont get the point in moving the lock here.

libretroshare/src/pqi/pqistreamer.cc
jolavillette: We don't save much by removing this lock, but it is still 100pct useless.
There is no competition at all for the mIncoming queue: there is only 1 thread accessing it (pushing in or poping from).

libretroshare/src/pqi/pqistreamer.cc
jolavillette: mAvgLastUpdate and mAvgReadCount don't need to be mutex protected: they are both read and write accessed only by our thread (pqi).

The case of mAvgSentCount is a little bit particular. It is also read and write accessed only by our thread (pqi) and therefore does not need mutex protection. But in pqistreamer::tick_send you will see that the whole handleoutgoing_locked() process (the part that pops data from the mOutPkts outqueue and pass it to SSL) is unchanged: I left it under global mutex protection. It would have been possible and more effective (but also more complicated) to lock the mutex only when actually poping data from the mOutPkts outqueue, but I decided to leave for later. Thats means that the write access to mAvgSentCount when sending data is done by a "locked" method. For consistency I decided to protect also the reset of mAvgSentCount in updateRates even if it is actually not needed. I should probably have asked you what the good practice would have been here. We can change that of course and remove the lock.

libretroshare/src/pqi/pqiservice.cc
jolavillette: I dont know. I left the copy operation under mutex protection because it is non atomic due to the structure of the service map.

csoler:
There's something inherently wrong in this PR, which you can see by doing the following test:
- launch peer A with max upload of 200kB/s
- launch peer B with max DL of 2000 kB/s and start transferring a file from A to B
With the PR, the DL speed will never excess half the upload speed of the peer, which is 100kB/s
With master, the upload speed goes to its allowed maximum on A, which is 200kB/s

jolavillette:
Initially pr/1914 had nothing to do with bandwidth management. I just pushed in pr/1914 the changes that were in pr/1942 because I wanted that people could test both mutex and bandwidth improvements together.

That being said I have done many tests between my home computer (100 kB/s max UP) and my server (10 MB/s max UP and DN) and I was always able to reach the UP limit. I have also tested between 100 Mb/s servers and I easily reached 10 MB/s with a UP limit of 10 MB/s set in preferences.

Maybe you are confused by how the BW is allocated: it is progressive, pqihandler always allocates only half the BW available. But when the BW usage increases it allocates half of the remaining bandwidth, and so on until the full BW is used. This code is old and unchanged in pr/1942.

Only 1 thing is new actually in pr/1942: we now never send more than the download limit that our friend has set in his preferences.

csoler:
This has been tested *a lot* and successfully passes tests. So I merge it. Thx a lot to all the testers.

PR #1924, opened 2020-05-09
Added New Spam Box

defnax:
* Added New Spam Box


retroshare-gui/src/gui/msgs/MessagesDialog.cpp
```diff
@@ -432,6 +437,16 @@ void MessagesDialog::fillQuickView()
 		itemToSelect = item;
 	}
 
+	item = new QListWidgetItem(tr("Spam"), ui.quickViewWidget);
+	item->setIcon(QIcon(IMAGE_SPAM_ON));
```
csoler, line 441: use FileDefs here!

csoler:
Almost ready. Just missing the FilesDefs I mentioned above. Thx!

defnax:
ok i can do it

retroshare-gui/src/gui/msgs/MessagesDialog.cpp
csoler: I'm realizing that these items will only be used once actually
so it's not that critical
it's only useful if the same icons are used somewhere else too
I originally thought that you where creating items in a loop, but no

PR #1925, opened 2020-05-10
Added more colors for highlighting

defnax:
* Added more colors for highlighting

this needs to be improved only pushed to latest master & renamed the voids .

It would be nice if we can add before a text a space, not worked on my tests.
Maybe after this is merged some one can make it better & can be improved later.
with to much green replies cant read more who wrote what on forums.


PhenomRetroShare:
I rewrote this at #1930

defnax:
> I rewrote this at #1930

thank you

PR #1931, opened 2020-05-12
Fix SharedFilesDialog take long time to update.

PhenomRetroShare:
This is due to Qt bug: https://bugreports.qt.io/browse/QTBUG-11438
Thanks to Jolavillette


felisucoibi:
+1 if it works

PR #1934, opened 2020-05-12
Fix AutoLogin

PhenomRetroShare:
Works at least with Win10 and arch.
Add a message to signal a possible cause why it's failed:
"Do have a key wallet installed?"
"Like gnome-keyring or other using "Secret Service" by DBus."

### DIFF
--- a/libretroshare/src/libretroshare.pro
+++ b/libretroshare/src/libretroshare.pro
@@ -327,6 +327,8 @@ INCLUDEPATH *= $${OPENPGPSDK_DIR}
 PRE_TARGETDEPS *= $${OPENPGPSDK_DIR}/lib/libops.a
 LIBS *= $${OPENPGPSDK_DIR}/lib/libops.a -lbz2
 
+################################### HEADERS & SOURCES #############################
+
 HEADERS +=	ft/ftchunkmap.h \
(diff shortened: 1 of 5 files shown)


libretroshare/src/libretroshare.pro
```diff
@@ -327,6 +327,12 @@ INCLUDEPATH *= $${OPENPGPSDK_DIR}
 PRE_TARGETDEPS *= $${OPENPGPSDK_DIR}/lib/libops.a
 LIBS *= $${OPENPGPSDK_DIR}/lib/libops.a -lbz2
 
+################################### Pkg-Config Stuff #############################
+
+LIBS *= $$system(pkg-config --libs $$PKGCONFIG)
```
G10h4ck: This got duplicated around .pro files which uses libretroshare so you probably need to move it into `use_libretroshare.pri` to avoid duplication

G10h4ck (changes requested):

libretroshare/src/libretroshare.pro
PhenomRetroShare: Done, but this should be called at end of file because if PKGCONFIG is changed after in .pro file.

G10h4ck (approved):

PR #1939, opened 2020-05-13
Do not allow send message if there is more then 10 peers selected.

defnax:


sehraf:
I dislike this artificial limit. While it might stop some random key spammed, you can simply just split the list and send multiple messages.
IIRC there is an option no to allow messages from strangers. Or was this chat only?

csoler:
@sehraf : there is such an option. The idea here is to prevent an "easy" way to select tons of users at once and send mass emails to them *and* display a warning about how to do this properly (e.g. use forums). It should still be possible to send messages to many persons at once of course.

sehraf:
Ah, so the change is only noticeable when you select many ids and then right click to send a mail to all. I thought it was a generic mail limitation.

Sorry for the noise :)

felisucoibi:
this is sueless, spammers on't use compose use the api....

jolavillette:
I am not happy with the idea of ennoying all because of the wrong behaviour of a few.

defnax:
some one needs to code the real anti spam system

ideas:
* Who can send Distant Messages or Invites (check before reputation or its a newbie. check?.. )
* Limit Invites per day or hourly... ( how much invites im allowed ?)
* other ideas?....

on social platforms you cant send everyone a message when you are not friends(Facebook) or following each other(Instagram)...

sehraf:
> * Who can send Distant Messages or Invites (check before reputation or its a newbie. check?.. )

The problem here is that many ids have no reputation, If >0 is the limit than nobody would be able to send you a message. Also you cannot filter based in the content easily, see below.

> * Limit Invites per day or hourly... ( how much invites im allowed ?)

Enforced on which end? 
Sender won't care, simply removes the check and sends as much as he want. 
Receiver will receive the first x messages and then drop all following?

> on social platforms you cant send everyone a message when you are not friends(Facebook) or following each other(Instagram)...

They are centralized and can handle this globally. This doesn't work for RS

-----
Sending mails to a circle would be a nice feature and could actually be useful. But considering the following Scenario:
|WhatsApp group|RS circle|
|-|-|
|"can you add *phone number* please" | "can you add *cert* please".|
|"it's *phone number* birthday, call him!" | "here is my archive node *cert* add it!".|

So we have a similar scenario to the spamming one but actually it is desired this time. 

------

> * other ideas?....

maybe something like this:
flag / block messages when
- more than x receivers **AND**
- contain a cert **AND**
- sender has no reputation **AND**
- sender is not in a circle together with us

The main problem is to detect which messages are actually wanted and which not.

defnax:
> 
> Sending mails to a circle would be a nice feature and could actually be useful. But considering the following Scenario:

Yes that would be nice
 for Circles we need Feature as Admin
* Send Message to Circle Members ( when want announce something importand to the members)
* Send Service Invites to ( predefined like a friend invite method, else nobody find the Forum/Channel/Boards for the Circle)
> 
> WhatsApp group	RS circle
> "can you add _phone number_ please"	"can you add _cert_ please".
> "it's _phone number_ birthday, call him!"	"here is my archive node _cert_ add it!".
> So we have a similar scenario to the spamming one but actually it is desired this time.

but we have now retroshare id :)

> 
> > * other ideas?....
> 
> maybe something like this:
> flag / block messages when
> 
> * more than x receivers **AND**
> * contain a cert **AND**
> * sender has no reputation **AND**
> * sender is not in a circle together with us
> 
> The main problem is to detect which messages are actually wanted and which not.

yes thats nice ideas but who can implement it?

felisucoibi:
I think we designed a spam system like the ones you suggest, was in 2018, was designed with itdea of cyril gohack and me and i think was merged, it filters how many packets are sended from not friends ids etc..... becase we knew who was the sender, not the destination.

PR #1942, opened 2020-05-14
bandwidth control improvement

jolavillette:
RS does not take into account the "download limit" set by our friends in their "preferences / network".
This can be verified in our "statistics / bandwidth" window: the OutMax value is the same for all friends, while it should be limited to the value in the "OutAllowed" row that our friends send to us through the BwCtrl service.

This has serious negative consequences: 
- in pqistreamer the value returned by outAllowedBytes_locked is wrongly calculated, so that during every round we push SSL to its limit (mBio->cansend = FALSE) instead of smoothly sending a quantity of data compatible with the available bandwidth on both side
- our OutQueue fills very fast during file transfer

I made some changes in pqihandler::UpdateRates to limit the "out max rate" to the "download limit" set by our friends and received via BwCtrl service. These limits are available in RsConfigDataRates.

Another reason leading to the clogging of our OutQueues is that during file transfer our peers request more data from us than they are actually ready to accept. They request up to 10 pct more than their maximum download limit. I canceled that in ft/fttransfermodule.cc. 

Modified methods are pqihandler.cc::UpdateRates and ftTransferModule::locked_tickPeerTransfer

According to my tests it seems to work: download limits set by our peers are now respected, SSL activity is more smooth, and there is much less clogging of our OutQueues. 

Note that this should also benefit to RTT.

### DIFF
--- a/libretroshare/src/ft/fttransfermodule.cc
+++ b/libretroshare/src/ft/fttransfermodule.cc
@@ -789,15 +789,15 @@ bool ftTransferModule::locked_tickPeerTransfer(peerInfo &info)
 	std::cerr << std::endl;
 #endif
 
-	// cap next_req to desiredRate in order to respect the bandwidth limit and to avoid clogging our outqueue
-	if (next_req > info.desiredRate)
+	if (next_req > info.desiredRate * 1.1)
 	{
-		next_req = info.desiredRate;
+		next_req = info.desiredRate * 1.1;
 #ifdef FT_DEBUG
 		std::cerr << "locked_tickPeerTransfer() Reached MaxRate: next_req: " << next_req;
 		std::cerr << std::endl;
 #endif
 	}
+
 
 	if (next_req > FT_TM_MAX_PEER_RATE)
 	{
--- a/libretroshare/src/ft/fttransfermodule.cc
+++ b/libretroshare/src/ft/fttransfermodule.cc
@@ -789,15 +789,15 @@ bool ftTransferModule::locked_tickPeerTransfer(peerInfo &info)
 	std::cerr << std::endl;
 #endif
 
-	if (next_req > info.desiredRate * 1.1)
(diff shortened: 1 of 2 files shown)


csoler:
@jolavillette : it's importnt to request 10% more data. It's the only posssibility we have to adapt the request sizes to the bandwidth of tunnels. When you request 10% more, if the tunnel allows it, you'll get a faster speed. If not, you get what you want without the 10%, and this is because of the outgoing max. So setting the outgoing max as yu did should already fix the problem.

jolavillette:
The OutMax is only used in pqistreamer (more specifically in outAllowedBytes_locked) to calculate the maximum quantity of data that we will pick from the outqueue and send to the SSL socket every second. 
It does not control at all how much data is put in the outqueue by the services. For FT the receiver has full control. If the receiver asks for more data we will put more data in the outqueue, and the outqueue will grow because pqistreamer will not accept to deliver more than OutMax.

csoler:
Seems to me that FT adapts to bndwidth automatically: it will only request 10% more than the speed it is measuring (instead of the theoretical maximum). So in the equilibrium, the outqueue of the end of the tunnel (or intermediate peers) will only be 10% larger in bytes (not in # of packets) than what the tunnel accepts.
Also, explicitly making FT account for the outmax of direct friends is inconsistent because the actual tunnel that comes from the friend has a bandwidth that depends on other factors and outmax of the friend is only a conservative maximum. It depends on how many tunnels come from this friend, some of which are not even for us, and it also depends on the bw use of other services.

jolavillette:
I suggest that as a first step you merge the changes in pqihandler.cc (taking into account the download limit set by friends) and leave fttransmodules.cc as it is (with the 1.1 factor). There is no hurry.
We will see how the outqueues behave, specially in case of high speed transfers between directly connected friends (no intermediate node in tunnel).

PR #1944, opened 2020-05-16
[Forums] Added new forum message window [Open in a new window]

defnax:
* Added new forum message window [Open in a new window]


retroshare-gui/src/retroshare-gui.pro
```diff
@@ -1322,6 +1324,7 @@ gxsforums {
 		gui/gxsforums/GxsForumThreadWidget.cpp \
 		gui/gxsforums/GxsForumModel.cpp \
 		gui/gxsforums/GxsForumUserNotify.cpp \
+		gui/gxsforums/ForumMessageView.cpp \
```
PhenomRetroShare: Why didn't you named it GxsForumMessageView or best GxsForumMessageDialog ?

defnax: i can rename it, but i never like add naming with GXS before , normaly on gui we doesnt need everything naming with gxs? only to look same when other classes starts with gxs right..?

PhenomRetroShare: Neither do I, but all other class are named like this so keep homogeneity.

defnax: ok

defnax: > Neither do I, but all other class are named like this so keep homogeneity.

hi i has renamed it, if you see some other things?

PhenomRetroShare: Yes some little, I'll comment them latter. I don't have the time now.

PR #1953, opened 2020-05-20
[RFC] Add API method to request forums sync

G10h4ck:


defnax:
i hope this for the adanced elrepio android app right?  hope the flutter mobile app goes as chat only version (like telegram/whatsapp)  (from b1rd)

G10h4ck:
Well this just goes in the API, every app can do whatever they want with it ;)

csoler:
it's a bit frustrating to see that we need this. Normally GXS should do it automatically. Could we make a list of when sync needs to be triggered? 
Also we need to not sync more often than every X seconds to avoid creating tons of sync each of which causes a cascade of sync events that usually get larger (the first one, TS for the whole DB is super fast, but then exchange of msg ids and grp ids and so on get larger. So it's importnt that a sync can fully complete before we start a new one).

G10h4ck:
So the good reason to expose this is that in mobile networks users often activate the mobile internet connection (very expensive depending on country) and then open the application and expect it to have fresh stuff almost instantly, exposing this via the JSON API when Android notify the apps that internet has been jut turned on and it is working we can trigger a sync just in that moment (we already have an API to trigger friends connection attempts). If we wait for the timer it is likely that the connection has been disabled already.

Other reason this might be **only temporarily** useful in the meanwhile we develop proper behaviour directly into libretroshare are:

- When a group is subscribed it seems a sync should be performed right away without waiting for the timer.
- When a message is published a sync should be automatically instantly triggered without waiting for the timer.
- When a friend connection is established (especially if we haven't been connected lately) it could be a good idea to trigger a sync without waiting for the timer

About multiple sync being too near each other, I suppose last sync timestamp is stored in some variable, I could check that it is not too recent, any suggestion about this @csoler ?

csoler:
About the connection not being available, we should make the sync more clever: remember for each friend last time it was sync-ed (not a global TS for all friends) and sync right away that friend when it comes online. Doing so, right after the friend gets available, the sync would happen automatically.

About last sync, it's not that easy.  What would happen if you e.g. edit 2 groups in less than 1 minute and the minimum time between 2 sync is 1 min ? The first edit would cause an immediate sync but not the second one.

csoler:
Also the way it is implemented in your PR, local modifications will not sync. When you call syncWithPeers() you send to each friend the last TS you know about *their* data. So it's only if their data is modified that the sync will actually happen, not your own data.
In this case what we need is a EITHER syncAskItem that you send to friends, which when received makes them call syncWithPeers with you only, OR directly send to the friend your new list of groups, which means you directly call the method that would be called in the syncWithPeers *from the friends* gives you an old TS.

G10h4ck:
Github is reporting this has been merged but this never got merged, it is still a WIP

G10h4ck:
> Also the way it is implemented in your PR, local modifications will not sync.

AFAIR from what we discussed a bit of time ago, locally created messages/groups are sent immediately to online peers, do I remember wrong?
In case I remember we should still send stuff that is newer locally but the problem would be someway less compelling right now.

> When you call syncWithPeers() you send to each friend the last TS you know about _their_ data. So it's only if their data is modified that the sync will actually happen, not your own data.
> In this case what we need is a EITHER syncAskItem that you send to friends, which when received makes them call syncWithPeers with you only, OR directly send to the friend your new list of groups, which means you directly call the method that would be called in the syncWithPeers _from the friends_ gives you an old TS.

I have dig a bit in the code and those methods should be void `RsGxsNetService::handleRecvSyncGroup(RsNxsSyncGrpReqItem *item)` and `void RsGxsNetService::handleRecvSyncMessage(RsNxsSyncMsgReqItem *item,bool item_was_encrypted)` am I on the right track?

Thanks for putting time into this!
Cheers

PR #1959, opened 2020-05-21
RsBase64 handle correcly 0 lenght buffer encoding and padding

G10h4ck:


G10h4ck:
Once CI finish this is ready to merge

PR #1967, opened 2020-05-22
Fix miss doxygen parameter

PhenomRetroShare:
Parameter: votes of: /rsGxsChannels/getChannelAllContent declared in:
retroshare/rsgxschannels.h miss doxygen parameter direction attribute!


G10h4ck:
too late :p I already fixed it with 77f818f0b37236f0a2994b9f607a3810fa5bf33f

PR #1970, opened 2020-05-24
TheWire rework UI, and update Data Messages.

drbob:
Significant rework of The Wire to support:
 * A interface closer to Twitter.
 * Replies / Republishes / Likes
 * Up to 4 Images


defnax:
i like this design lots better then the other one :)

libretroshare/src/services/p3wire.cc
```diff
@@ -31,6 +31,23 @@
 RsWire *rsWire = NULL;
 
 
+uint32_t RsWirePulse::ImageCount()
+{
+	if (!mImage4.empty()) {
```
csoler: The logic seems risky here. One can craft a msg with mImage3=nullptr and mImage4 not empty. But still the number of images will not be 4.

defnax:
at the moment i doesnt like the Show Group & Show Parent Buttons which not needed there
best we need a Back Button on the same place like on twitter on top left
image

@drbob i have a little pr changes for the group item
https://github.com/drbob/RetroShare/pull/1

image

drbob:
Defnax: I like the idea of a back button + history, I'll try and add this in the next PR.
I think the "Show Parent" and "Show Group" are necessary - just to allow fluid navigation.
Obviously, I'll change the buttons into clickable Labels, in the future.
PS. I've merged your PR.

libretroshare/src/services/p3wire.cc
drbob: made this more robust

PR #1971, opened 2020-05-24
i2p refactoring

sehraf:
This patchset factors out some common i2p code (which can later be reused by SAM)
It also does some maintenance:
 - convert bob and autoproxy to `RsDebug`
 - handle timeouts during `recv()` (instead of endless loop)
 - use `RsThread::async()` to make autoproxy's asynchronous tickets actually async
 - fix a bug when manually setting a private key

### DIFF
--- a/libretroshare/src/services/autoproxy/p3i2pbob.cc
+++ b/libretroshare/src/services/autoproxy/p3i2pbob.cc
@@ -44,8 +44,8 @@ static const std::string kConfigKeyOutLength   = "OUT_LENGTH";
 static const std::string kConfigKeyOutQuantity = "OUT_QUANTITY";
 static const std::string kConfigKeyOutVariance = "OUT_VARIANCE";
 
-/// Sleep duration for receiving loop
-static const useconds_t sleepTimeRecv = 10; // times 1000 = 10ms
+/// Sleep duration for receiving loop in error/no-data case
+static const useconds_t sleepTimeRecv = 250; // times 1000 = 250ms
 /// Sleep duration for everything else
 static const useconds_t sleepTimeWait = 50; // times 1000 = 50ms or 0.05s
 static const int sleepFactorDefault   = 10; // 0.5s
@@ -1069,38 +1069,62 @@ void p3I2pBob::updateSettings_locked()
 
 std::string p3I2pBob::recv()
 {
+	// BOB works line based
+	// -> \n indicates and of the line
+
+	constexpr uint16_t bufferSize = 128;
+	char buffer[bufferSize];
+
 	std::string ans;
-	ssize_t length;
-	const uint16_t bufferSize = 128;
-	std::vector<char> buffer(bufferSize);
+	uint16_t retry = 10;
 
 	do {
-		doSleep(sleepTimeRecv);
+		memset(buffer, 0, bufferSize);
 
-		// there is only one thread that touches mSocket - no need for a lock
-		length = ::recv(mSocket, buffer.data(), buffer.size(), 0);
-		if (length < 0)
+		// peek at data
+		auto length = ::recv(mSocket, buffer, bufferSize, MSG_PEEK);
+		if (length <= 0) {
+			if (length < 0) {
+				// error
+				perror(__PRETTY_FUNCTION__);
+			}
+			retry--;
+			doSleep(sleepTimeRecv);
 			continue;
+		}
 
-		ans.append(buffer.begin(), buffer.end());
+		// at least one byte was read
 
-		// clean received string
-		ans.erase(std::remove(ans.begin(), ans.end(), '\0'), ans.end());
-		ans.erase(std::remove(ans.begin(), ans.end(), '\n'), ans.end());
+		// search for new line
+		auto bufferStr = std::string(buffer);
+		size_t pos = bufferStr.find('\n');
 
-#if 0
-		std::stringstream ss;
-		ss << "recv length: " << length << " (bufferSize: " << bufferSize << ") ans: " << ans.length();
-		Dbg4() << __PRETTY_FUNCTION__ << " " + ss.str() << std::endl;
-#endif
+		if (pos == std::string::npos) {
+			// no new line found -> more to read
 
-		// clear and resize buffer again
-		buffer.clear();
-		buffer.resize(bufferSize);
+			// sanity check
+			if (length != bufferSize) {
+				// expectation: a full buffer was peeked)
+				Dbg1() << __PRETTY_FUNCTION__ << " peeked less than bufferSize but also didn't found a new line character" << std::endl;
+			}
+			// this should never happen
+			assert(length <= bufferSize);
+		} else {
+			// new line found -> end of message
 
-		if (this->shouldStop())
-			break;
-	} while(length == bufferSize || ans.size() < 4);
+			// calculate how much there is to read, read the \n, too!
+			length = pos + 1;
+
+			// end loop
+			retry = 0;
+		}
+
+		// now read for real
+		memset(buffer, 0, bufferSize);
+		length = ::recv(mSocket, buffer, length, 0);
+		bufferStr = std::string(buffer);
+		ans.append(bufferStr);
+	} while(retry > 0);
 
 	return ans;
 }
--- a/libretroshare/src/services/autoproxy/p3i2pbob.cc
+++ b/libretroshare/src/services/autoproxy/p3i2pbob.cc
@@ -26,7 +26,6 @@
 
 #include "pqi/p3peermgr.h"
 #include "rsitems/rsconfigitems.h"
-#include "util/rsdebug.h"
 #include "util/radix32.h"
 #include "util/radix64.h"
 #include "util/rsdebug.h"
@@ -51,8 +50,6 @@ static const useconds_t sleepTimeWait = 50; // times 1000 = 50ms or 0.05s
 static const int sleepFactorDefault   = 10; // 0.5s
 static const int sleepFactorFast      = 1;  // 0.05s
 static const int sleepFactorSlow      = 20; // 1s
-
-RS_SET_CONTEXT_DEBUG_LEVEL(0)
 
 static const rstime_t selfCheckPeroid = 30;
 
@@ -83,7 +80,7 @@ bool p3I2pBob::isEnabled()
 
 bool p3I2pBob::initialSetup(std::string &addr, uint16_t &/*port*/)
 {
-	std::cout << "p3I2pBob::initialSetup" << std::endl;
+	RS_DBG("") << std::endl;
 
 	// update config
 	{
@@ -96,7 +93,7 @@ bool p3I2pBob::initialSetup(std::string &addr, uint16_t &/*port*/)
 		}
 	}
 
-	std::cout << "p3I2pBob::initialSetup config updated" << std::endl;
+	RS_DBG("config updated") << std::endl;
 
 	// request keys
 	// p3I2pBob::stateMachineBOB expects mProcessing to be set therefore
@@ -106,12 +103,12 @@ bool p3I2pBob::initialSetup(std::string &addr, uint16_t &/*port*/)
 	fakeTicket->task = autoProxyTask::receiveKey;
 	processTaskAsync(fakeTicket);
 
-	std::cout << "p3I2pBob::initialSetup fakeTicket requested" << std::endl;
+	RS_DBG("fakeTicket requested") << std::endl;
 
 	// now start thread
 	start("I2P-BOB gen key");
 
-	std::cout << "p3I2pBob::initialSetup thread started" << std::endl;
+	RS_DBG("thread started") << std::endl;
 
 	int counter = 0;
 	// wait for keys
@@ -125,24 +122,24 @@ bool p3I2pBob::initialSetup(std::string &addr, uint16_t &/*port*/)
 			break;
 
 		if (++counter > 30) {
-			std::cout << "p3I2pBob::initialSetup timeout!" << std::endl;
+			RS_DBG4("timeout!") << std::endl;
 			return false;
 		}
 	}
 
-	std::cout << "p3I2pBob::initialSetup got keys" << std::endl;
+	RS_DBG("got keys") << std::endl;
 
 	// stop thread
 	fullstop();
 
-	std::cout << "p3I2pBob::initialSetup thread stopped" << std::endl;
+	RS_DBG("thread stopped") << std::endl;
 
 	{
 		RS_STACK_MUTEX(mLock);
 		addr = mSetting.address.base32;
 	}
 
-	std::cout << "p3I2pBob::initialSetup addr '" << addr << "'" << std::endl;
+	RS_DBG4("addr '" + addr + "'") << std::endl;
 
 	return true;
 }
@@ -160,7 +157,7 @@ void p3I2pBob::processTaskAsync(taskTicket *ticket)
 	}
 		break;
 	default:
-		RsDbg() << __PRETTY_FUNCTION__ << " unknown task" << std::endl;
+		RS_DBG("unknown task") << std::endl;
 		rsAutoProxyMonitor::taskError(ticket);
 		break;
 	}
@@ -175,7 +172,7 @@ void p3I2pBob::processTaskSync(taskTicket *ticket)
 	case autoProxyTask::status:
 		// check if everything needed is set
 		if (!data) {
-			RsDbg() << __PRETTY_FUNCTION__ << " autoProxyTask::status data is missing" << std::endl;
+			RS_DBG("autoProxyTask::status data is missing") << std::endl;
 			rsAutoProxyMonitor::taskError(ticket);
 			break;
 		}
@@ -189,7 +186,7 @@ void p3I2pBob::processTaskSync(taskTicket *ticket)
 	case autoProxyTask::getSettings:
 		// check if everything needed is set
 		if (!data) {
-			RsDbg() << __PRETTY_FUNCTION__ << " autoProxyTask::getSettings data is missing" << std::endl;
+			RS_DBG("autoProxyTask::getSettings data is missing") << std::endl;
 			rsAutoProxyMonitor::taskError(ticket);
 			break;
 		}
@@ -203,7 +200,7 @@ void p3I2pBob::processTaskSync(taskTicket *ticket)
 	case autoProxyTask::setSettings:
 		// check if everything needed is set
 		if (!data) {
-			RsDbg() << __PRETTY_FUNCTION__ << " autoProxyTask::setSettings data is missing" << std::endl;
+			RS_DBG("autoProxyTask::setSettings data is missing") << std::endl;
 			rsAutoProxyMonitor::taskError(ticket);
 			break;
 		}
@@ -223,7 +220,7 @@ void p3I2pBob::processTaskSync(taskTicket *ticket)
 		break;
 	case autoProxyTask::getErrorInfo:
 		if (!data) {
-			RsDbg() << __PRETTY_FUNCTION__ << " autoProxyTask::getErrorInfo data is missing" << std::endl;
+			RS_DBG("autoProxyTask::getErrorInfo data is missing") << std::endl;
 			rsAutoProxyMonitor::taskError(ticket);
 		} else {
 			RS_STACK_MUTEX(mLock);
@@ -232,7 +229,7 @@ void p3I2pBob::processTaskSync(taskTicket *ticket)
 		}
 		break;
 	default:
-		RsDbg() << __PRETTY_FUNCTION__ << " unknown task" << std::endl;
+		RS_DBG("unknown task") << std::endl;
 		rsAutoProxyMonitor::taskError(ticket);
 		break;
 	}
@@ -253,7 +250,7 @@ void p3I2pBob::threadTick()
 		RS_STACK_MUTEX(mLock);
 		std::stringstream ss;
 		ss << "data_tick mState: " << mState << " mTask: " << mTask << " mBOBState: " << mBOBState << " mPending: " << mPending.size();
-		Dbg4() << __PRETTY_FUNCTION__ << " " + ss.str() << std::endl;
+		RS_DBG4("" + ss.str()) << std::endl;
 	}
 
 	sleepTime += stateMachineController();
@@ -294,13 +291,13 @@ int p3I2pBob::stateMachineBOB()
 		while (answer.find("OK Listing done") == std::string::npos) {
 			std::stringstream ss;
 			ss << "stateMachineBOB status check: read loop, counter: " << counter;
-			Dbg3() << __PRETTY_FUNCTION__ << " " + ss.str() << std::endl;
+			RS_DBG3("" + ss.str()) << std::endl;
 			answer += recv();
 			counter++;
 		}
 
 		if (answer.find(mTunnelName) == std::string::npos) {
-			RsDbg() << __PRETTY_FUNCTION__ << " status check: tunnel down!" << std::endl;
+			RS_DBG("status check: tunnel down!") << std::endl;
 			// signal error
 			*((bool *)mProcessing->data) = true;
 		}
@@ -340,8 +337,8 @@ int p3I2pBob::stateMachineBOB_locked_failure(const std::string &answer, const bo
 		return sleepFactorDefault;
 	}
 
-	RsDbg() << __PRETTY_FUNCTION__ << " FAILED to run command '" + currentState.command + "'" << std::endl;
-	RsDbg() << __PRETTY_FUNCTION__ << " '" + answer + "'" << std::endl;
+	RS_DBG("FAILED to run command '" + currentState.command + "'") << std::endl;
+	RS_DBG("'" + answer + "'") << std::endl;
 
 	mErrorMsg.append("FAILED to run command '" + currentState.command + "'" + '\n');
 	mErrorMsg.append("reason '" + answer + "'" + '\n');
@@ -388,14 +385,14 @@ int p3I2pBob::stateMachineController()
 		return stateMachineController_locked_idle();
 	case csDoConnect:
 		if (!connectI2P()) {
-			RsDbg() << __PRETTY_FUNCTION__ << " doConnect: unable to connect" << std::endl;
+			RS_DBG("doConnect: unable to connect") << std::endl;
 			mStateOld = mState;
 			mState = csError;
 			mErrorMsg = "unable to connect to BOB port";
 			return sleepFactorSlow;
 		}
 
-		Dbg4() << __PRETTY_FUNCTION__ << " doConnect: connected" << std::endl;
+		RS_DBG4("doConnect: connected") << std::endl;
 		mState = csConnected;
 		break;
 	case csConnected:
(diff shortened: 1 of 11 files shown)


chelovechishko:
Could you also add a flag to not compiling in an i2pbob support (inlcuding a thread for i2pbob serving)?
Something around: 
```c++
diff --git a/libretroshare/src/pqi/pqisslpersongrp.cc b/libretroshare/src/pqi/pqisslpersongrp.cc
index 7016a2fa6..84fcdafae 100644
--- a/libretroshare/src/pqi/pqisslpersongrp.cc
+++ b/libretroshare/src/pqi/pqisslpersongrp.cc
@@ -75,13 +75,17 @@ pqiperson * pqisslpersongrp::locked_createPerson(const RsPeerId& id, pqilistener
 #endif
 
 		// Use pqicI2PBOB for I2P
-		pqiconnect *pqicSOCKSProxy, *pqicI2PBOB;
+		pqiconnect *pqicSOCKSProxy;
+#ifdef I2PBOB
+        pqiconnect *pqicI2PBOB;
+#endif
 		{
 			pqisslproxy  *pqis = new pqisslproxy((pqissllistener *) listener, pqip, mLinkMgr);
 			RsSerialiser *rss  = new RsSerialiser();
 			rss->addSerialType(new RsRawSerialiser());
 			pqicSOCKSProxy = new pqiconnect(pqip, rss, pqis);
 		}
+#ifdef I2PBOB
 		if (rsAutoProxyMonitor::instance()->isEnabled(autoProxyType::I2PBOB))
 		{
 			pqissli2pbob *pqis = new pqissli2pbob((pqissllistener *) listener, pqip, mLinkMgr);
@@ -92,6 +96,7 @@ pqiperson * pqisslpersongrp::locked_createPerson(const RsPeerId& id, pqilistener
 		} else {
 			pqicI2PBOB = pqicSOCKSProxy;
 		}
+#endif
 
 
 		/* first select type based on peer */
@@ -100,17 +105,21 @@ pqiperson * pqisslpersongrp::locked_createPerson(const RsPeerId& id, pqilistener
 		case RS_HIDDEN_TYPE_TOR:
 			pqip -> addChildInterface(PQI_CONNECT_HIDDEN_TOR_TCP, pqicSOCKSProxy);
 			break;
+#ifdef I2PBOB
 		case RS_HIDDEN_TYPE_I2P:
 			pqip -> addChildInterface(PQI_CONNECT_HIDDEN_I2P_TCP, pqicI2PBOB);
 			break;
+#endif
 		default:
 			/* peer is not a hidden one but we are */
 			/* select type based on ourselves */
 			uint32_t typeOwn = mPeerMgr->getHiddenType(AuthSSL::getAuthSSL()->OwnId());
 			switch (typeOwn) {
+#ifdef I2PBOB
 			case RS_HIDDEN_TYPE_I2P:
 				pqip -> addChildInterface(PQI_CONNECT_HIDDEN_I2P_TCP, pqicI2PBOB);
 				break;
+#endif
 			default:
 				/* this case shouldn't happen! */
 				std::cerr << "pqisslpersongrp::locked_createPerson WARNING INVALID HIDDEN TYPES - THIS SHOULD NOT HAPPEN!" << std::endl;
diff --git a/libretroshare/src/rsserver/p3face.h b/libretroshare/src/rsserver/p3face.h
index 01a51caeb..7f7263a57 100644
--- a/libretroshare/src/rsserver/p3face.h
+++ b/libretroshare/src/rsserver/p3face.h
@@ -43,7 +43,9 @@
 
 class p3heartbeat;
 class p3discovery2;
+#ifdef I2PBOB
 class p3I2pBob;
+#endif
 
 /* GXS Classes - just declare the classes.
    so we don't have to totally recompile to switch */
@@ -165,7 +167,9 @@ public:
 		p3ChatService *chatSrv;
 		p3StatusService *mStatusSrv;
 		p3GxsTunnelService *mGxsTunnels;
+#ifdef I2PBOB
 		p3I2pBob *mI2pBob;
+#endif
 
         // This list contains all threaded services. It will be used to shut them down properly.
 
diff --git a/libretroshare/src/rsserver/rsinit.cc b/libretroshare/src/rsserver/rsinit.cc
index e3f1f141e..29cd748bf 100644
--- a/libretroshare/src/rsserver/rsinit.cc
+++ b/libretroshare/src/rsserver/rsinit.cc
@@ -142,7 +142,10 @@ struct RsInitConfig
 		std::string hiddenNodeAddress;
 		uint16_t    hiddenNodePort;
 
+#undef I2PBOB
+#ifdef I2PBOB
 		bool        hiddenNodeI2PBOB;
+#endif
 
 		/* Logging */
 		bool haveLogFile;
@@ -687,7 +690,9 @@ void RsInit::SetHiddenLocation(const std::string& hiddenaddress, uint16_t port,
 	rsInitConfig->hiddenNodeSet = true;
 	rsInitConfig->hiddenNodeAddress = hiddenaddress;
 	rsInitConfig->hiddenNodePort = port;
+#ifdef I2PBOB
 	rsInitConfig->hiddenNodeI2PBOB = useBob;
+#endif
 }
 
 
@@ -946,9 +951,12 @@ int RsServer::StartupRetroShare()
 	mPeerMgr->setManagers(mLinkMgr, mNetMgr);
 	mNetMgr->setManagers(mPeerMgr, mLinkMgr);
 
+// 
+#ifdef I2PBOB
 	rsAutoProxyMonitor *autoProxy = rsAutoProxyMonitor::instance();
 	mI2pBob = new p3I2pBob(mPeerMgr);
 	autoProxy->addProxy(autoProxyType::I2PBOB, mI2pBob);
+#endif
 
 	//load all the SSL certs as friends
 	//        std::list<std::string> sslIds;
@@ -1669,7 +1677,9 @@ int RsServer::StartupRetroShare()
 	//mConfigMgr->addConfiguration("photo.cfg", photo_ns);
 	//mConfigMgr->addConfiguration("wire.cfg", wire_ns);
 #endif //RS_ENABLE_GXS
+#ifdef I2PBOB
 	mConfigMgr->addConfiguration("I2PBOB.cfg", mI2pBob);
+#endif
 
 	mPluginsManager->addConfigurations(mConfigMgr) ;
 
@@ -1721,6 +1731,7 @@ int RsServer::StartupRetroShare()
 	{
 		std::cout << "RsServer::StartupRetroShare setting up hidden locations" << std::endl;
 
+#ifdef I2PBOB
 		if (rsInitConfig->hiddenNodeI2PBOB) {
 			std::cout << "RsServer::StartupRetroShare setting up BOB" << std::endl;
 
@@ -1751,8 +1762,11 @@ int RsServer::StartupRetroShare()
 				/// TODO add notify for failed bob setup
 			}
 		} else {
+#endif
 			mPeerMgr->setupHiddenNode(rsInitConfig->hiddenNodeAddress, rsInitConfig->hiddenNodePort);
+#ifdef I2PBOB
 		}
+#endif
 
 		std::cout << "RsServer::StartupRetroShare hidden location set up" << std::endl;
 	}
@@ -1773,17 +1787,23 @@ int RsServer::StartupRetroShare()
 		// trigger updates for auto proxy services
 		std::vector<autoProxyType::autoProxyType_enum> types;
 
+#ifdef I2PBOB
 		// i2p bob need to rebuild its command map
 		types.push_back(autoProxyType::I2PBOB);
 
+// 
 		rsAutoProxyMonitor::taskSync(types, autoProxyTask::reloadConfig);
+#endif
 	}
 
 	/**************************************************************************/
 	/* startup (stuff dependent on Ids/peers is after this point) */
 	/**************************************************************************/
 
+// 
+#ifdef I2PBOB
 	autoProxy->startAll();
+#endif
 
 	pqih->init_listener();
 	mNetMgr->addNetListener(pqih); /* add listener so we can reset all sockets later */
@@ -1814,8 +1834,10 @@ int RsServer::StartupRetroShare()
 	/* Start up Threads */
 	/**************************************************************************/
 
+#ifdef I2PBOB
 	// auto proxy threads
 	startServiceThread(mI2pBob, "I2P-BOB");
+#endif
 
 #ifdef RS_ENABLE_GXS
 	// Must Set the GXS pointers before starting threads.
diff --git a/libretroshare/src/services/autoproxy/rsautoproxymonitor.cc b/libretroshare/src/services/autoproxy/rsautoproxymonitor.cc
index aef160590..656fdd0d5 100644
--- a/libretroshare/src/services/autoproxy/rsautoproxymonitor.cc
+++ b/libretroshare/src/services/autoproxy/rsautoproxymonitor.cc
@@ -306,7 +306,7 @@ autoProxyService *rsAutoProxyMonitor::lookUpService(autoProxyType::autoProxyType
 	if ((itService = mProxies.find(t)) != mProxies.end()) {
 		return itService->second;
 	}
-	std::cerr << "sAutoProxyMonitor::lookUpService no service for type " << t << " found!" << std::endl;
+	//std::cerr << "sAutoProxyMonitor::lookUpService no service for type " << t << " found!" << std::endl;
 	return NULL;
 }

```

sehraf:
> Could you also add a flag to not compiling in an i2pbob support (inlcuding a thread for i2pbob serving)?

will have a look

sehraf:
@chelovechishko this should do the trick! Might give it a try?
Default is still BOB enabled but you can now use `CONFIG+=no_rs_bob` to remove it.

G10h4ck:
@sehraf what do you mean by SAM ?

chelovechishko:
Yes, for sure, many thanks, but it will take a time to start
compiling, sorry for inconveniences.

sehraf:
> @sehraf what do you mean by SAM ?

[SAM3](https://geti2p.net/en/docs/api/samv3) is another api (similar to BOB) to configure i2p. 
To quote the [BOB](https://geti2p.net/en/docs/api/bob) page 
> At this point, most of the good ideas from BOB have been incorporated into SAMv3, which has more features and more real-world use. BOB still works, but it is not gaining the advanced features available to SAMv3 and is essentially unsupported at this time.

So in the long run i'd like to switch to SAM3 instead of BOB (while BOB still works fine right now). This will add support for other (i2p) encryption/signature schemes as well as fix a bug related to RS crashes (the bug does not *cause* the crash it is the result).

I've a working branch (locally) but i'm unhappy with some design elements right now. The ~plan~ idea is to merge it once RS has a new release.

> Yes, for sure, many thanks, but it will take a time to start
compiling, sorry for inconveniences.

absolutely no problem :)

chelovechishko:
Had to `sed -i "/hiddenport_/,$ {s/\(.\)$/\1\r/M;s/^$/ \r/}" pr_i2p_refactoring.patch` to apply patch, but otherwise is working.

G10h4ck:
@sehraf so should this patchset be already considered for merging or should we wait after 0.6.6 ? If we should wait could you please put [WIP] indication on the title of the PR ?

sehraf:
This can be merged. No new functionality is introduced (expect the new qmake option to build without BOB)

libretroshare/src/services/autoproxy/p3i2pbob.cc
```diff
@@ -26,6 +26,7 @@
 
 #include "pqi/p3peermgr.h"
 #include "rsitems/rsconfigitems.h"
+#include "util/rsdebug.h"
```
G10h4ck: this got included twice, look at line 32

libretroshare/src/services/autoproxy/p3i2pbob.cc
```diff
@@ -43,20 +44,15 @@ static const std::string kConfigKeyOutLength   = "OUT_LENGTH";
...
 
-static const bool   kDefaultBOBEnable = false;
-static const int8_t kDefaultLength    = 3;
-static const int8_t kDefaultQuantity  = 4;
-static const int8_t kDefaultVariance  = 0;
-
-/// Sleep duration for receiving loop
-static const useconds_t sleepTimeRecv = 10; // times 1000 = 10ms
+/// Sleep duration for receiving loop in error/no-data case
+static const useconds_t sleepTimeRecv = 250; // times 1000 = 250ms
```
G10h4ck, line 47: from where that `useconds_t` came from? The name of the type suggests it to be microseconds but then the comment seems to suggest they are milliseconds

libretroshare/src/services/autoproxy/p3i2pbob.cc
```diff
@@ -287,7 +253,7 @@ void p3I2pBob::threadTick()
 		RS_STACK_MUTEX(mLock);
 		std::stringstream ss;
 		ss << "data_tick mState: " << mState << " mTask: " << mTask << " mBOBState: " << mBOBState << " mPending: " << mPending.size();
-		rslog(RsLog::Debug_All, &i2pBobLogInfo, ss.str());
+		Dbg4() << __PRETTY_FUNCTION__ << " " + ss.str() << std::endl;
```
G10h4ck: This is just a suggestion not a request for change. For conditional debug is better to use the newly introduced macros `RS_DBG4(...)`, so when compiling with a lower debug level the code literally disappear at pre-processing time, the level is then set by including `util/rsdebuglevel4.h` or lower.

libretroshare/src/services/autoproxy/p3i2pbob.cc
```diff
@@ -374,8 +340,8 @@ int p3I2pBob::stateMachineBOB_locked_failure(const std::string &answer, const bo
 		return sleepFactorDefault;
 	}
 
-	rslog(RsLog::Warning, &i2pBobLogInfo, "stateMachineBOB FAILED to run command '" + currentState.command + "'");
-	rslog(RsLog::Warning, &i2pBobLogInfo, "stateMachineBOB '" + answer + "'");
+	RsDbg() << __PRETTY_FUNCTION__ << " FAILED to run command '" + currentState.command + "'" << std::endl;
+	RsDbg() << __PRETTY_FUNCTION__ << " '" + answer + "'" << std::endl;
```
G10h4ck: To guarantee log atomicity, use the same instance of `RsDbg()` aka remove the `;` in the first line and `RsDbg()` in the second

libretroshare/src/services/autoproxy/rsautoproxymonitor.cc
```diff
@@ -168,7 +174,11 @@ void rsAutoProxyMonitor::task(taskTicket *ticket)
 			*tt = *ticket;
 			tt->types.clear();
 			tt->types.push_back(*it);
-			s->processTaskAsync(tt);
+
+			// it's async!
+			RsThread::async([s, tt] {
+				s->processTaskAsync(tt);
```
G10h4ck, line 180: Maybe it is better to put `RsThread::async` directly inside `processTaskAsync` so the name of the function is not misleading anymore.

libretroshare/src/util/i2pcommon.cpp
```diff
@@ -0,0 +1,172 @@
...
+	 * https://geti2p.net/spec/common-structures#destination
+	 * https://geti2p.net/spec/common-structures#keysandcert
+	 * https://geti2p.net/spec/common-structures#certificate
+	 */
+	if (priv.length() < 884) // base64 ( = 663 bytes = KeyCert + priv Keys)
+		return std::string();
+
+	// creat a copy to work on, need to convert it to standard base64
+	auto priv_copy(priv);
+	std::replace(priv_copy.begin(), priv_copy.end(), '~', '/');
```
G10h4ck, line 58: Damn they could have used base64url instead of their custom table...

libretroshare/src/util/i2pcommon.cpp
```diff
@@ -0,0 +1,172 @@
...
+	std::replace(priv_copy.begin(), priv_copy.end(), '-', '+');
```
G10h4ck: Remove this line, as RsBase64::decode supports both base64 and base64url

libretroshare/src/util/i2pcommon.h
```diff
@@ -0,0 +1,222 @@
...
+*/
+enum class CryptoKeyType : uint16_t {
+	ElGamal = 0,
+	P256    = 1,
+	P384    = 2,
+	P521    = 3,
+	X25519  = 4
+};
+
+static const std::map<CryptoKeyType, std::pair<uint16_t, uint16_t>> cryptoKeyLengths {
```
G10h4ck: Isn't an std::array better for this?

libretroshare/src/util/i2pcommon.h
```diff
@@ -0,0 +1,222 @@
...
+	{CryptoKeyType::ElGamal, {256, 256}},
+	{CryptoKeyType::P256,    { 64,  32}},
+	{CryptoKeyType::P384,    { 96,  48}},
+	{CryptoKeyType::P521,    {132,  66}},
+	{CryptoKeyType::X25519,  { 32,  32}},
+};
+
+static const std::map<SigningKeyType, std::pair<uint16_t, uint16_t>> signingKeyLengths {
```
G10h4ck: Isn't an std::array better for this?

G10h4ck (changes requested):

libretroshare/src/services/autoproxy/p3i2pbob.cc
sehraf: I just kept the type that fitted `rstime::rs_usleep()`

libretroshare/src/util/i2pcommon.cpp
sehraf: Currently i only need decoding, so i could expand `rDict`
This should also handle the ~ as 63, right?
```diff
        static constexpr int8_t rDict[256] = {
        /* index  +0  +1  +2  +3  +4  +5  +6  +7  +8  +9 +10 +11 +12 +13 +14 +15 */
        /*   0 */ -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
@@ -117,7 +117,7 @@ private:
        /*  64 */ -1,  0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14,
        /*  80 */ 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, -1, -1, -1, -1, 63,
        /*  96 */ -1, 26, 27, 28, 29, 30, 31, 32, 33, 34, 35, 36, 37, 38, 39, 40,
-       /* 112 */ 41, 42, 43, 44, 45, 46, 47, 48, 49, 50, 51, -1, -1, -1, -1, -1,
+       /* 112 */ 41, 42, 43, 44, 45, 46, 47, 48, 49, 50, 51, -1, -1, -1, 63, -1,
        /* 128 */ -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
        /* 144 */ -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
        /* 160 */ -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
```

libretroshare/src/services/autoproxy/p3i2pbob.cc
sehraf: Do i still need the `std::endl`?

libretroshare/src/services/autoproxy/rsautoproxymonitor.cc
sehraf: Would work, too. But this moves the responsibility to do it async to the implementation of `processTaskAsync`. 
The idea is that, who ever implements a auto proxy service and thus Implements `processTaskAsync`, can't do too much wrong, as the function is already executed in its own thread.

sehraf:
I've rebased the pr and implemented most changes

libretroshare/src/services/autoproxy/p3i2pbob.cc
G10h4ck: nope

libretroshare/src/services/autoproxy/rsautoproxymonitor.cc
G10h4ck: ok

libretroshare/src/util/i2pcommon.cpp
G10h4ck: Probably it would work, but if it is not standardized somewhere so it is easy to just add to the documentation something like `decoding dictionary for base64 + base64url +base64XXXXX`, I would prefer to not touch the dictionary and keep the first std::replace

libretroshare/src/services/autoproxy/p3i2pbob.cc
```diff
@@ -287,7 +250,7 @@ void p3I2pBob::threadTick()
 		RS_STACK_MUTEX(mLock);
 		std::stringstream ss;
 		ss << "data_tick mState: " << mState << " mTask: " << mTask << " mBOBState: " << mBOBState << " mPending: " << mPending.size();
-		rslog(RsLog::Debug_All, &i2pBobLogInfo, ss.str());
+		RS_DBG4("" + ss.str()) << std::endl;
```
G10h4ck: No need of extra `std::endl` also if you use `<<` the param evaluation will not disappear from compiled code even if debug level is less then 4, while everything inside the parentheses will disappear in pre-processing phase, so what I would do here if `ss` is not used for other things more then debug, I would put all those stuff between the parenteses like this `RS_DBG4("data_tick mState: ", mState, " mTask: ", mTask, " mBOBState: ", mBOBState, " mPending: ", mPending.size());` This way when the debugging level is lower all of hat stuff disappear and there is not even param evaluation.

libretroshare/src/services/autoproxy/p3i2pbob.cc
```diff
@@ -328,13 +291,13 @@ int p3I2pBob::stateMachineBOB()
 		while (answer.find("OK Listing done") == std::string::npos) {
 			std::stringstream ss;
 			ss << "stateMachineBOB status check: read loop, counter: " << counter;
-			rslog(RsLog::Debug_Basic, &i2pBobLogInfo, ss.str());
+			RS_DBG3("" + ss.str()) << std::endl;
```
G10h4ck: same here

libretroshare/src/services/autoproxy/p3i2pbob.cc
```diff
@@ -328,13 +291,13 @@ int p3I2pBob::stateMachineBOB()
...
 			answer += recv();
 			counter++;
 		}
 
 		if (answer.find(mTunnelName) == std::string::npos) {
-			rslog(RsLog::Warning, &i2pBobLogInfo, "stateMachineBOB status check: tunnel down!");
+			RS_DBG("status check: tunnel down!") << std::endl;
```
G10h4ck: no need of `std::endl`

libretroshare/src/services/autoproxy/p3i2pbob.cc
```diff
@@ -374,8 +337,8 @@ int p3I2pBob::stateMachineBOB_locked_failure(const std::string &answer, const bo
 		return sleepFactorDefault;
 	}
 
-	rslog(RsLog::Warning, &i2pBobLogInfo, "stateMachineBOB FAILED to run command '" + currentState.command + "'");
-	rslog(RsLog::Warning, &i2pBobLogInfo, "stateMachineBOB '" + answer + "'");
+	RS_DBG("FAILED to run command '" + currentState.command + "'") << std::endl;
```
G10h4ck: no need of `std::endl` and also no need of `+` just put `,` and the the macro will do the rest

G10h4ck (changes requested):
Just remove the extras `std::stringstream` and `std::endl` and it is fine for me ;)

libretroshare/src/services/autoproxy/p3i2pbob.cc
sehraf: Ah, I wasn't aware that I can replace the << with commas. Will do tomorrow!

sehraf:
So, next round :grinning: I hope i've done everything right this time. Also :+1: for improving RS technologically.

libretroshare/src/services/autoproxy/p3i2pbob.cc
```diff
@@ -979,8 +936,8 @@ void p3I2pBob::finalizeSettings_locked()
 	sockaddr_storage_setipv4(mI2PProxyAddr, (sockaddr_in*)&proxy);
 	sockaddr_storage_setport(mI2PProxyAddr, 2827);
 
-	rslog(RsLog::Debug_Basic, &i2pBobLogInfo, "finalizeSettings_locked using " + sockaddr_storage_tostring(mI2PProxyAddr));
-	rslog(RsLog::Debug_Basic, &i2pBobLogInfo, "finalizeSettings_locked using " + mSetting.addr);
+	RS_DBG4("using ", sockaddr_storage_tostring(mI2PProxyAddr));
```
G10h4ck: AFAIR it would probably accept also `mI2PProxyAddr` and print it nicely without need of calling explicitly `sockaddr_storage_tostring`

libretroshare/src/util/i2pcommon.h
```diff
@@ -0,0 +1,224 @@
...
+	/*SigningKeyType::RedDSA_SHA512_Ed25519  */ std::make_pair<uint16_t, uint16_t>( 32,  32),
+};
+
+
+/**
+ * @brief generateNameSuffix Generates a base32 name suffix for tunnel identification
+ * @param len lenght of random bytes, will be expanded by base32 encoding
+ * @return base32 string
+ *
+ * RSRandom::random_alphaNumericString can return very weird looking strings like: ,,@z+M
```
G10h4ck: That seems a bug or at least a misleading function name in  RsRandom::random_alphaNumericString

G10h4ck (approved):

libretroshare/src/util/i2pcommon.h
sehraf: Indeed, thats the line that generates the characters:
`s += (char)( (random_u32()%94) + 33) ;`

libretroshare/src/services/autoproxy/p3i2pbob.cc
sehraf: looks good:
> D 1591975806.104 void p3I2pBob::finalizeSettings_locked() using ipv4://127.0.0.1:2827

libretroshare/src/services/autoproxy/rsautoproxymonitor.cc
```diff
@@ -244,20 +256,22 @@ void rsAutoProxyMonitor::taskDone(taskTicket *t, autoProxyStatus::autoProxyStatu
...
+			print_stacktrace();
 			cleanUp = true;
 		}
 	} else if (t->async){
 		// async and no callback
 		// we must take care of deleting
 		cleanUp = true;
 		if(t->data)
-			std::cerr << "(WW) rsAutoProxyMonitor::taskFinish async call with data attached but no callback set!" << std::endl;
+			RS_ERR("sync call with data attached but no callback set!");
```
G10h4ck: you probably forgot an "a" at the begin of the message

libretroshare/src/util/i2pcommon.cpp
```diff
@@ -0,0 +1,174 @@
...
+	std::transform(out.begin(), out.end(), out.begin(), ::tolower);
+	out.append(".b32.i2p");
+
+	return out;
+}
+
+const std::string makeOption(const std::string &lhs, const int8_t &rhs) {
+	// convert number to int
+	std::ostringstream oss;
+	oss << (int)rhs;
```
G10h4ck: You can also use `std::to_string` instead so it becomes just one line ;)

G10h4ck (changes requested):

G10h4ck:
With this PR https://github.com/RetroShare/RetroShare/pull/2014 RsRandom should return a proper random alpha numeric string

libretroshare/src/services/autoproxy/rsautoproxymonitor.cc
sehraf: nice catch!

libretroshare/src/util/i2pcommon.cpp
sehraf: done :+1:

sehraf:
The last commit relies on #2014 and is thus untested. 
If you want i can rebase the pr once #2014  is merged.

G10h4ck (approved):

PR #1979, opened 2020-05-27
Implementing a cache for MsgMeta

csoler:


felisucoibi:
wiii cache

csoler:
Has been tested a lot, without problems.

PR #1982, opened 2020-05-28
rsdebug supports variadic style for atomic log

G10h4ck:
This PR add a new syntax for `RsDbg` and all related classes.
now it is possible to write something like

`RsInfo("libretroshare version: " , RS_HUMAN_READABLE_VERSION);`

which behave similarly to 

`RsInfo() << "libretroshare version: " << RS_HUMAN_READABLE_VERSION
	         << std::endl;`

But being one single function call it guarantee atomic printing since C++11 thus avoiding lines getting overlapped in the logs. Manipulators and stuff like `std::endl` (which is easy to substitute with `\n` if necessary) are not supported in the new syntax, you can stick to the `<<` one as it is still supported if you need manipulators.

### DIFF
--- a/libretroshare/src/util/rsdebug.h
+++ b/libretroshare/src/util/rsdebug.h
@@ -46,6 +46,18 @@ struct t_RsLogger
 {
 	inline t_RsLogger() = default;
 
+	/** Offer variadic style too, as a benefit this has better atomicity then
+	 * << style, but doesn't supports manipulators and things like std::endl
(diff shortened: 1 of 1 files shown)


G10h4ck:
@jolavillette this should solve the issue with overlapping logs you where experiencing once and for all, please test it.

felisucoibi:
+1

jolavillette:
Thank you Gio, will test it.

PR #1987, opened 2020-05-30
Fix Message Sent keeps Tags

PhenomRetroShare:


libretroshare/src/services/p3msgservice.cc
```diff
@@ -1187,11 +1187,15 @@ bool 	p3MsgService::MessageSend(MessageInfo &info)
...
 		msg->msgId = getNewUniqueMsgId();
-                
+
 		msg->msgFlags |= RS_MSG_OUTGOING;
 
 		imsg[msg->msgId] = msg;
 
+		// Update info for caller
+		info.msgId = std::to_string(msg->msgId);
+		info .msgflags = msg->msgFlags;
```
chelovechishko, line 1197: a space

PhenomRetroShare: Oups, too late.
But seems to work...

PR #1990, opened 2020-05-31
function to set sync/store parameters for all subscriptions

RetroPooh:


felisucoibi:
+1 something that should be there since along long time ago

libretroshare/src/gxs/rsgenexchange.cc
```diff
@@ -1891,6 +1891,13 @@ void     RsGenExchange::setSyncPeriod(const RsGxsGroupId& grpId,uint32_t age_in_
     else
         std::cerr << "(EE) No network service available. Cannot set storage period. " << std::endl;
 }
+void     RsGenExchange::setSyncPeriodAll(uint32_t age_in_secs)
+{
+	if(mNetService != NULL)
+		return mNetService->setSyncAgeAll(age_in_secs) ;
+	else
+		std::cerr << "(EE) No network service available. Cannot set storage period. " << std::endl;
```
PhenomRetroShare, line 2008: Please, use RsErr()

libretroshare/src/gxs/rsgxsnetservice.cc
```diff
@@ -4729,6 +4729,48 @@ void RsGxsNetService::setKeepAge(const RsGxsGroupId &grpId, uint32_t age_in_secs
...
+	{
+		if(it->second.msg_keep_delay != age_in_secs)
+		{
+			it->second.msg_keep_delay = age_in_secs;
+			if(age_in_secs != 0)
+			{
+				if(it->second.msg_req_delay == 0 || it->second.msg_keep_delay < it->second.msg_req_delay)
+					it->second.msg_req_delay = it->second.msg_keep_delay ;
+			}
+			IndicateConfigChanged();
```
PhenomRetroShare: You indicateConfigChanged for each?

retroshare-gui/src/gui/settings/GroupFrameSettingsWidget.cpp
```diff
@@ -81,3 +106,47 @@ void GroupFrameSettingsWidget::saveSettings()
 		NotifyQt::getInstance()->notifySettingsChanged();
 	}
 }
+
+void GroupFrameSettingsWidget::saveSyncAllValue()
+{
+	uint32_t t = ui->cmbSync->currentData().toUInt() * 86400;
```
PhenomRetroShare, line 112: I prefer 24*60*60

retroshare-gui/src/gui/settings/GroupFrameSettingsWidget.ui
```diff
@@ -40,6 +40,72 @@
      </property>
     </widget>
    </item>
+   <item>
+    <spacer name="verticalSpacer_2">
```
PhenomRetroShare, line 44: Specify item name, don't use default one. How can we know reading text where is it?
And this help diff file block move.

retroshare-gui/src/gui/settings/GroupFrameSettingsWidget.ui
```diff
@@ -40,6 +40,72 @@
...
+      </size>
+     </property>
+    </spacer>
+   </item>
+   <item>
+    <layout class="QHBoxLayout" name="horizontalLayout">
+     <item>
+      <widget class="QLabel" name="lblSync">
+       <property name="text">
+        <string>Change sync period for all subcriptions</string>
```
PhenomRetroShare, line 64: Add a tooltip to say that new subscription will not take care of that.

csoler:
There's two ways to see that problem:
1 - iterating over all subscribed groups and change the sync/store numbers. This is what is proposed in this PR and has some drawbacks: it's not a default value and new subscriptions will not follow it. As a consequence, when changing all groups and afterward setting some of them to a specific value, updating newly subscribed groups will either need to update them explicitely, or update all again, and set the specific group times again.
2 - change the default value. The effect is that times that are set to Default, will all be changed, and newly subscribed groups will as well. Groups that have been explicitly changed later on will keep their values.
Seems to me that 2 is much more convenient.

RetroPooh:
1 This whole pr is not for people who carefully select values for every single sub. This mostly for who subscribed a lot and just after that realized the need to extend values.
2 Adjusting defaults for new subs was considered harder to do and so postponed, i'm going to do it later. Probably.

retroshare-gui/src/gui/settings/GroupFrameSettingsWidget.cpp
PhenomRetroShare: * 24x60x60

defnax:
nice idea

PR #1993, opened 2020-06-01
Fix build with Qt 5.15+

PhenomRetroShare:


retroshare-gui/src/gui/gxs/GxsIdDetails.cpp
```diff
@@ -18,21 +18,23 @@
...
-#include <iostream>
+#include <QApplication>
+#include <QMutexLocker>
 #include <QPainter>
+#include <QPainterPath>
+#include <QThread>
+#include <QTimerEvent>
+
+#include <iostream>
+#include <math.h>
```
G10h4ck: now that you touched it plase include `cmath` instead of `math.h`

PhenomRetroShare: Done.

G10h4ck:
Thanks!

Issue #1837, opened 2020-03-28
UPNP related crash in RetroShare-0.6.5 + libupnp-1.6.*


G10h4ck:
```
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
Caught signal 6 (SIGABRT)
stack trace:
retroshare(+0xa9eb94)[0x55603529ab94]
retroshare(_ZN15CrashStackTrace12abortHandlerEi+0xeb)[0x55603529b00a]
/lib64/libc.so.6(+0x38f80)[0x7f63d0607f80]
/lib64/libc.so.6(gsignal+0x141)[0x7f63d0607ec1]
/lib64/libc.so.6(abort+0x121)[0x7f63d05f1535]
/usr/lib/gcc/x86_64-pc-linux-gnu/9.2.0/libstdc++.so.6(+0xa069e)[0x7f63d09a369e]
/usr/lib/gcc/x86_64-pc-linux-gnu/9.2.0/libstdc++.so.6(+0xcf7a6)[0x7f63d09d27a6]
/usr/lib/gcc/x86_64-pc-linux-gnu/9.2.0/libstdc++.so.6(+0xcf7f1)[0x7f63d09d27f1]
/usr/lib/gcc/x86_64-pc-linux-gnu/9.2.0/libstdc++.so.6(+0xcfa24)[0x7f63d09d2a24]
/usr/lib/gcc/x86_64-pc-linux-gnu/9.2.0/libstdc++.so.6(_ZSt19__throw_logic_errorPKc+0x3d)[0x7f63d09a5dae]
retroshare(_ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEE12_M_constructIPKcEEvT_S8_St20forward_iterator_tag+0xcc)[0x556035e340ac]
retroshare(_ZN17CUPnPControlPoint21PrivateAddPortMappingER16CUPnPPortMapping+0x56)[0x556035d164ce]
retroshare(_ZN17CUPnPControlPoint15AddPortMappingsERSt6vectorI16CUPnPPortMappingSaIS1_EE+0x203)[0x556035d15e7b]
retroshare(_ZN11upnphandler10start_upnpEv+0x417)[0x5560359a0729]
retroshare(doSetupUPnP+0x76)[0x5560359a01ff]
/lib64/libpthread.so.0(+0x8427)[0x7f63d0b80427]
[...]

PR #1603, opened 2019-07-24
Fix UPnP

PhenomRetroShare:
For version > 1.6.24 and < 1.8.0
Maybe version numbers to adjust.

PR #1995, opened 2020-06-01
[WIP] Libupnp fix crash

G10h4ck:
Few a few crashes reported at https://github.com/RetroShare/RetroShare/issues/1837

### DIFF
--- a/libretroshare/src/rs_upnp/UPnPBase.cpp
+++ b/libretroshare/src/rs_upnp/UPnPBase.cpp
@@ -36,6 +36,8 @@
 #include "util/rsstring.h"
 #include "rs_upnp/upnp18_retrocompat.h"
 #include "util/rstime.h"
+#include "util/cxx17retrocompat.h"
+#include "util/rsnet.h"
 
 #ifdef __GNUC__
 	#if __GNUC__ >= 4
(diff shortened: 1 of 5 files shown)


libretroshare/src/rs_upnp/UPnPBase.cpp
```diff
@@ -1372,10 +1335,11 @@ int CUPnPControlPoint::Callback(
...
 	{
 #if (UPNP_VERSION > 10624) && (UPNP_VERSION < 10800)
-		std::cerr <<  "********************************************************************************" << std::endl
-		          <<  "***                THIS SHOULD NOT HAPPEN !!!  TELL IT TO DEVS               ***" << std::endl
-		          <<  "*** UPnPBase.cpp CUPnPControlPoint::Callback() UPNP_DISCOVERY_SEARCH_TIMEOUT ***" << std::endl
-		          <<  "********************************************************************************" << std::endl;
+		RsErr() << __PRETTY_FUNCTION__
+		        << " ******** (UPNP_VERSION > 10624) && (UPNP_VERSION < 10800) "
+		           " UPNP_DISCOVERY_SEARCH_TIMEOUT  ***************************"
+		           " THIS SHOULD NEVER HAPPEN !!!  TELL IT TO DEVS ************"
```
G10h4ck: @PhenomRetroShare  This needs clarification https://github.com/RetroShare/RetroShare/issues/1996

PhenomRetroShare: Look at this comment:
https://github.com/RetroShare/RetroShare/pull/1603#issuecomment-515194637
But I don't remember...

csoler:
For the record, it seems that 95% of this PR is about introducing a new debug output. When fixing bugs, it would be really nice if the PRs where only dedicated to fixing that particular bug, because reviewing all these changes makes it much more risky to overlook a mistake.

G10h4ck:
> For the record, it seems that 95% of this PR is about introducing a new debug output. When fixing bugs, it would be really nice if the PRs where only dedicated to fixing that particular bug, because reviewing all these changes makes it much more risky to overlook a mistake.

This is still a WIP (Work In Progress)... Digging into the code I discovered that our way to handle UPnP (at least when using libupnp) is completely flawed...
To fix the bug became a chain reaction :-| and I am quite demotivated about it... I am tempted to build RS with miniupnpc which is not affected, and let it go...

PhenomRetroShare:
LibUPnP needs more reworks since 1.14.0 release.
Read the change log:
https://github.com/pupnp/pupnp/blob/master/ChangeLog#L43

sehraf:
> LibUPnP needs more reworks since 1.14.0 release.

related to #2072

G10h4ck:
Yeah libupnp have been changing it's API a lot lately, I am a bit pissed off of keeping adding `#ifdef` to libretroshare libupnp related code also because this part of our code is quite a mess and have also a few crashes and race conditions because it is badly designed. Even if this time it is easy to make RS compile against latest libupnp how much it will last,and it will keep crashing anyway, I think we should compile against miniupnpc in all platforms it is available until the whole RS libpnp code get overhauled.

G10h4ck:
@sehraf thanks for the snippet in the issue, it looks easy to support last API change for this time, would you mind to test it a bit on Arch which have the latest libupnp? Or your router doesn't have UPnP at all? I might be able to test it in a couple of weeks

sehraf:
As far as I know, my router has no upnp. I have never used it and have never seen any option regarding it.

G10h4ck:
After fiddling with this stuff, my conclusion is that o have a stable UPnP implementation based on libupnp retroshare libupnp interface must be rewritten from scratch as is rely on wrong assumptions, hence i close this PR. For the moment I suggest anyone building RS to use miniupnp or none.

PR #2004, opened 2020-06-04
fixed long-term bug due to sending an empty list of dest keys for GXSS…

csoler:
…ecurity to encrypt an item


jolavillette:
I think that we should also fix the methods that call encrypt without a key.
It should not happen and is probably the sign that something else is wrong.

csoler:
@jolavillette : I agree with you. I'm merging this because it's kind of urgent and we never have crash reports about encrypting with an empty list of keys. But we should fix it too.

PR #2006, opened 2020-06-05
remove retroshare password from json tokens

sehraf:
Currently your RS password ends up as an json api token, which is visible in the settings menu,

This patch does two things:
- add a new parameter when creating the location: `createToken`, this token is added instead of the user:password combination
- remove the token creation from login. This removes the option to add a token during login. So the token must be added before (either manually or during node creating time). This could also use the extra parameter (like above). (I personally don't see any point in constantly adding the same token again and again)

An app (e.g Android) can still insert user:password as a token, as before, using the new parameter.

Additionally `parseToken` is fixed, which previously always returned true.

This PR is untested due to 
```
../../../RetroShare-origin/retroshare-gui/src/gui/gxs/GxsCommentTreeWidget.cpp: In member function 'virtual void MultiLinesCommentDelegate::paint(QPainter*, const QStyleOptionViewItem&, const QModelIndex&) const':
../../../RetroShare-origin/retroshare-gui/src/gui/gxs/GxsCommentTreeWidget.cpp:112:22: error: aggregate 'QPainterPath path' has incomplete type and cannot be defined
  112 |         QPainterPath path ;
      |                      ^~~~
At global scope:
```

### DIFF
--- a/libretroshare/src/jsonapi/jsonapi.cpp
+++ b/libretroshare/src/jsonapi/jsonapi.cpp
@@ -180,6 +180,7 @@ JsonApiServer::JsonApiServer(): configMutex("JsonApiServer config"),
 			std::string errorMessage;
 			bool makeHidden = false;
 			bool makeAutoTor = false;
+			std::string createToken;
 
 			// deserialize input parameters from JSON
 			{
@@ -189,6 +190,7 @@ JsonApiServer::JsonApiServer(): configMutex("JsonApiServer config"),
 				RS_SERIAL_PROCESS(password);
 				RS_SERIAL_PROCESS(makeHidden);
 				RS_SERIAL_PROCESS(makeAutoTor);
+				RS_SERIAL_PROCESS(createToken);
 			}
 
 			// call retroshare C++ API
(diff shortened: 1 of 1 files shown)


libretroshare/src/jsonapi/jsonapi.cpp
```diff
@@ -144,6 +144,9 @@ bool RsJsonApi::parseToken(
 	for(uint32_t i=0; i < tkLen; ++i)
 		if(clear_token[i] == ':') { colonIndex = i; break; }
 
+	if (colonIndex == std::string::npos)
+		return false;
```
G10h4ck: a token without ":" means that is a password-less user which is acceptable as per HTTP authentication and also explicitly asked by few people, so please remove this check

G10h4ck (changes requested):

G10h4ck:
It seems a good idea to me! Once you do the requested changes and are able to test it I'll merge it ASAP

libretroshare/src/jsonapi/jsonapi.cpp
sehraf: Will do

sehraf: done

sehraf:
Tested successfully login and node creation!

G10h4ck:
For locations that were created before and that potentially have API users that use the old method of logging in and using the old token what does happens ?

sehraf:
Nothing, they have an *existing* token (the old one) that they can still use.

G10h4ck (approved):

PR #2009, opened 2020-06-07
[Testers needed] Reworked msys2 build scripts

hunbernd:
* Fixed missing dependencies
* Replaced problematic external tools:
  * sigcheck
  * depends
  * winhttpjs
* Added clang support
* Installer


NyvenZA:
Pulled and compiled using the clean.bat, Build.bat:
used:
clean.bat 64 release 
and 
clean.bat 64 debug
With no errors. 

Used:
build 64 release
and it build without errors,

Pack 64 release worked properly. 

Really like the new Changelog facility.

PR #2011, opened 2020-06-09
Improve API to create locations

G10h4ck:
Improve on top of sehraf work, add documentation for API token params
Deprecate old version and disable it once RS 0.6.6 is released
New method take relevant paramethers explicitely instead of having them
  incapsualted inside a RsLoginHelper::Location struct
New method return a proper error_condition instead of boolean + string
RsLoginHelper::attemptLogin doesn't need manual wrapper anymore

### DIFF
--- a/libretroshare/src/jsonapi/jsonapi.cpp
+++ b/libretroshare/src/jsonapi/jsonapi.cpp
@@ -165,6 +165,7 @@ JsonApiServer::JsonApiServer(): configMutex("JsonApiServer config"),
 	RsThread::setStopTimeout(10);
 #endif
 
+#if !RS_VERSION_AT_LEAST(0,6,6)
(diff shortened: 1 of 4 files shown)


sehraf:
Is it intentional to remove `/rsLoginHelper/attemptLogin`? How does one login without this function?

G10h4ck:
> Is it intentional to remove `/rsLoginHelper/attemptLogin`? How does one login without this function?

Because we don't do token magic anymore on that method we don't need the manually written wrapper anymore, so now this method got the wrapper generated automatically too

https://github.com/RetroShare/RetroShare/pull/2011/files#diff-f1632af92fe3679d15ff946283745552R388

sehraf:
Ah, I've missed that!

PR #2012, opened 2020-06-11
show banned persons in lobby participants list

RetroPooh:


csoler:
I'm not sure about what you want to achieve here: banned identities are meant to disappear asap. So it's likely that if your friends ban them as well, you wont see any data coming from these identities. So showing activity from them is likely to be misleading. Could you elaborate a bit about what your PR is doing?

RetroPooh:
Malicious bans, that's basically it

PR #2014, opened 2020-06-12
RsRandom return an alfanumeric string as per function name

G10h4ck:


csoler:
no!!! You're killing the entropy of the SSL passwd here, and furthermore makes it much easier to copy/remember for whom sees on screen because of keyring "features" of some linux distributions. If ou need a random alphanumeric string of less than 6.5 bits of entry per character, please create a new function. 
I agree that the name "alphanumeric" here is misleading, but the goal is really to have any printable character in the string. That makes 93 of them.

G10h4ck:
Heheheheh I saw that function was used also for generating the SSL password so I did know this wasn't acceptable, but the idea of making you shiver about this was fun and I pushed this one waiting for your comment :tongue:
I am pushing again to this PR so the function with misleading name get deprecated but behave exactly as before to not break stuff, plus two functions with proper names are added ;)

csoler:
Plz update all previous calls to alphaNumeric into the new method that actually do the same thing now. Otherwise you reduce some security. A global search will tell you. Thx

libretroshare/src/util/rsrandom.h
```diff
@@ -46,9 +46,19 @@ class RsRandom
 
 	static bool     seed(uint32_t s);
 
-	static std::string random_alphaNumericString(uint32_t length);
 	static void        random_bytes(uint8_t* data, uint32_t length);
 
+	/** Return a random alphanumeric string of the given lenght */
```
csoler: I would give in the comment the charset that is used, for more clarity.

G10h4ck:
> Plz update all previous calls to alphaNumeric into the new method that actually do the same thing now. Otherwise you reduce some security. A global search will tell you. Thx

The calls to the old method are routed to printable so it is the same as before the PR ;)

```
	static inline std::string random_alphaNumericString(uint32_t length)
	{ return printable(length); }
```

PR #2015, opened 2020-06-15
Fixing distant channel search

csoler:


csoler:
The bug that causes a crash is due to a bad handling of info in ChannelPostsWidget. PR 1994 will take care of this. Still, this PR fixes a few useful things.

felisucoibi:
btw could be nice distant forums too

G10h4ck:
I have tested this and enabled JSON API, made a PR here https://github.com/csoler/RetroShare/pull/16  the extra commits you see there are already included in master so it is not really unrelated work ;)

PR #2018, opened 2020-06-23
Reduce memory usage due to copying in RsQThreadUtils::postToObject

G10h4ck:
Heaptrack reported 141MB of RAM where used just by IdDialog::updateCircles most proably due to the group metadata list being copied accross lambdas and threads more then necessary.
Use an unique_ptr to safely avoid copying of big structure around.

### DIFF
--- a/retroshare-gui/src/gui/Circles/CreateCircleDialog.cpp
+++ b/retroshare-gui/src/gui/Circles/CreateCircleDialog.cpp
@@ -24,6 +24,7 @@
 #include <QMenu>
 
 #include <algorithm>
+#include <memory>
 
 #include <retroshare/rspeers.h>
 #include <retroshare/rsidentity.h>
@@ -696,34 +697,32 @@ void CreateCircleDialog::loadIdentities()
 {
 	RsThread::async([this]()
 	{
-        std::list<RsGroupMetaData> ids_meta;
+		std::list<RsGroupMetaData> ids_meta;
 
 		if(!rsIdentity->getIdentitiesSummaries(ids_meta))
 		{
-			std::cerr << __PRETTY_FUNCTION__ << " failed to retrieve identities ids for all identities" << std::endl;
+			RS_ERR("failed to retrieve identities ids for all identities");
 			return;
-        }
-        std::set<RsGxsId> ids;
+		}
 
-		for(auto& meta:ids_meta)
-			ids.insert(RsGxsId(meta.mGroupId)) ;
+		std::set<RsGxsId> ids;
+		for(auto& meta:ids_meta) ids.insert(RsGxsId(meta.mGroupId));
 
-        std::vector<RsGxsIdGroup> id_groups;
-
-        if(!rsIdentity->getIdentitiesInfo(ids,id_groups))
+		auto id_groups = std::make_unique<std::vector<RsGxsIdGroup>>();
+		if(!rsIdentity->getIdentitiesInfo(ids, *id_groups))
 		{
-			std::cerr << __PRETTY_FUNCTION__ << " failed to retrieve identities group info for all identities" << std::endl;
+			RS_ERR("failed to retrieve identities group info for all identities");
 			return;
-        }
+		}
 
-        RsQThreadUtils::postToObject( [id_groups,this]()
+		RsQThreadUtils::postToObject(
+		            [id_groups = std::move(id_groups), this]()
 		{
 			/* Here it goes any code you want to be executed on the Qt Gui
 			 * thread, for example to update the data model with new information
 			 * after a blocking call to RetroShare API complete */
 
-			fillIdentitiesList(id_groups) ;
-
+			fillIdentitiesList(*id_groups);
 		}, this );
 	});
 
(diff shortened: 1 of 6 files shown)


csoler:
impressive gain! But it's only temporary memory right? Anyway, it is always helpful to save it.
For the record, some similar code parts already do it using pointers. Seems to me that using smart pointers here (and std::move) only adds unnecessary code complexity  as compared to allocating a pointer to the list and deleting it inside the posted method. See for instance GxsForumModel.cpp:820

G10h4ck:
> impressive gain! But it's only temporary memory right?
> Anyway, it is always helpful to save it.

Yeah but it seems that stuff is called often enough so temporary becomes more or less always


> For the record, some similar code parts already do it using pointers.

Yeah I saw them and was tempted to refactor them to unique_ptr too but restrained myself ;-)

> Seems to me that using smart pointers here (and std::move) only adds unnecessary code complexity as compared to allocating a pointer to the list and deleting it inside the posted method. See for instance GxsForumModel.cpp:820

Nope it make the code safer, unique_ptr have no performance hit, the compiler just make sure the stuff is deleted, with the manual `delete` approach we should remember to delete it in every case even when we call return because of some error, and also it is easier to mess it due to some code change. `unique_ptr` also it would help to avoid an extra copy safely in some cases with very little refactoring (left a few comments in the code about that). The BIG BIG BIG plus about `unique_ptr` is that when we debug memory stuff we are 100% sure there is no leak without having to think about it as the compiler does the job for us.

csoler:
> Nope it make the code safer, unique_ptr have no performance hit, the compiler just make sure the stuff is deleted, with the manual `delete` approach we should remember to delete it in every case even when we call return because of some error, and also it is easier to mess it due to some code change. `unique_ptr` also it would help to avoid an extra copy safely in some cases with very little refactoring (left a few comments in the code about that). The BIG BIG BIG plus about `unique_ptr` is that when we debug memory stuff we are 100% sure there is no leak without having to think about it as the compiler does the job for us.

No it doesn't. You assume that:
- shared pointers are correctly implemented on all plateforms (we use windows remember?)
- shared pointers are thread safe. Maybe they are, maybe not. Maybe it depends on the specific implementation of each platform. See for instance https://stackoverflow.com/questions/11482580/is-unique-ptr-thread-safe, and this https://news.ycombinator.com/item?id=11698784
- all the mutex bollux that is needed to make them thread-safe is negligible.

In any case, since the pointer is allocated line X and deleted line X+10, it's *much safer* to use pointers since you do not have to assume the above. Furthermore, with pointers
- the code is *much easier* to understand
- the memory management accross threads is much more clear.
- the code is *much easier* to debug. Last time we had to debug a piece of code where you put these things it was a nightmare. Ask sehraf.

Generally, I'm not against shared pointers, but only if they are needed. Here they are clearly not needed.

G10h4ck:
It is not that I assume it is that the compilers we use on every platforms are fully C++14 compliant

>     * shared pointers are correctly implemented on all plateforms (we use windows remember?)
> 
>     * shared pointers are thread safe. Maybe they are, maybe not. Maybe it depends on the specific implementation of each platform.

This PR make no use of `shared_ptr`!

`unique_ptr` is a much simpler beast, it's implementation t's just that the pointed object is destroyed once the pointer get destroyed (aka goes out of scope)

> See for instance https://stackoverflow.com/questions/11482580/is-unique-ptr-thread-safe,

The linked example is clearly wrong! Is not the `unique_ptr` to be thread unsafe but the code written by the guy is just wrong... it captures the `unique_ptr` by reference on two different threads and then attempt to move it on both, then it is of course wrong to expect the pointer to be valid on both threads! As the name clearly states `unique_ptr` have one and only one owner.

> and this https://news.ycombinator.com/item?id=11698784
>
>     * all the mutex bollux that is needed to make them thread-safe is negligible.

Both those two things apply to `shared_ptr` not to `unique_ptr`


>     * the code is _much easier_ to debug. Last time we had to debug a piece of code where you put these things it was a nightmare. Ask sehraf.

I would like to see a concrete example in which `unique_ptr` made the code a nightmare to debug...


> Generally, I'm not against shared pointers, but only if they are needed. Here they are clearly not needed.

Sure, in fact `unique_ptr` are used and not `shared_ptr`

csoler:
I stand to all I said replacing "shared_ptr" with "unique_ptr". You cannot deny that the code is simpler to understand without all these fancy things and I do not understand why you insist on using them in places where they are actually not needed. Our goal is to make the code simpler and easier to understand. If you want new devs to come and help, these type of fancy modifications is not going to help.

G10h4ck:
Frankly since a few years new developers if any have been attracted by the fancy modification I have made, like JSON API and a few more not by old style wild dangling pointers we have all around the code that have produced the current situation where RetroShare goes from 0 to 5GB of squatted RAM in a couple of hours, ad we haven't been able to discover where all that memory gets eaten after weeks if not months of digging...


> I stand to all I said replacing "shared_ptr" with "unique_ptr".

Most of the arguments you provided may apply to `shared_ptr` but they definitely don't  to `unique_ptr`

PR #2021, opened 2020-06-25
service descriptions in permission matrix

RetroPooh:


retroshare-gui/src/gui/settings/RSPermissionMatrixWidget.cpp
```diff
@@ -531,6 +533,40 @@ void RSPermissionMatrixWidget::paintEvent(QPaintEvent *)
   _max_height = S*fMATRIX_START_Y + (peer_ids.size()+3) * S*fROW_SIZE ;
   _max_width  = matrix_start_x + (service_ids.size()+3) * S*fCOL_SIZE ;
 
+	if(n_row_selected == -1)
+	{
+		for(uint32_t i=0;i<service_ids.size();++i)
+		{
+			if(_current_service_id == service_ids[i])
+			{
+				uint16_t serviceOrig = (uint16_t)((service_ids[i] - (((uint32_t)RS_PKT_VERSION_SERVICE) << 24) ) >> 8);
```
csoler: not so nice. Normally it's better to do serviceOrig=(service_ids[i]>>8) & 0xffff; It avoids substracting uint32_t values.

csoler (changes requested):
Nice contribution. I tried it and noticed that the blox you display is always at the same place and extends far beyond the actual size of the text. This is the opposite of how the text for switches displays: it fits the text and shows up next to the mouse. If possible, could you stick to that display style? There's also one small issue in the code (see comments).

retroshare-gui/src/gui/settings/RSPermissionMatrixWidget.cpp
RetroPooh: ok, your formula works, but why are we ok with dropping RS_PKT_VERSION_SERVICE ?

csoler: RS_PKT_VERSION_SERVICE is just a tag at the start of the RsItem ID. It should be discarded to leave the service ID.

sehraf:
This would not have happened with dedicated types and functions for converting :wink:

retroshare-gui/src/gui/settings/RSPermissionMatrixWidget.cpp
sehraf: This would not have happened with dedicated types and functions for converting :wink:

csoler:
thx! Looks very nice.

PR #2031, opened 2020-07-15
[Channels] Fixing posts tab selection for rs links

defnax:


NyvenZA:
Does this fix the issue where by you:
* Select a channel
* select a post that has a comment. 
* Click the comments tab at the bottom of the channel 
* Click on a different channel

Now the comment of the original post is still showing. 

Doesn't happen with the other tabs: Details and Files.

defnax:
> Does this fix the issue where by you:
> 
> * Select a channel
> * select a post that has a comment.
> * Click the comments tab at the bottom of the channel
> * Click on a different channel
> 
> Now the comment of the original post is still showing.
> 
> Doesn't happen with the other tabs: Details and Files.

this needs to be fixed by cyril or some one else its a different issue seems something not clearing or update there.
my fix is for navigate feature (rs links), for posts to show the right tab when using a rs link for a channel post.

retroshare-gui/src/gui/gxschannels/GxsChannelPostsWidgetWithModel.cpp
```diff
@@ -808,8 +808,15 @@ void GxsChannelPostsWidgetWithModel::insertChannelDetails(const RsGxsChannelGrou
...
-	link = RetroShareLink::createMessage(group.mMeta.mAuthorId, "");
-	ui->infoAdministrator->setText(link.toHtml());
+	if (!group.mMeta.mAuthorId.isNull())
+	{
+		link = RetroShareLink::createMessage(group.mMeta.mAuthorId, "");
+		ui->infoAdministrator->setText(link.toHtml());
+	}
+	else
+	{
+		ui->infoAdministrator->setText(tr("Unknown"));
```
csoler: It shouldn't be "unknown": when the author id is null, it means there is no author at all. So I suppose that a more appropriate text is "[No author]"

defnax:
seems this needs one more fix, channel links dont show more details tab...  

i dont know yet how to handle when channel link used to show details tab & when channel msg link posts tab

csoler:
I merged it into PR 2026, also fixing the conflicts at the same time.

PR #2043, opened 2020-08-15
attempt to remove the hardcoded background colors for channel thumbnails

defnax:
the hardcoded colors doesnt allow to set right selection/none selection(background) colors for Thumbnails on different stylesheets like on dark stylesheet it a big issue.

this is not working yet im not sure why... i get always black background, maybe some one know it? 
@chelovechishko  @PhenomRetroShare maybe you know it?

### DIFF
--- a/retroshare-gui/src/gui/gxschannels/GxsChannelPostsWidgetWithModel.cpp
+++ b/retroshare-gui/src/gui/gxschannels/GxsChannelPostsWidgetWithModel.cpp
@@ -111,9 +111,9 @@ void ChannelPostDelegate::paint(QPainter * painter, const QStyleOptionViewItem &
 	QPixmap pixmap(w.size());
 
     if((option.state & QStyle::State_Selected) && post.mMeta.mPublishTs > 0) // check if post is selected and is not empty (end of last row)
-		pixmap.fill(QRgb(0xff308dc7));	// I dont know how to grab the backgroud color for selected objects automatically.
+		pixmap.fill(mBackgroundColorSelected );	// I dont know how to grab the backgroud color for selected objects automatically.
 	else
-		pixmap.fill(QRgb(0x00ffffff));	// choose a fully transparent background
+		pixmap.fill(mBackgroundColorNotSelected);	// choose a fully transparent background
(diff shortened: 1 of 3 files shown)


chelovechishko:
For some reason mail answer is not being displayed here, so (duplicate?):
For me it was just working (right after Phenom's work (the day after i pulled sources)).

defnax:
> For some reason mail answer is not being displayed here, so (duplicate?):
> For me it was just working (right after Phenom's work (the day after i pulled sources)).

can you check if i maked mistakes? you mean the code is right?

chelovechishko:
I meant my color changes works. Yours "qproperty stuff" is good as i
can see. Do not sure about pixmap.fill as there was alpha channel
before. That's about what was added.

csoler:
it's essential to keep the alpha channel. I suppose that it is possible to copy the RGB part of the color property into a RGBA with the same alpha.

defnax:
> it's essential to keep the alpha channel. I suppose that it is possible to copy the RGB part of the color property into a RGBA with the same alpha.

i have tested with rgba it doesnt worked yet

im thinking that this doesnt gets the colors properly from the qss.default file
`		QColor mBackgroundColorSelected;
		QColor mBackgroundColorNotSelected;`
```
im not sure about ChannelPostDelegate its not a standalone class which is defined in a *.pro file .

ChannelPostDelegate
{
	qproperty-backgroundColorSelected: rgba(120, 153, 34, 1);
	qproperty-backgroundColorNotSelected: rgba(255, 255, 255, 1);
}
```

defnax:
but without the qproperty it works 

```
  pixmap.fill(qRgba(120, 153, 34, 1));	
else
  pixmap.fill(qRgba(255, 255, 255, 1));
```

defnax:
maybe the ChannelPostDelegate is needed as standalone class which will be defined in the *.pro file
possible the customizeable colors will work then? but im not sure on other classes all are standalone where working the colors from a qss file... but the ChannelPostDelegate is only defined into a another file like in GxsChannelPostsWidgetWithModel

PR #2045, opened 2020-08-17
file upload stats

RetroPooh:
for current session only, for now.
TODO for future: add context menu option to show only uploaded files, similar to show empty in friend shares

### DIFF
--- a/libretroshare/src/ft/ftserver.cc
+++ b/libretroshare/src/ft/ftserver.cc
@@ -1272,6 +1272,7 @@ bool	ftServer::sendData(const RsPeerId& peerId, const RsFileHash& hash, uint64_t
 #ifdef SERVER_DEBUG
 	FTSERVER_DEBUG() << "ftServer::sendData() to " << peerId << ", hash: " << hash << " offset: " << baseoffset << " chunk: " << chunksize << " data: " << data << std::endl;
 #endif
+
 	while(tosend > 0)
 	{
 		//static const uint32_t	MAX_FT_CHUNK  = 32 * 1024; /* 32K */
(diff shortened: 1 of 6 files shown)


NyvenZA:
Tried this version and I see it uses one column to store the upload could and the amount of data uploaded. Nice! 👍 
image

This shows you the upload count in the top row and the total size. Per file you show the total size. 
Would it be possible to show the count and size in separate columns to allow sorting per size and count. 

Some questions on where this is going.
* Would this type of information be available per node as well?
* Will there be some kind of stats that you can see per interval ( Day, week, month, all time)

I know there type of stats [age that was started, but it is unfinished:
https://github.com/RetroShare/RetroShare/blob/master/retroshare-gui/src/gui/unfinished/StatisticDialog.cpp

RetroPooh:
There is no count per-file, as data sent in chunks and i don't think it's reasonable (or even possible) to make some logic to know if file uploaded completely.
Files mostly go in tunnels, so anonymous. It is possible to count if they go f2f, it is also possible to count if tunnel goes thru some friend and sum it. But do we need this?
It is also possible to summarize over some time periods. I just wanted to have per session and overall.

csoler:
I agree with RetroPooh. Besides, storing all that information (nodes x files) will take *a lot* of space, for not much of an improvement. Seems to me that storing how much data is uploaded from each file is enough. It will give you a statistical estimate of how much of a file has already spread.

felisucoibi:
+1

defnax:
nice idea, what happend with this?
@csoler

defnax:
hi, why this not implement we need this kind stats we has only down and speed

csoler:
I'll take a look at this asap.

PR #2046, opened 2020-08-18
Add proper header path for 'retroshare-gui/' headers

drbob:
This is simplifies the include directories needed for plugins.


csoler:
For the record we keep "" for local files and <> for system files. Do not bother changing all of them anyway.

PR #2047, opened 2020-08-18
extrafiles - fix links, add age, open folder, unshare icon

RetroPooh:


csoler:
Plz fix conflicts!

retroshare-gui/src/gui/RemoteDirModel.cpp
```diff
@@ -521,7 +521,19 @@ QVariant TreeStyle_RDM::displayRole(const DirDetails& details,int coln) const
...
 			case COLUMN_AGE:
-				return  (details.type == DIR_TYPE_FILE)?(misc::timeRelativeToNow(details.max_mtime)):QString();
+			{
+				if(details.type == DIR_TYPE_FILE)
+					return misc::timeRelativeToNow(details.max_mtime);
+				else if(details.type == DIR_TYPE_EXTRA_FILE)
+				{
+					FileInfo fi;
+					if (rsFiles->FileDetails(details.hash, RS_FILE_HINTS_EXTRA , fi))
+						return misc::timeRelativeToNow((rstime_t)fi.age-(30 * 3600 * 24)); // AFI_DEFAULT_PERIOD
```
csoler, line 531: I don't understand the idea beyond the 30 days shift here. What is the goal?

RetroPooh: Because stored value is not add time (as normally with shares) but time when to forget it instead.
https://github.com/RetroShare/RetroShare/blob/master/libretroshare/src/ft/ftextralist.cc#L160

csoler: ok thx. Then it's probably a good idea to replace the numbers by the AFI_DEFAULT_PERIOD variable, to keep consistency accross the code when that value changes.

RetroPooh: There is also GXSCHANNEL_STOREPERIOD may be used as a value for erase time (now the two equals), and i don't see how to count both of them - once the file added we don't know where it came from and what value to subtract. So maybe leave it like that for now?

PR #2055, opened 2020-08-31
Update bdbood.txt plus add a script to keep it fresh

G10h4ck:


libbitdht/src/bitdht/bdboot_generate.sh
```diff
@@ -0,0 +1,55 @@
...
+cat /tmp/bdboot_generated.txt | sort -u > bdboot.txt
+--------------------------------
+README
+
+function check_dht_host()
+{
+	mHost="$1"
+	mPort="$2"
+	
+	sudo nmap -oG - -sU -p $mPort $mHost | grep open | \
```
csoler, line 41: Seems overkill to go admin for testing each server. Maybe you can keep sudo outside the script, so you call sudo once for the entire script. 

Also nmap is known to be pretty aggressive.

G10h4ck: `nmap` needs to be run as root to run that specific test, with the options provided it just test that UDP port for some kind of answer, so the test which is run is not aggressive ;)
I agree it is not the most elegant solution but it is the best we have ATM ;)
Putting `sudo` there is to minimize the code that run as root, otherwise the whole script would run as root :-|

PR #2073, opened 2020-10-05
Compatibility with Tor v3, with retrocompatibility for existing nodes

csoler:


csoler:
Needs to be tested. 
The PR does the following:
- it reads existing keys, but always get the tor ID from tor (which tor recomputes from the key) instead of reading it from disc. It is still stored on disc for practical UI reasons.
- uses a new format for keys in hidden_node/private_key that is oblivious of the key details, except for the type. The key string iswhat is sent/received to/from tor.
- all tor key parsing/signing/validating removed from the TorManager since it wasn't used
- when generating a new tor key, NEW:BEST is sent, meaning that Tor will probably use a v3 key (based on EC 25519). If only v2 is available, it will generate a RSA1024 key.

Normally, this PR should be backward compatible:
- the new code still starts on old v2 keys (stored with explicit DER format). The private_key file will be left untouched.
- the new code generates a v3 key when no key is available. That means if you want to upgrade your tor node to v3, just delete the private_key file (or rename it). I haven't tested this. Also clear nodes will not find you anymore before you send them your new tor ID.

PR #2079, opened 2020-10-16
[WIP] save chosen ID in posted

csoler:
Fixing RC1 glitches


defnax:
but the load and save from Boards Composer still not yet implemented.
i saw you moved only storing/load to the Models id combobox .

Store-> if i changed on boards composer the identity then update for the Models ID Combobox.
Load-> Load the Identity which is saved on the Boards composers Combobox.

Or you want do that soon?

PR #2089, opened 2020-10-28
Fixing the Post buttons to get disabled when the limit has been exceeded

defnax:


retroshare-gui/src/gui/gxs/GxsCreateCommentDialog.cpp
```diff
@@ -26,16 +26,19 @@
 #include <QMessageBox>
 #include <iostream>
 
+static const uint32_t MAX_ALLOWED_GXS_MESSAGE_SIZE = 199000;
```
csoler, line 29: This is probably far too much for a comment. I mean, it's not critical either, but we dont want comments to be too large either, otherwise it becomes impossible to properly display them.

defnax: you can change it...we will allow pictures on comments posts? 
but what i dont like that the pics takes maximum width/height into a comment post which breakes the layout.

PR #2090, opened 2020-10-30
Fix the Windows build failing

amir28:
:)
This problem was causing the 404 Errors. shoud be fixed now.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,7 +16,7 @@
 
 # version format
 version: RetroShare-git-{branch}-{build}
-
+image: Visual Studio 2019
 # you can use {branch} name in version format too
 # version: 1.0.{build}-{branch}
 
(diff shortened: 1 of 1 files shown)


amir28:
It failed like before :(
Due to check signature agree or no!
I guess use pacman -Sy**u** may solve this signature problem but I'm not sure

csoler:
thx for looking into this. Apparently it's more complicated, since your PR doesn't pass the appveyor test either. Is it -Syu that you should use? I'm not sure. See https://github.com/msys2/MSYS2-packages/issues/2058

hunbernd:
https://www.msys2.org/news/#2020-06-29-new-packagers

But instead of fixing the currently installed msys2 version, I suggest to run the msys2 build script from RS source tree.
It installs a working msys2 version and every other stuff needed.

amir28:
Yes, You are right and thanks for helpful links.
after some searching I found an example which is very straightforward.
https://github.com/analogdevicesinc/libiio/pull/578

What is your decision? modify this existing msys or create reinstall msys?

csoler:
the solution by AlexandraTrifan seems ok. You could test whether it works for RS.

amir28:
It seems I can't do anything to solve the signature problem because zstd-1.4.1-1-x86_64.pkg.tar.xz is no longer in msys2 repository and newer pacman and zstd versions signs with new key.
I'm so sorry. I think can't do anything more :(
I'm also not familiar with how to run msys2 build script from RS source tree so it's better to you do it. 
But I think it should be a better way than running msys2 build script.

I do not close this PR to be able to test some another tests(If with closed PR also can continue testing inform me to close this).

Thanks

amir28:
I think I found. The maintainer(or who creates Appveyor build for this repo) must go to her/his Appveyor(ci.appveyor.com) and select this repo(RetroShare). then go to settings related that and select Environment and in Build worker image select "Visual Studio 2019".
This uses new image with updated msys2

csoler:
Looks like you found it! Thx a lot for this.

csoler:
would you mind closing this PR and make a new one with just 1 commit? There's so little changes in the end that it's a bit confusing to merge all these commits.

amir28:
> would you mind closing this PR and make a new one with just 1 commit? There's so little changes in the end that it's a bit confusing to merge all these commits.

Yes, I'll do it :)

chelovechishko:
You could just force push to left only needed commit.

PR #2091, opened 2020-10-30
[WIP] Bug fixing before 0.6.6

csoler:


retroshare-gui/src/gui/gxsforums/GxsForumModel.cpp
```diff
@@ -445,15 +445,15 @@ QVariant RsGxsForumModel::textColorRole(const ForumModelPostEntry& fmpe,int /*co
     if( (fmpe.mPostFlags & ForumModelPostEntry::FLAG_POST_IS_MISSING))
         return QVariant(mTextColorMissing);
 
+    if(fmpe.mPostFlags & ForumModelPostEntry::FLAG_POST_IS_PINNED)
```
chelovechishko, line 448: Why pinned post couldn't be in (un)read shape (via color change) at all?

retroshare-gui/src/gui/gxsforums/GxsForumModel.cpp
```diff
@@ -595,8 +595,8 @@ QVariant RsGxsForumModel::pinnedRole(const ForumModelPostEntry& fmpe,int /*colum
 
 QVariant RsGxsForumModel::backgroundRole(const ForumModelPostEntry& fmpe,int /*column*/) const
 {
-    if(fmpe.mPostFlags & ForumModelPostEntry::FLAG_POST_IS_PINNED)
-        return QVariant(QBrush(mBackgroundColorPinned));
+//    if(fmpe.mPostFlags & ForumModelPostEntry::FLAG_POST_IS_PINNED)
+//        return QVariant(QBrush(mBackgroundColorPinned));
```
chelovechishko, line 599: Then in stylesheets qproperty need to be commented out too?

retroshare-gui/src/gui/gxsforums/GxsForumModel.cpp
csoler: It doesn't mean much that a pinned post is unread or not, and the place to put the unread icon is occupied by the pin icon. Anyway, it's still possible to use bold of course.

chelovechishko: From my point of view whether post (any) read or not is quite important. When there are many pinned posts and one of them not yet read, guessing what exactly is, is not something good for me (whether i just have not time to read yet, or i saved it in unread state for the purpose).

csoler: Agreed. Thx for checking.

chelovechishko: Thx for agreeding.

defnax:
Im wondering why you removed the background color feature for pinned posts? it was nice feature where the color can changed from the qss files standard and darkstyles. which was added by @chelovechishko 

but only the hardcoded colors at the beginning where you added that was not so good.

csoler:
Seemed like a legitimate request to me. In the default style at least, it looks pretty bad. Having it using an icon seemed like a better idea.

defnax:
on my changes i added here too before it doesnt worked not sure if its needed 
image

now i must wait till your PR is merged? my comments changes depends on this PR.

csoler:
@defnax: I will merge this asap. Just waiting for appveyor checks to complete.

PR #2093, opened 2020-11-01
Added back counting on Subscribe button

defnax:


csoler:
Is this compatible with distant search? Normally when you select a channel that is the result of distant search, the subscribe button should be disabled, and the count should be shown because it is meaningless.

defnax:
> Is this compatible with distant search? Normally when you select a channel that is the result of distant search, the subscribe >button should be disabled, and the count should be shown because it is meaningless.

its only for subscribed channels with counting

PR #2097, opened 2020-11-06
Fix RSTreeView Header first Painting

PhenomRetroShare:


csoler:
Damn! Totally my fault. Thx for Fixing it!

PR #2099, opened 2020-11-07
Change Chat History Dialog to no modal.

PhenomRetroShare:
It will be updated when new message comes.

### DIFF
--- a/retroshare-gui/src/gui/chat/ChatWidget.cpp
+++ b/retroshare-gui/src/gui/chat/ChatWidget.cpp
@@ -42,7 +42,6 @@
 #include "gui/settings/rsharesettings.h"
 #include "gui/settings/rsettingswin.h"
 #include "gui/settings/RsharePeerSettings.h"
-#include "gui/im_history/ImHistoryBrowser.h"
 #include "gui/common/StatusDefs.h"
 #include "gui/common/FilesDefs.h"
 #include "gui/common/Emoticons.h"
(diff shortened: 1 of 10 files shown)


retroshare-gui/src/gui/chat/ChatWidget.cpp
```diff
@@ -1607,8 +1606,9 @@ void ChatWidget::deleteChatHistory()
 
 void ChatWidget::messageHistory()
 {
-    ImHistoryBrowser imBrowser(chatId, ui->chatTextEdit, window());
-	imBrowser.exec();
+	if (!imBrowser)
+		imBrowser = new ImHistoryBrowser(chatId, ui->chatTextEdit, this->title, window());
```
csoler, line 1610: I get the idea, but this has a drawback: it keeps the widget in memory even if not used. I guess it's fine for now. If you want to keep a pointer to that to update it, it's going to need to somehow register it. The more elegant way to do that would be to create new rsEvents in history, such that when the history is changed, a rsEvent is posted, and your widget (which should register a event handler in its constructor and de-register it in its destructor) catches that event in order to update its content. This is not a required change but we'll have to do it at some point because the old notify system is now deprecated.

csoler: So if you feel doing this, tell me, and I'll wait for merging this PR. Otherwise we can merge it already and do that later when we get totally rid of p3notify.

PhenomRetroShare: Hello,
I don't know how new rsEvent works. And doing this may change  a lot ImHistoryBrowser. This not the subject of this PR.
We should change it later 0.6.6. Let it in TODO List.

csoler: ok. No worries.

PR #2104, opened 2020-11-09
Add View Source in RSTextBrowser

PhenomRetroShare:


retroshare-gui/src/gui/common/RSTextBrowser.cpp
```diff
@@ -291,3 +296,40 @@ QString RSTextBrowser::anchorForPosition(const QPoint &pos) const
 	}
 	return anchor;
 }
+
+
+#ifndef QT_NO_CONTEXTMENU
```
csoler: What's the goal of QT_NO_CONTEXT_MENU? Is it to be able to disable this feature when you do not need it anymore? If so, I would rather name it "DEBUG_something" and change ifndef into ifdef.

PhenomRetroShare: Look at QTextEdit declaration:
https://code.woboq.org/qt5/qtbase/src/widgets/widgets/qtextedit.h.html#187

csoler: This macro is defined by QT (activated when compiling Qt) to cope with hardware like touch screens where context menus cannot be used. Since RS uses context menus everywhere, and since the GUI is only meant to work on computers, I dont think that you need it.

see https://stackoverflow.com/questions/60143222/when-is-qt-no-contextmenu-defined

PR #2105, opened 2020-11-09
Fix Missing Forum Message Text

PhenomRetroShare:
When asking data for missing message we got this:
More than one entry for this message. Is the database corrupted?

### DIFF
--- a/retroshare-gui/src/gui/gxsforums/GxsForumThreadWidget.cpp
+++ b/retroshare-gui/src/gui/gxsforums/GxsForumThreadWidget.cpp
@@ -97,7 +97,7 @@ class DistributionItemDelegate: public QStyledItemDelegate
 public:
     DistributionItemDelegate() {}
 
-    virtual void paint(QPainter *painter, const QStyleOptionViewItem &option, const QModelIndex &index) const
+    virtual void paint(QPainter *painter, const QStyleOptionViewItem &option, const QModelIndex &index) const override
     {
         if(!index.isValid())
         {
@@ -105,7 +105,7 @@ public:
             return ;
         }
 
-        QStyleOptionViewItemV4 opt = option;
+        QStyleOptionViewItem opt = option;
         initStyleOption(&opt, index);
         // disable default icon
         opt.icon = QIcon();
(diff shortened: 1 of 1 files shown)


retroshare-gui/src/gui/gxsforums/GxsForumThreadWidget.cpp
```diff
@@ -1204,7 +1204,7 @@ void GxsForumThreadWidget::insertMessage()
 
         int current_index = 0 ;
 
-        for(int i=0;i<post_versions.size();++i)
+        for(int i=0;i<static_cast<int>(post_versions.size());++i)
```
csoler, line 1207: Please use for(uint32_t i=0;) or for(size_t i=0;) instead of this static_cast. I know it's the same, but it makes the code easier to read.

retroshare-gui/src/gui/gxsforums/GxsForumThreadWidget.cpp
```diff
@@ -1227,8 +1227,17 @@ void GxsForumThreadWidget::insertMessage()
...
     ui->versions_CB->blockSignals(false) ;
 
-    /* request Post */
-    updateMessageData(mThreadId);
+	/* request Post */
+	bool missing = index.sibling(index.row(),RsGxsForumModel::COLUMN_THREAD_DATA).data(RsGxsForumModel::MissingRole).toBool();
+	if (missing)
+	{
+		// Don't update data for missing message else get multiple entry.
+		setMessageLoadingError(tr("Missing Message:\nThis message is missing. You should receive it later."));
```
csoler, line 1235: You should => you may. Not sure about this. Maybe an english native speaker could advise here.

PhenomRetroShare: You may receive => Vous pouvez recevoir
You should receive => Vous devriez recevoir
But maybe Google is wrong. ;)

retroshare-gui/src/gui/gxsforums/GxsForumThreadWidget.cpp
PhenomRetroShare: i is used after as integer.
But I can change with older (int) cast.

retroshare-gui/src/gui/gxsforums/GxsForumThreadWidget.cpp
csoler: "You should" can  mean that there actually is something to do on your own to fix it. That what I find ambiguous about this sentence.

PR #2107, opened 2020-11-10
Fix GXS (Forum/Channel/Boards) RSLink

PhenomRetroShare:
Have to add ID: tag in Board Filter due to range view.


defnax:
thx rs links for board posts works now nice 👍

PR #2122, opened 2020-11-22
Priority optimization and fixes

jolavillette:
This PR changes the priority levels of RS items. It also add a few fixes in the related code.

The improvement will be mostly visible when the UP bandwidth is heavily used:
- smoother GXS sync (no more outqueue clogging caused by repeated identical GXS sync requests)
- faster opening of distant chat session
- faster start of file download

In details:
- Tunnel opening items use priority 7
- Turtle control items such as Search request, File request, File Map request, Chunk CRC request use priority 6
- Turtle search result items uses priority 6 
- Turtle data items use priority 5
- GXS items use priority 6
- Diffie–Hellman key exchange distant chat item uses priority 7

Please test and comment

Update 25nov2020:
- Answers to File Map and Chunk CRC requests now also use priority 6. These items are neither numerous nor big, and they are required to complete file transfers.

### DIFF
--- a/libretroshare/src/ft/ftserver.cc
+++ b/libretroshare/src/ft/ftserver.cc
@@ -1083,6 +1083,8 @@ bool ftServer::sendTurtleItem(const RsPeerId& peerId,const RsFileHash& hash,RsTu
 		if(!encryptItem(item, hash, encrypted_item))
 			return false ;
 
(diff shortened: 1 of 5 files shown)


csoler:
Seems all ok to me. Any last thoughts before we merge this?

PR #2127, opened 2020-11-24
Add current PID in Lock file.

PhenomRetroShare:


csoler:
bad idea! This lock is here to prevent that different RS instance will write at the same place concurrently.

PhenomRetroShare:
Yes, so you know which one have it...

csoler:
Ah ok, I get it. I thought you'd be using a lock name based on who has the lock ;-) Then it's important to write to the file *after* locking it.

PR #2132, opened 2020-11-27
Replace I2P BOB with SAMv3

### LINKED ISSUE #2087: I2P: Switch from BOB to SAMv3
I2P has deprecated the BOB interface https://geti2p.net/en/docs/api/bob and would like to get all users moved over to SAMv3 https://geti2p.net/en/docs/api/samv3 and set a schedule for removing BOB completely in the next year.

It appears that RetroShare may be the last remaining BOB application with active users. We don't want to break RetroShare for them but we'd like them to have the stronger encryption options and good support offered by SAMv3.

sehraf:
Fixes #2087

Users can switch between SAM and BOB builds, the settings are kept compatible, see `    DEFINES *= RS_I2P_SAM3_BOB_COMPAT`
I've tested Java I2P and C++ i2pd and on Linux and Windows. i2pd needs to be version 2.34 or later.

There is a more informativ post on the developer forum: retroshare://forum?name=Upcoming%20Change%3A%20BOB%20to%20SAMv3%20migration%0D%20&id=8fd22bd8f99754461e7ba1ca8a727995&msgid=abf7ece17146ef48b91aebd7c7f13c731ab0c614


sehraf:
I've fixed all bugs i'm aware of. There is a small bug in libsam3 but I've added some sanity check that take care of it inside RS.

sehraf:
rebased due to 6e6fcc4e1de3ee411f59bd4e01b2f8a7a6290a0d

csoler:
I dont understand why you want to keep the two options BOB and Samv3. It seems that they do the same, and there's no backward compatibility between peers right? As far as I understand it's just a configuration utility for I2P.

sehraf:
> I dont understand why you want to keep the two options BOB and Samv3.

This PR replaced the BOB implementation with a SAMv3 one. Only the (persistent) options are kept backwards compatible, so you can downgrade your RS back to a BOB one. This can deactivated (there is a qmake option) or completely be removed.

> It seems that they do the same, 


They are the same and do the same (beside technical differences). From a peers point of view there is no difference. 

> and there's no backward compatibility between peers right?

It's all I2P and is thus compatible. I2P itself or I2P based connections do not care whether it is BOB or SAM, these are just APIs for interacting with I2P.  In other words peers are fully compatible and there is no visible difference.

> As far as I understand it's just a configuration utility for I2P.

exactly! 
SAMv3 supports different key types (old and new one), while BOB only supports one (the old one).

kodxana:
Any progress on that PR?

sehraf:
> Any progress on that PR?

The branch still merges fine with 0.6.6. I can rebase it though if required.

kodxana:
> > Any progress on that PR?
> 
> The branch still merges fine with 0.6.6. I can rebase it though if required.

Do you think you could build 0.6.6 with SAM support cause it fails on my side.

sehraf:
There was one mismatch, I've rebased and force pushed the branch.

eyedeekay:
Any timeline on this PR? Do you need anything from the I2P side that I can help you with?

csoler:
Sorry sehraf for keep that for so long. It's been fully tested now, and ready to merge. Thx a lot for this.

csoler:
Actually it doesn't compile. Seems that the support lib system is trying to copy the samv3 directory in place, which produces an error and stops compilation.

sehraf:
yep noticed that too with `makepkg`. 

Can you try 
```diff
diff --git a/libretroshare/src/libretroshare.pro b/libretroshare/src/libretroshare.pro
index 30aee9671..c654b2e61 100644
--- a/libretroshare/src/libretroshare.pro
+++ b/libretroshare/src/libretroshare.pro
@@ -1030,8 +1030,8 @@ rs_sam3_libsam3 {
         cd $${RS_SRC_PATH} && ( \
         git submodule update --init supportlibs/libsam3 || \
         true ) && \
-        mkdir -p $${UDP_DISCOVERY_BUILD_PATH} && \
-        cp -r $${LIBSAM3_SRC_PATH}/* $${LIBSAM3_BUILD_PATH} && \
+        mkdir -p $${LIBSAM3_BUILD_PATH} && \
+        (cp -r $${LIBSAM3_SRC_PATH}/* $${LIBSAM3_BUILD_PATH} || true) && \
         cd $${LIBSAM3_BUILD_PATH} && \
         $(MAKE) build
     QMAKE_EXTRA_COMPILERS += libsam3

```
This works here with "normal" `qmake` and `make`

G10h4ck:
> There is a more informativ post on the developer forum: retroshare://forum?name=Upcoming%20Change%3A%20BOB%20to%20SAMv3%20migration%0D%20&id=8fd22bd8f99754461e7ba1ca8a727995&msgid=abf7ece17146ef48b91aebd7c7f13c731ab0c614

never got that post on my node :-\

This PR broke Android build too, I have already committed the fix trivial ;)
Thanks for contributing @sehraf I'll be happy to read from you on RS forums and chat soon, how it is going your rustyshare experimentation?

G10h4ck:
After making it optional it seems it is still broken here https://github.com/RetroShare/RetroShare/blob/master/libretroshare/src/pqi/pqisslpersongrp.cc#L88 putting an `#ifdef` around is ok?

sehraf:
b6596437481fa7fff30fc22d9122904359c0566a should do the trick :+1:

PR #2135, opened 2020-11-28
Still fixing GUI issues for 0.6.6

csoler:


defnax:
On Channels & Boards changing to a new Thumbnail doesnt updates:
* Tree updates the Channel logo on restart only (not before)

Not so importend but not updates too (channels & boards):
* logoLabel doesnt updates, only after switch to different Channel & back

PR #2136, opened 2020-11-29
Added ZoomableLabel for the Channel Post Feeds

defnax:


csoler:
Something's probably missing, because images are deformed, whereas they should be only cropped. See how the Zoomable label is used e.g. in channel thumbnail widget. It's possible that after setPicture() you'd need something like adjustSizes(). I'm not sure.

defnax:
> Something's probably missing, because images are deformed, whereas they should be only cropped. See how the Zoomable label is used e.g. in channel thumbnail widget. It's possible that after setPicture() you'd need something like adjustSizes(). I'm not sure.

i dont know yet to fix that issue adjustSizes() dont helped.

PR #2137, opened 2020-11-30
Add ability to manage canceled token in Gxs.

PhenomRetroShare:


PhenomRetroShare:
@csoler : Conflicts solved.
Can you merge it as I need it to run with valgrind?
Else I've got:
p3IdService::getOwnIds(): own identities are not loaded yet.
and People doesn't populate.

libretroshare/src/gxs/gxstokenqueue.h
```diff
@@ -23,19 +23,22 @@
...
 #include "util/rsthreads.h"
 
 
 struct GxsTokenQueueItem
 {
+public:
 
 	GxsTokenQueueItem(const uint32_t token, const uint32_t req_type) :
-	    mToken(token), mReqType(req_type) {}
+	    mToken(token), mReqType(req_type), mStatus(RsTokenService::PENDING) {}
```
csoler, line 35: I would have used RsTokenService::UNKNOWN, as a way to indicate that it's not initialized.

csoler (commented):
About the "UNKNOWN" thing, I'm not 100% sure it's necessary. I'm just mentioning it because it's preferable that the default value doesn't have an other meaning than "it's not initialized or used yet".

PR #2141, opened 2020-12-03
Fix RsGxsDataAccess req delete when failed.

PhenomRetroShare:
278,432 bytes in 8,701 blocks are still reachable in loss record 7,348 of 7,383
  in RsGxsDataAccess::requestGroupInfo(unsigned int&, unsigned int, RsTokReqOptions const&, std::__cxx11::list<t_RsGenericIdType<16u, false, (RsGenericIdType)4>, std::allocator<t_RsGenericIdType<16u, false, (RsGenericIdType)4> > > const&) in /libretroshare/src/gxs/rsgxsdataaccess.cc:67

### DIFF
--- a/libretroshare/src/gxs/rsgxsdataaccess.cc
+++ b/libretroshare/src/gxs/rsgxsdataaccess.cc
@@ -833,8 +833,8 @@ void RsGxsDataAccess::processRequests()
 			}
 			else
 			{
-				req->status = FAILED;
 				mPublicToken[req->token] = FAILED;
+				delete req;//req belongs to no one now
 #ifdef DATA_DEBUG
 				RsDbg() << "  Request failed. Marking as FAILED." << std::endl;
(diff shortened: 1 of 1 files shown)


libretroshare/src/gxs/rsgxsdataaccess.cc
```diff
@@ -833,7 +833,7 @@ void RsGxsDataAccess::processRequests()
 			}
 			else
 			{
-				req->status = FAILED;
+				delete req;//req belongs to no one now
 				mPublicToken[req->token] = FAILED;
```
csoler, line 836: do it the other way, otherwise req->token will possibly cause an access to freed memory,

PhenomRetroShare: Oups.
Done

PR #2144, opened 2020-12-04
[ Boards ] Removed the fixed size from the thumb label & moved the subscribe button under the title

defnax:


csoler:
Could you change "last post" for "last activity" ? (last post TS actually accounts for all messages, comments and votes, so "activity" is better than "post" I suppose)

PR #2147, opened 2020-12-05
Still bug fixing 0.6.6-RC1

csoler:


defnax:
When you disable the Chat flag in settings
The Network and tray gets a notify but not more on the new Friendslist
The Old Friends List always displayed it.
￼￼

on rs 0.6.5 old friendslist
jpg

PR #2153, opened 2020-12-08
fixed bug in mutexes causing a deadlock in gxs integrity check

csoler:


NyvenZA:
Compile fails on Msys2 Windows. Might be the issue that was fixed in PR2149

[20201209-compile- PR2153.txt](https://github.com/RetroShare/RetroShare/files/5663481/20201209-compile-.PR2153.txt)

NyvenZA:
Merged this patch into the #2149 and it compiled.

NyvenZA:
I have been running this version for 7 hours and the channels have not locked up. 
👍

NyvenZA:
Without the #2149 I can't compile this code in the master.

PR #2159, opened 2020-12-10
[Channel POST News Feeds] Fixed to display 3:4 Thumbnail formats better

defnax:
* Fixed to display 3:4 Thumbnail formats better

### DIFF
--- a/retroshare-gui/src/gui/feeds/GxsChannelPostItem.cpp
+++ b/retroshare-gui/src/gui/feeds/GxsChannelPostItem.cpp
@@ -442,8 +442,10 @@ void GxsChannelPostItem::fill()
 		GxsIdDetails::loadPixmapFromData(mPost.mThumbnail.mData, mPost.mThumbnail.mSize, thumbnail,GxsIdDetails::ORIGINAL);
 		// Wiping data - as its been passed to thumbnail.
 
-		// Check if the Thumbnail is a 3:4 Format.
-		if (thumbnail.width() >= 130 && thumbnail.height() >= 190){
+		// Check if the Thumbnail is a 2:3 Format.
+		float r = thumbnail.width()/(float)thumbnail.height();
+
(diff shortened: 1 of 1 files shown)


retroshare-gui/src/gui/feeds/GxsChannelPostItem.cpp
```diff
@@ -434,18 +434,28 @@ void GxsChannelPostItem::fill()
...
+	int desired_height_3_4 = QFontMetricsF(font()).height() * 12;
 
 	if(mPost.mThumbnail.mData != NULL)
 	{
 		QPixmap thumbnail;
 		GxsIdDetails::loadPixmapFromData(mPost.mThumbnail.mData, mPost.mThumbnail.mSize, thumbnail,GxsIdDetails::ORIGINAL);
 		// Wiping data - as its been passed to thumbnail.
 
+		// Check if the Thumbnail is a 3:4 Format.
+		if (thumbnail.height() >= 150){
```
csoler: how is the height of 150 related to the aspect ratio?

csoler:
If you want to compute the best aspect ratio, it's better to use a consistent formula, such as this one:
        // compute closest aspect ratio
        float r = p.width()/(float)p.height();

        if(r < 0.8)
            setAspectRatio(ChannelPostThumbnailView::ASPECT_RATIO_2_3);
        else if(r < 1.15)
            setAspectRatio(ChannelPostThumbnailView::ASPECT_RATIO_1_1);
        else
            setAspectRatio(ChannelPostThumbnailView::ASPECT_RATIO_16_9);

This code is part of GxsChannelPostThumbnail.cpp. Maybe you could take that code and create a static method in e.g. ChannelPostThumbnailView class that does the job, given the width and height of the image. So everybody can call the same method to ensure a consistent choice of aspect ratios.

PR #2162, opened 2020-12-14
Fix Friend Request Toaster

PhenomRetroShare:


csoler:
The proper way to fix this would be to catch the rsEvent in GUI and from there launch a toaster. Indeed, it doesn't make sense in libretroshare, to request any kind of GUI event. 
Anyway, because the old system is still there, this should be fine for now, but we really need to get rid of these AddPopupMessage, rsNotify, and even toasters because they really look bad.

PR #2185, opened 2020-12-26
[WIP] using shared_ptrs in MetaData cache

csoler:
Replaced GxsMetaData pointers by shared_ptrs in dataaccess and dataservice, so that we can easily disable caching, more freely get rid of cache items, and have a simpler memory management scheme

### DIFF
--- a/libretroshare/src/gxs/rsdataservice.cc
+++ b/libretroshare/src/gxs/rsdataservice.cc
@@ -482,7 +482,7 @@ bool RsDataService::finishReleaseUpdate(int release, bool result)
     return result;
 }
 
-RsGxsGrpMetaData* RsDataService::locked_getGrpMeta(RetroCursor &c, int colOffset,bool use_cache)
+std::shared_ptr<RsGxsGrpMetaData> RsDataService::locked_getGrpMeta(RetroCursor& c, int colOffset,bool use_cache)
 {
 #ifdef RS_DATA_SERVICE_DEBUG
     std::cerr << "RsDataService::locked_getGrpMeta()" << std::endl;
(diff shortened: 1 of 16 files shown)


libretroshare/src/gxs/rsdataservice.cc
```diff
@@ -1340,15 +1337,16 @@ void RsDataService::locked_retrieveGrpMetaList(RetroCursor *c, std::map<RsGxsGro
 
 	while(valid)
 	{
-		RsGxsGrpMetaData* m = locked_getGrpMeta(*c, 0,true);
+        auto m = locked_getGrpMeta(*c, 0,true);
 
-		if(m)
+        if(m != nullptr)
```
G10h4ck, line 1344: You probably alrrady know that the `if(m)` syntax works also with `shared_ptr<T>` ;)

libretroshare/src/gxs/rsdataservice.h
```diff
@@ -38,54 +38,45 @@ class MsgUpdate
...
+    {}
+    virtual ~t_MetaDataCache() = default;
 
     bool isCacheUpToDate() const { return mCache_ContainsAllMetas ; }
     void setCacheUpToDate(bool b) { mCache_ContainsAllMetas = b; }
 
-	void getFullMetaList(std::map<ID,MetaDataClass*>& mp) const { mp = mMetas ; }
-    void getFullMetaList(std::vector<const MetaDataClass*>& mp) const { for(auto& m:mMetas) mp.push_back(m.second) ; }
+    void getFullMetaList(std::map<ID,std::shared_ptr<MetaDataClass> >& mp) const { mp = mMetas ; }
+    void getFullMetaList(std::vector<std::shared_ptr<MetaDataClass> >& mp) const { for(auto& m:mMetas) mp.push_back(m.second) ; }
```
G10h4ck, line 50: If relevant shared_ptr can point to a const object like this `std::shared_ptr<const MetaDataClass>`

G10h4ck (commented):
Using shared_ptr for caches instead of manual memory handling seems a great idea!

PR #2191, opened 2020-12-29
Fixed to use for share on channel as file attachment

defnax:


felisucoibi:
+1 it works with directories?

PR #2213, opened 2021-01-10
bandwith management optimization

jolavillette:
This PR tries to improve the BW management.
A is sending data to B at a rate near the maximum down set by B.
When A sends more data than B can accept because of its down bandwidth limit, the operating system automatically increases the TCP Received buffer size on B.
This results in an increasing latency, up to several seconds, for any packet traveling from A to B, including high priority packets like RTT or Chat items.

How we fix that:
1. pqistreamer was not correctly handling the irregularity observed in traffic flow: sometimes B was discarding some down BW that should have been kept for the next round of data polling. There is now a credit mechanism for that.
2. in pqihandler the unused down bandwidth was not fully allocated to peers, only 50 pct was. That was supposed to make the operation more smooth but actually resulted in the BW not being fully used, which made issue #1 above even worse.
3. in pqihandler also, in order to avoid overwhelming B reception capacity we now limit the up BW on A side to 95 pct of the down limit provided by B via the BwCtrl service.

Please test and comment

### DIFF
--- a/libretroshare/src/pqi/pqihandler.cc
+++ b/libretroshare/src/pqi/pqihandler.cc
@@ -437,43 +437,45 @@ int     pqihandler::UpdateRates()
 	RsDbg() << "UPDATE_RATES pqihandler::UpdateRates mod_index " << mod_index << " in_max_bw " << in_max_bw << " remaining in bw " << in_remaining_bw << std::endl;
 #endif
 
-	/* Allocate only half the remaining in bw, if any, to make it smoother */
-	in_max_bw = in_max_bw + in_remaining_bw / 2;
+	// allocate all remaining in bw
+	in_max_bw = in_max_bw + in_remaining_bw;
 
-	// store current total in and ou used bw 
+	// store current total in and out used bw 
 	locked_StoreCurrentRates(used_bw_in, used_bw_out);
 
 #ifdef UPDATE_RATES_DEBUG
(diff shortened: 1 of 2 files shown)


libretroshare/src/pqi/pqihandler.cc
```diff
@@ -437,43 +437,45 @@ int     pqihandler::UpdateRates()
 	RsDbg() << "UPDATE_RATES pqihandler::UpdateRates mod_index " << mod_index << " in_max_bw " << in_max_bw << " remaining in bw " << in_remaining_bw << std::endl;
 #endif
 
-	/* Allocate only half the remaining in bw, if any, to make it smoother */
-	in_max_bw = in_max_bw + in_remaining_bw / 2;
+	// allocate all remaining in bw
+	in_max_bw = in_max_bw + in_remaining_bw;
```
csoler: That originally was made to smooth the curve of bandwidth allocation. Seconds after seconds, the available bw tends (quite rapidely) to in_max_bw+in_remaining_bw. I have the feeling that this change is what made you use the 0.95 factor below. Is that right?

jolavillette:
Restore the smoothing mechanism for the down BW, with a factor 0.75 instead of 0.5
Restore the full use of the down BW advertised by peer
Tests show that with these settings the RTT stays low while better transfer rate is achieved

PR #2215, opened 2021-01-10
Added to show the message length on Composer and Message Widget

defnax:


csoler:
Does it disable the "send " button when the size is too large?

defnax:
> Does it disable the "send " button when the size is too large?

No, it needs a code for check if mail recipient is gxs or peer id. i dont know yet that part to code. for node mail doesnt need to check the message size

csoler:
Honestly I wouldn't mind for the type of receiver, given that mail to nodes is meant to disappear anyway.

PR #2218, opened 2021-01-11
Implement proper GXS message deletion notification

G10h4ck:
GxsForums propagate message deletion notification as RsEvent

### DIFF
--- a/libretroshare/src/retroshare/rsgxsforums.h
+++ b/libretroshare/src/retroshare/rsgxsforums.h
@@ -118,7 +118,7 @@ enum class RsForumEventCode: uint8_t
 	SYNC_PARAMETERS_UPDATED  = 0x0a, /// sync and storage times have changed
 	PINNED_POSTS_CHANGED     = 0x0b, /// some posts where pinned or un-pinned
 	DELETED_FORUM            = 0x0c, /// forum was deleted by cleaning
-	DELETED_POSTS            = 13    /// Posts deleted by cleaning
+	DELETED_POSTS            = 0x0d  /// Posts deleted by cleaning
 };
 
 struct RsGxsForumEvent: RsEvent
@@ -129,9 +129,9 @@ struct RsGxsForumEvent: RsEvent
 
 	RsForumEventCode mForumEventCode;
 	RsGxsGroupId mForumGroupId;
-	std::set<RsGxsMessageId> mForumMsgsId;
-    std::list<RsGxsId> mModeratorsAdded;
-    std::list<RsGxsId> mModeratorsRemoved;
+	RsGxsMessageId mForumMsgId;
+	std::list<RsGxsId> mModeratorsAdded;
+	std::list<RsGxsId> mModeratorsRemoved;
 
 	///* @see RsEvent @see RsSerializable
 	void serial_process(
@@ -141,7 +141,7 @@ struct RsGxsForumEvent: RsEvent
 		RsEvent::serial_process(j, ctx);
 		RS_SERIAL_PROCESS(mForumEventCode);
 		RS_SERIAL_PROCESS(mForumGroupId);
-		RS_SERIAL_PROCESS(mForumMsgsId);
+		RS_SERIAL_PROCESS(mForumMsgId);
 		RS_SERIAL_PROCESS(mModeratorsAdded);
 		RS_SERIAL_PROCESS(mModeratorsRemoved);
 	}
--- a/libretroshare/src/retroshare/rsgxsforums.h
+++ b/libretroshare/src/retroshare/rsgxsforums.h
@@ -115,9 +115,10 @@ enum class RsForumEventCode: uint8_t
 	READ_STATUS_CHANGED      = 0x06, /// msg was read or marked unread
 	STATISTICS_CHANGED       = 0x07, /// suppliers and how many messages they have changed
 	MODERATOR_LIST_CHANGED   = 0x08, /// forum moderation list has changed.
-    SYNC_PARAMETERS_UPDATED  = 0x0a, /// sync and storage times have changed
-    PINNED_POSTS_CHANGED     = 0x0b, /// some posts where pinned or un-pinned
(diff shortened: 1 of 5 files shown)


libretroshare/src/retroshare/rsgxsforums.h
```diff
@@ -128,7 +129,7 @@ struct RsGxsForumEvent: RsEvent
 
 	RsForumEventCode mForumEventCode;
 	RsGxsGroupId mForumGroupId;
-	RsGxsMessageId mForumMsgId;
+	std::set<RsGxsMessageId> mForumMsgsId;
```
csoler: I understand why you did that: msgChange has a list of msg ids. But in practice it's only one of them at a time. So all RsEvents for GXS services use the paradigm "one event per message". Therefore it looks a bit inconsistent with the rest of the code that forums would pack multiple messages in the same event whereas all the other services don't.

libretroshare/src/retroshare/rsgxsforums.h
```diff
@@ -115,9 +115,10 @@ enum class RsForumEventCode: uint8_t
 	READ_STATUS_CHANGED      = 0x06, /// msg was read or marked unread
 	STATISTICS_CHANGED       = 0x07, /// suppliers and how many messages they have changed
 	MODERATOR_LIST_CHANGED   = 0x08, /// forum moderation list has changed.
-    SYNC_PARAMETERS_UPDATED  = 0x0a, /// sync and storage times have changed
-    PINNED_POSTS_CHANGED     = 0x0b, /// some posts where pinned or un-pinned
-    DELETED_FORUM            = 0x0c, /// forum was deleted by cleaning
+	SYNC_PARAMETERS_UPDATED  = 0x0a, /// sync and storage times have changed
+	PINNED_POSTS_CHANGED     = 0x0b, /// some posts where pinned or un-pinned
+	DELETED_FORUM            = 0x0c, /// forum was deleted by cleaning
+	DELETED_POSTS            = 13    /// Posts deleted by cleaning
```
csoler: Plz use 0x0d.

libretroshare/src/retroshare/rsgxsforums.h
csoler: also because that, delete messages were send as a GrpChange, where it makes sense to have multiple messages at once, because it's one group change. If we use MsgChange now, it should be single.

csoler (changes requested):

libretroshare/src/retroshare/rsgxsforums.h
G10h4ck: > But in practice it's only one of them at a time.

Ok in doing that change, so I should also add a comment in the code, it is always just one message at time or is like that most of the time?

csoler: It's like that in all services up to now. We could have taken the other path (one RsEvents for multiple msgs) but it's not the case. I dont think there's a cost issue beyond this since as you say it's always one msg at once in practice.

libretroshare/src/services/p3gxsforums.cc
```diff
@@ -192,221 +192,202 @@ RsSerialiser* p3GxsForums::setupSerialiser()
...
-#endif
+	RS_DBG2(changes.size(), " changes to notify");
 
 	std::vector<RsGxsNotify *>::iterator it;
 	for(it = changes.begin(); it != changes.end(); ++it)
 	{
-		RsGxsMsgChange *msgChange = dynamic_cast<RsGxsMsgChange *>(*it);
-
-		if (msgChange)
+		rs_view_ptr<RsGxsNotify> gxsChange = *it;
```
csoler: I understand that you'd like to use this across fonction calls, but here it's just a local variable. It looks a bit over complicated for what it actually brings.

libretroshare/src/services/p3gxsforums.cc
```diff
@@ -192,221 +192,202 @@ RsSerialiser* p3GxsForums::setupSerialiser()
...
+		switch(gxsChange->getType())
 		{
-			if (msgChange->getType() == RsGxsNotify::TYPE_RECEIVED_NEW || msgChange->getType() == RsGxsNotify::TYPE_PUBLISHED) /* message received */
-				if (rsEvents)
+		case RsGxsNotify::TYPE_RECEIVED_NEW: // [[fallthrough]]
+		case RsGxsNotify::TYPE_PUBLISHED:
+		{
+			rs_view_ptr<RsGxsMsgChange> msgChange =
```
csoler: it's a bit inconsistent to dynamic_cast after entering the types of changes, given that MsgChange and GrpChange do not have the same subtypes. I understand it factors the code and maybe makes it a little smaller, but its harder to read now.

libretroshare/src/services/p3gxsforums.cc
G10h4ck: Ok, I can change to an unnannotated pointer if you prefer

libretroshare/src/services/p3gxsforums.cc
G10h4ck: `dynamic_cast` is very costly compared to an almost costless switch on an integer, so we should do that only in strictly necessary cases. also I think it would be much better to have more detailed RsGxsNotify enum type so each case can get only one object type, so we do just one cast per case in the eventuality it is strictly needed, so were I perceive the inconsistency is that the same `GxsNotify::TYPE_XXX` we can get different actual object types, or did I misunderstood the code?

csoler: The cost for sure would be a problem if you're in a loop. I'm not so sure it has any impact here. But surely, the different way of treating GxsNotify objects in forums (as opposed to all other GXS services) raises questions when reading the code.

G10h4ck: Actually that code is inside a for loop ;-p

G10h4ck:
Superseeded by https://github.com/RetroShare/RetroShare/pull/2301

PR #2227, opened 2021-01-15
[WIP] Added Dark theme for the graphframe

defnax:
Added Dark theme for the graphframe
Added Combobox for switch


retroshare-gui/src/gui/common/RSGraphWidget.cpp
```diff
@@ -345,7 +345,11 @@ void RSGraphWidget::paintEvent(QPaintEvent *)
   _painter->setRenderHint(QPainter::TextAntialiasing);
   
   /* Fill in the background */
-  _painter->fillRect(_rec, QBrush(BACK_COLOR));
+  if (_flags & RSGRAPH_FLAGS_DARK_STYLE){
```
PhenomRetroShare, line 348: Why not use qproperty so you can change it in qss file?

defnax: i doesnt used so much with qproperity, dont know yet how to format as qproperty, that feature

PR #2243, opened 2021-01-23
Added option to disable the ui for the unfinished services on plugins…

defnax:
… page

### DIFF
--- a/retroshare-gui/src/gui/MainWindow.cpp
+++ b/retroshare-gui/src/gui/MainWindow.cpp
@@ -425,18 +425,24 @@ void MainWindow::initStackedPage()
   addPage(peopleDialog = new PeopleDialog(ui->stackPages), grp, &notify);
   #endif
 #ifdef RS_USE_WIKI
+  if(Settings->getWikiService()){ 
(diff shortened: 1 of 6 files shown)


csoler:
This code is not correct IMHO: the plugins page is not supposed to "know" in advance what plugins to expect. So having a "enable wiki" button there really doesn't make sense. Same for other wiki-related things.
So, as far as I understand, wiki, photo, etc are not yet plugins. So something's weird here anyway.

defnax:
> This code is not correct IMHO: the plugins page is not supposed to "know" in advance what plugins to expect. So having a "enable wiki" button there really doesn't make sense. Same for other wiki-related things.
> So, as far as I understand, wiki, photo, etc are not yet plugins. So something's weird here anyway.

its a bad idea, i hope for a real plugin this year only a temporary solution thinked generate a PR for people which can hide the compiled services on the ui..
I wish as real gxs plugins for wire/wiki/photos
i doesnt like so much this thing what i did.

csoler:
ok, then keep this as a PR, and whoever wants to work on wiki+photo+etc can use this PR in his own fork.

PR #2258, opened 2021-01-27
font emoji

defnax:


retroshare-gui/src/gui/common/Emoticons.cpp
```diff
@@ -236,8 +236,12 @@ void Emoticons::showSmileyWidget(QWidget *parent, QWidget *button, const char *s
 
 			QString key = it.next();
 			QPushButton *button = new QPushButton("", tabGrpWidget);
+			QFont font = button->font();
+			font.setPointSize(22);
```
csoler, line 240: That's an absolute size. Doesn't seem right at all.

defnax: this is only a experimental hack for myself, for the font emoji

PR #2263, opened 2021-01-29
Attempt to improve BW graph

csoler:
Warning: the existing saved settings may not apply correctly the first time that window is used. You need to change something in it (to trigger a re-save) then restart RS.


defnax:
sorry had no time for this, maybe tonight, can you add a option to show after speed value store/load completed downloaded/uploaded in MB/GB ?
image

PR #2273, opened 2021-02-01
People page - different icon for people with downvotes

RetroPooh:


csoler:
People already show a yellow "bio hazard" icon for people with down votes. What is this PR changing? Also, when I tried it, all people bullet icons where grey for me, which is very unexpected.

RetroPooh:
It makes difference when you change autoban from default -1 (what is too strict) vote to -2 or more, so you can see when some id was voted down but still okay and so can support him if you want. That kind in red, like is about to go.
When id is positive but still has some votes down it is yellow, this is just FYI.

PR #2274, opened 2021-02-01
display available message counters for unsubscribed groups

RetroPooh:


csoler:
This is not consistent to use the same place to display two different types of information: in subscribed groups, numbers tell the count of unread messages (which is irrelevant in unsubscribed groups). In your PR you add that unsubscribed groups show the messages available at friends, but it looks like the messages are here and new (as it is the meaning of that number).

RetroPooh:
In compact view (aka default, with hidden f.count column) it's useful to quickly see what is alive and what is not.

PR #2277, opened 2021-02-01
allow sharing more chatlobbys

RetroPooh:
(cherry picked from commit 75de450e31dcdb63051d7f457ce84974d3ba890d)

# Conflicts:
#	libretroshare/src/chat/distributedchat.cc


csoler:
50 chat rooms is already quite a lot. Why increase this even more?

RetroPooh:
People subscribe a lot, it may be their problem, but chats more than limit are silently dropped and not processed. It's no good.

csoler:
then the proper response should be to warn them, not just increase the limits. In the end there always will be limits.

PR #2280, opened 2021-02-02
chat options - unhide don't send typing notif

RetroPooh:


csoler:
I vote against this. I think the typing notification is important and it cannot harm to see that somebody's writing. On the contrary, hiding it allows to spam chat rooms a bit more easily since you only get noticed when the message is already sent.
In the first place, I dont even understand why this option to hide the typing info would be implemented.

RetroPooh:
https://github.com/RetroShare/RetroShare/pull/653
And when people want to spam, they won't be stopped with some gui feature.

csoler:
Closing because this is an old PR that implements a feature that is not wanted.

PR #2281, opened 2021-02-02
Add RsTreeWidget Sort Context Menu

PhenomRetroShare:
Now Header sort when visible or In context menu is hidden.

### DIFF
--- a/retroshare-gui/src/gui/common/GroupTreeWidget.cpp
+++ b/retroshare-gui/src/gui/common/GroupTreeWidget.cpp
@@ -31,23 +31,21 @@
 #include "util/QtVersion.h"
 #include "util/DateTime.h"
 
+#include <QDesktopWidget>
 #include <QMenu>
+#include <QToolButton>
 
 #include <stdint.h>
+
+Q_GUI_EXPORT int qt_defaultDpi();
 
 #define ROLE_ID              Qt::UserRole
 #define ROLE_NAME            Qt::UserRole + 1
 #define ROLE_DESCRIPTION     Qt::UserRole + 2
 #define ROLE_POPULARITY      Qt::UserRole + 3
-#define ROLE_LASTPOST        Qt::UserRole + 4
-#define ROLE_POSTS           Qt::UserRole + 5
-#define ROLE_UNREAD          Qt::UserRole + 6
-#define ROLE_SEARCH_SCORE    Qt::UserRole + 7
-#define ROLE_SUBSCRIBE_FLAGS Qt::UserRole + 8
-#define ROLE_COLOR           Qt::UserRole + 9
-#define ROLE_SAVED_ICON      Qt::UserRole + 10
-#define ROLE_SEARCH_STRING   Qt::UserRole + 11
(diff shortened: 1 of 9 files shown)


retroshare-gui/src/gui/common/GroupTreeWidget.cpp
```diff
@@ -167,8 +160,8 @@ void GroupTreeWidget::addToolButton(QToolButton *toolButton)
 
 	/* Initialize button */
 	toolButton->setAutoRaise(true);
-	toolButton->setIconSize(ui->displayButton->iconSize());
-	toolButton->setFocusPolicy(ui->displayButton->focusPolicy());
+	toolButton->setIconSize(QSize(24,24));
```
csoler: Bad way to do it: you need 24*QFontMetricsF(font()).height()/14.0, so it remains proportional to font size and keeps the same look on high DPI screens.

retroshare-gui/src/gui/common/RSTreeWidget.cpp
```diff
@@ -283,13 +283,57 @@ QMenu *RSTreeWidget::createStandardContextMenu(QMenu *contextMenu)
 		actShowHeader->setCheckable(true);
 		actShowHeader->setChecked(!isHeaderHidden());
 
-		QMenu *headerMenu = contextMenu->addMenu(QIcon(),tr("Show column..."));
-
 		QTreeWidgetItem *item = headerItem();
 		int columnCount = item->columnCount();
+
```
csoler, line 290: Is this generic enough to work with all different instances of RSTreeWidget? It's used quite a lot.

retroshare-gui/src/gui/common/GroupTreeWidget.cpp
PhenomRetroShare: OK,
For next, could you test with this:
The Qt::AA_EnableHighDpiScaling application attribute, introduced in Qt 5.6, enables automatic scaling based on the monitor's pixel density.
https://doc.qt.io/qt-5/highdpi.html

retroshare-gui/src/gui/common/RSTreeWidget.cpp
PhenomRetroShare: Yes

retroshare-gui/src/gui/common/GroupTreeWidget.cpp
PhenomRetroShare: Now, I use:
	int i = qt_defaultDpi();
	auto desktopWidget = QApplication::desktop();
	auto y = desktopWidget->logicalDpiY();

	toolButton->setIconSize(QSize(24*y/i,24*y/i));
If it's OK, we can make it static and in utils function.

jolavillette:
thx for this PR
to fix:
- when sorting by "last post" a forum with no post should **not** appear as the most recently updated forum
- when sorting by "number of posts" or "popularity" please use numeric sort **not** alphanumeric

PhenomRetroShare:
@jolavillette : Done...

PR #2285, opened 2021-02-03
Fix bug in pqihandler preventing UP BW management when bwctrl service is OFF

jolavillette:
When bandwidth ctrl service was OFF, there was no call to setMaxRate(false, out_max_bw), and the max UP BW was never updated. 
Even worse, if RS was started with bandwidth ctrl OFF, the max UP BW remained at the start value of 0, making impossible almost any network activity.

### DIFF
--- a/libretroshare/src/pqi/pqihandler.cc
+++ b/libretroshare/src/pqi/pqihandler.cc
@@ -470,18 +470,11 @@ int     pqihandler::UpdateRates()
 
 		// for our up bandwidth we take into account the max down provided by peers via BwCtrl
 		// because we don't want to clog our outqueues, the TCP buffers, and the peers inbound queues
+		mod -> pqi -> setMaxRate(false, out_max_bw);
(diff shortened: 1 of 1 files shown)


libretroshare/src/pqi/pqihandler.cc
```diff
@@ -470,18 +470,11 @@ int     pqihandler::UpdateRates()
 
 		// for our up bandwidth we take into account the max down provided by peers via BwCtrl
 		// because we don't want to clog our outqueues, the TCP buffers, and the peers inbound queues
+		mod -> pqi -> setMaxRate(false, out_max_bw);
```
csoler, line 473: plz put a comment (basically what you said in the PR text) here. It's not really easy to figure out why you added this otherwise. thx!

jolavillette: OK but we need to talk about that first.
IMO the comments already explains what we do: we limit our UP to the DN value passed by our friend via the BwCtrl service. 
And there is of course no need to mention a bug that is not there anymore.

PR #2289, opened 2021-02-05
Optimization of TR and SR lifespan

jolavillette:
The purpose of this very simple PR is to improve the management of TR and SR

Tests show that TR and RS bounce a lot in RS network. After being cleared from the cache when their lifespan is reached (at the moment 240 seconds) they come again from somewhere else. We consider them as new and we propagate them again. This can last for hours.

Furthermore sometimes very old TR finally give a positive answer after having bounced everywhere, and a tunnel is opened. Such zombie tunnels don't carry any data and are killed after 1 minute inactivity. In the mean time they will have filled our statistics with garbage and created unneeded traffic between peers.

So in this PR:
- First we increase the lifespan of TR and SR in the turtle cache from 240 to 600 seconds. Keeping known TR & SR in cache during a longer period will help avoid duplicates caused by requests bouncing all over the network. Of course the caches will be initially a little bit bigger, but it does not hurt performance at all.
- Second we drastically reduce the time during which we take into account or forward to our friends the answers we receive to known TR & SR. The previous value of 240 seconds is now reduced to 10 seconds only. Any tunnel result or search result received more than 10 seconds after we put it in cache will be ignored and discarded.

### DIFF
--- a/libretroshare/src/turtle/p3turtle.cc
+++ b/libretroshare/src/turtle/p3turtle.cc
@@ -77,8 +77,10 @@ void TS_dumpState() ;
 //    - The total number of TR per second emmited from self will be MAX_TUNNEL_REQS_PER_SECOND / TIME_BETWEEN_TUNNEL_MANAGEMENT_CALLS = 0.5
 //    - I updated forward probabilities to higher values, and min them to 1/nb_connected_friends to prevent blocking tunnels.
 //
-static const rstime_t TUNNEL_REQUESTS_LIFE_TIME                = 240 ; /// life time for tunnel requests in the cache.
-static const rstime_t SEARCH_REQUESTS_LIFE_TIME                = 240 ; /// life time for search requests in the cache
+static const rstime_t TUNNEL_REQUESTS_LIFE_TIME                = 600 ; /// life time for tunnel requests in the cache.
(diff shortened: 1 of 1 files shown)


RetroPooh:
Isn't 10secs too little for forwarding SRs? With 5000 results limit x how many bytes each... how will this fit thru slow bw links like torified nodes?

csoler:
I'm definitely on RetroPooh side here: 10 secs is really too low. Should be 60. Sometimes TRs gets slowed down for a while and it's too back losing a tunnel opening just because of a timeout. If you increase cache to 600 then 60 will not allow the same bouncing artifacts that we had before. In any way, that needs some testing by different nodes.

jolavillette:
I don't believe that a TR that needed 60 seconds to come back to us (and still needs to reach the node it originates from!!!) will do anything good in a network where RTT is now usually better than 200ms. 
The truth is that these late TR hurt the network a lot. They are responsible for zombie tunnels. But not only. Each useless TR that we forward instead of dropping it, takes the place of a brand new TR that will be discarded because of the probability system. 
So we save the bad ones and we kill the good ones. We are about to miss a unique chance to do things right for file transfer and distant chat in 0.6.6.

PR #2291, opened 2021-02-05
statistics - remove gxs tunnels from ft; add filenames for known hashes

RetroPooh:


libretroshare/src/gxs/rsgxsnettunnel.cc
```diff
@@ -1155,22 +1155,7 @@ void RsGxsNetTunnelService::getStatistics(std::map<RsGxsGroupId,RsGxsNetTunnelGr
...
-
-
-
-
-
-
-
-
-
+bool RsGxsNetTunnelService::isGXSHash(RsFileHash hash) const
```
csoler, line 1158: misleading because this is only valid for local hashes.

csoler:
Which GXS tunnels are you removing from which FT? Can you be more precise plz? thx.

RetroPooh:
Removing tunnels providing distant gxs sync (currently channels) from listing on statistics - turtle - file transfer tunnels.
For that searching for tunnel hash (aka RsFileHash) among mHandledHashes of RsGxsNetTunnelService

libretroshare/src/gxs/rsgxsnettunnel.cc
RetroPooh: I don't understand. This is wrong why?

RetroPooh:
consider this one https://github.com/RetroShare/RetroShare/pull/2294 instead
this one locks up deep in ftserver when looking for filenames

PR #2302, opened 2021-02-08
Fix IdDialog Update

PhenomRetroShare:
No update while received new Id, wait last (5s) to update all.
Fix expanded items in same time.

### DIFF
--- a/retroshare-gui/src/gui/Identity/IdDialog.cpp
+++ b/retroshare-gui/src/gui/Identity/IdDialog.cpp
@@ -128,8 +128,8 @@ class TreeWidgetItem : public QTreeWidgetItem
 {
   public:
   TreeWidgetItem(int type=Type): QTreeWidgetItem(type) {}
-  TreeWidgetItem(QTreeWidget *tree): QTreeWidgetItem(tree) {}
-  TreeWidgetItem(const QStringList& strings): QTreeWidgetItem (strings) {}
+  explicit TreeWidgetItem(QTreeWidget *tree): QTreeWidgetItem(tree) {}
+  explicit TreeWidgetItem(const QStringList& strings): QTreeWidgetItem (strings) {}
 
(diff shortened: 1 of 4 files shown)


retroshare-gui/src/gui/Identity/IdDialog.cpp
```diff
@@ -1560,23 +1600,16 @@ void IdDialog::loadIdentities(const std::map<RsGxsGroupId,RsGxsIdGroup>& ids_set
...
+	ui->idTreeWidget->insertTopLevelItem(0, contactsItem );
+
 	/* Insert new items */
 	for (std::map<RsGxsGroupId,RsGxsIdGroup>::const_iterator vit = ids_set.begin(); vit != ids_set.end(); ++vit)
 	{
 		RsGxsIdGroup data = vit->second ;
 
-	    item = NULL;
-
-	    ui->idTreeWidget->insertTopLevelItem(0, ownItem);
```
csoler, line 1570: what the hell what this doing here???

PhenomRetroShare: I don't know...
3b45fd8389 (csoler              2016-04-09 17:50:46 -0400)
You want I move them on constructor?

csoler: ah yes, good idea! So we know they are already here and we dont need to add them multiple times. Make sure there's no "ui->idTreeWidget->clear()' then. If there is, maybe it's preferable to leave it as it is in your current PR.

PhenomRetroShare: Done

csoler:
For the record, generally we should avoid these timer-based updates, in favor of buffering the notifications in the service and use service_tick() to trigger them at regular time intervals when necessary. Let's keep this for 0.6.6. We'll have some time to make it better later on.

PR #2329, opened 2021-02-16
fix hashstatus appearing on extrafile unshare

RetroPooh:


csoler:
seems to me that a much simpler way to fix this would be to just add a break right after the default in the case you changed in hashstatus.cpp

PR #2334, opened 2021-02-18
Fix Chat Style Setting

PhenomRetroShare:


defnax:
Now Chat Bubble style will work again?
Need to test tonight

PhenomRetroShare:
No, But now you can select it. ;)

defnax:
Only select will not help 😅 i hoped you will fix that

PR #2344, opened 2021-02-20
Fix Bubble Chat Style

PhenomRetroShare:


defnax:
tested my self last few days

PR #2347, opened 2021-02-20
Added Edit Identity button to People view

defnax:


csoler:
Isn't there a edit identity in the context menu already?

defnax:
Random reported on Forums that is not easy to go every time on context menu to edit Identity for Avatar.
He requested this, or he wanted a Avatar edit Button on People View

csoler:
you dont edit an identity very often right? in this case, it's probably saving a lot of space to keep the edit button in the drop menu. If you add a button permanently for something that is really not common, isn't it a waste of space?

defnax:
Its visible when invite button is hidden, i get idea only from Random.
Im not fan like you do all things from context menu.
It looks like before one button only visible, instead of a disabled invite button it shows a Edit one

csoler:
ok. If it doesn't add yet another button, I suppose it's good to have.

PhenomRetroShare:
Why not by just clicking on avatar?
This could be bordered if it is editable.

csoler:
> Why not by just clicking on avatar?
> This could be bordered if it is editable.

I fully agree. The less buttons everywhere, the better.

defnax:
> Why not by just clicking on avatar?
> This could be bordered if it is editable.

you can do a pr for Avatars Edit via one click, that needs recode to the avatarwidget more complex ;).
But again the AvatarDialog needs to fixed for GXS identities too.

It need then more work, maybe if you can do it via new PR :)

PR #2349, opened 2021-02-20
Fixed to hide the new thread button when not subscribed

defnax:


csoler:
why hide it? It's already disabled.

defnax:
> why hide it? It's already disabled.

to get more space for the forum title, when browsing not subscribed forums
else it uses uneeded space when its only disabled
now the search field stretches too, it takes only space

csoler:
On one hand, it's good to see the button, even if it's disabled, because it tells you that the functionality might exist and you need something to enable it (subscribe).
On the other hand, I agree with the space problem. Maybe you may just remove the text 'new thread', and only keep the button, with appropriate tooltip: when enabled: "start a new thread", and when it's disabled: "subscribe to the forum if you want to post to it".
Also, it's weird that the "start new thread" button exists, and not the "reply to message". Both of them are in the context menu. It' a bit inconsistent to have only one of them in the bar. I would prefer either none of them, or both.

PR #2352, opened 2021-02-22
keyring - fix removing accepted keys; allow multi-select

RetroPooh:


csoler:
Removing accepted keys is a very weird idea. It will prevent RS from connecting to these friends.

RetroPooh:
But RS now allows doing that, and it will crash after you remove a key from keyring and then try to see his details in friends list.
With this PR it asks first, then allows to do it safely.

You may name this as a feature for fast key cleaning, with it it's easy to arrange by last usage, select them all who wasn't online for whatever you like, and goodbye.

csoler:
Then the way to fix that is to disallow the removal of friend keys. It wouldn't make sense to remove them anyway.

PR #2372, opened 2021-03-10
Disable animation in file transfer tree view widgets

altcoinpirate:
Speed up folder expansion by disabling animation.


G10h4ck:
this will probably help people with big file collections, thanks for the contribution

PR #2374, opened 2021-03-11
restore back changes

defnax:


retroshare-gui/src/util/HandleRichText.cpp
```diff
@@ -621,10 +621,10 @@ QString RsHtml::formatText(QTextDocument *textDocument, const QString &text, ulo
 	QDomDocument doc;
 	if (doc.setContent(formattedText, &errorMsg, &errorLine, &errorColumn) == false) {
 
-		// convert text with QTextDocument
-		QTextDocument textDoc;
-		textDoc.setPlainText(text);
-		formattedText=textDoc.toHtml();
+		// convert text with QTextBrowser
+		QTextBrowser textBrowser;
```
PhenomRetroShare, line 625: QTextBrowser needs to be in GUI thread... That why this change.
I've posted in lobby a fix but doesn't get reports.

--- retroshare-gui/src/util/HandleRichText.cpp
+++ retroshare-gui/src/util/HandleRichText.cpp
@@ -625,3 +625,7 @@ QString RsHtml::formatText(QTextDocument *textDocument, const QString &text, ulo
     QTextDocument textDoc;
+ if (Qt::mightBeRichText(formattedText))
+ textDoc.setHtml(formattedText);
+ else
       textDoc.setPlainText(text);
+
     formattedText=textDoc.toHtml();

PhenomRetroShare (changes requested):

retroshare-gui/src/util/HandleRichText.cpp
defnax: Hi, i hear nothing from you then created this PR,
I will test then tonight your fix

jolavillette: It would be a good idea to make a post in RS dev forum when one creates a PR.
Good idea also to provide a description of what the PR does.

defnax: @PhenomRetroShare can you make a new PR for test or sent me the right patch for it, i doesnt saw your chat lobby post

PR #2377, opened 2021-03-13
Implement pull request mechanism in RsGxsNetService

G10h4ck:
This could be used to request the online peers to pull updates from us
  ASAP, as an example when a group is created a pull request can be
  emitted too so the online peers pull the groups from us ASAP instead
  of waiting for the usual 60 seconds. A mechanism like this is
  especially useful on mobile phones where the internet connection is
  usually turned on only in a few moments (as an example while the user
  is interacting with the app).
Cleanup a few old corners in the code keeping retro-compatibility and
  make the code more welcoming to new developers.
Put a bunch of dead code under #ifdef.

### DIFF
--- a/libretroshare/src/gxs/rsgxsnetservice.cc
+++ b/libretroshare/src/gxs/rsgxsnetservice.cc
@@ -35,7 +35,7 @@
 //    |
 //    +----------- sharePublishKeys()
 //    |
-//    +----------- syncWithPeers()
+//    +----------- pullFromPeers()
 //    |              |
 //    |              +--if AutoSync--- send global UpdateTS of each peer to itself => the peer knows the last
 //    |              |                 time current peer has received an updated from himself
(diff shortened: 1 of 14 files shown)


libretroshare/src/rsitems/rsnxsitems.cc
```diff
@@ -64,6 +64,12 @@ RsItem *RsNxsSerialiser::create_item(uint16_t service_id,uint8_t item_subtype) c
     if(service_id != SERVICE_TYPE)
         return NULL ;
 
+	switch(static_cast<RsNxsSubtype>(item_subtype))
```
csoler, line 67: There's a switch below on that same variable. Why make a separate switch here?

libretroshare/src/rsitems/rsnxsitems.h
```diff
@@ -33,8 +35,13 @@
 #include "serialiser/rstlvkeys.h"
 #include "gxs/rsgxsdata.h"
 
-// These items have "flag type" numbers, but this is not used.
+enum class RsNxsSubtype : uint8_t
```
csoler, line 38: Plz add one entry the types below. Otherwise it makes the code more complex. I know you prefer to have typed enums, but here you just need to add one to an existing structure. Changing all of them would cause a massive change as well that is really not needed.

libretroshare/src/serialiser/rsserial.cc
```diff
@@ -166,11 +167,17 @@ uint8_t    RsItem::PacketSubType() const
...
+	// This value triggers PQIInterface to complain about undefined priorities
+	_priority_level = QOS_PRIORITY_UNKNOWN;
 	type = (ver << 24) + (service << 8) + subtype;
-	return;
 }
 
+RsItem::RsItem( uint8_t ver, RsServiceType service, uint8_t subtype,
+        RsItemPriority prio ):
+    type(static_cast<uint32_t>(
+             (ver << 24) + (std::to_underlying(service) << 8) + subtype )),
```
csoler, line 178: WTH? There was a very simple and clean line before. Why all this complexity?

libretroshare/src/gxs/rsgxsnetservice.cc
```diff
@@ -33,7 +35,7 @@
 //    |
 //    +----------- sharePublishKeys()
 //    |
-//    +----------- syncWithPeers()
+//    +----------- pullFromPeers()
```
csoler: I dont think this renaming is appropriate: syncWithPeers() sends to all peers the last change time you know for their own data. So you're not pulling anything. If they see that this time is too old (as compared to their own data) they send you the list of new stuff. So it's not pulling either.

libretroshare/src/gxs/rsgxsnetservice.cc
```diff
@@ -744,7 +745,7 @@ void RsGxsNetService::syncWithPeers()
 #endif
 }
 
-void RsGxsNetService::generic_sendItem(RsNxsItem *si)
+void RsGxsNetService::generic_sendItem(rs_owner_ptr<RsItem> si)
```
csoler, line 751: Only you know what that template means. A brief comment saying that the method takes ownership of the item would be far more helpful.

csoler (changes requested):
It seems to do what's needed.

PR #2390, opened 2021-03-21
Fixing branch icon

defnax:
image


csoler:
I don't get it. These files were here already not so long ago. Did you remove them when you switched to big arrows?

defnax:
no i added some time ago new arrows, issues with background stylesheet, when background stylesheet is used then it requires new icons, background stylesheet overwrites the tree icon display.

my attempt was to fix a issue, better color background needs a new tree icon...
another issue is when friend selected the color for text needs to be black or white.

PR #2393, opened 2021-03-25
Fix: WebUI truncating binary files

hunbernd:
Fixes font files not loading properly in the webui problem.


jolavillette:
RsDbg() does not need std::endl. It does it by itself. 
Putting a std::endl in a RsDbg() call will add an unnecessary blank line in the log.

PR #2399, opened 2021-03-28
fixed use of no avatar instead of default avatar image when creating a new ID

csoler:
Dont merge this in master, but in v0.6.6-OfficialRelease. I leave it in master for now for easier testing.


csoler:
merged by cherry-picking, because when rebased on v0.6.6-OfficialRelease branch, it has lots of unexpected extra commits.

PR #2411, opened 2021-04-19
Fix plugins compilation issue

ghost:
Plugins compilation issue on Linux. VOIP plugin requires different QT settings depending on the distribution.

Comment out plugins/VOIP/VOIP.pro:56 until most major distributions have opencv-4 available. Right now it's opencv4 for Ubuntu and opencv3 for everything else.

This does not break anything, just a friendly automated check for installed packages.

### DIFF
--- a/plugins/VOIP/VOIP.pro
+++ b/plugins/VOIP/VOIP.pro
@@ -53,7 +53,11 @@ linux-* {
 
 	PKGCONFIG += libavcodec libavutil
 	PKGCONFIG += speex speexdsp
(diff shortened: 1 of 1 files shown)


csoler:
I'm not sure you committed every changes you wanted to make. Given what github shows now, your fix breaks compilation on all systems since it simply removes opencv.

ghost:
Basically, yes, thanks for not letting this one through. I lost track of different Linux distros a little bit there.

PR #2414, opened 2021-04-23
[WIP] Implement perceptual indexing and search

G10h4ck:
This pull request is a work in progress to implement perceptual file indexing and search capabilities in libretroshare. Once complete it should be possible to run searches over the RetroShare network to find similar files (at least images, audio and videos) like big G does for images. As an example with perceptual search it will be easy to find a version with better quality of an audio or an image having a worse one, as the two versions should produce very "similar" perceptual hashes.


felisucoibi:
is nice but is something nobody asked for and lot of things that still needs to be done like a working web RS

G10h4ck:
> is nice but is something nobody asked for and lot of things that still needs to be done like a working web RS

None that you know, but someone that you don't may even have donated support for this cause

felisucoibi:
good to know this is the way

PR #2416, opened 2021-04-25
Fix libupnp autodetection for qmake

ghost:
Fix libupnp autodetection in qmake on Ubuntu and OpenSuse.

Qmake script (retroshare.pri) is unable to find installed libupnp files. This results in a warning message 

"[Inexact] Project WARNING: RS_UPNP_LIB detection failed, UPnP support disabled!"

and UPNP-related code being disabled.

Lookup directory is hardcoded as "/usr/lib" in retroshare.pri, only few distributions actually install libupnp files there.

This PR fixes the issue for Ubuntu and OpenSuse.

### DIFF
--- a/retroshare.pri
+++ b/retroshare.pri
@@ -609,6 +609,8 @@ linux-* {
     isEmpty(RS_PLUGIN_DIR) : RS_PLUGIN_DIR  = "$${PREFIX}/lib/retroshare/extensions6"
 
     QMAKE_LIBDIR *= "$$RS_LIB_DIR"
+    QMAKE_LIBDIR += "$${PREFIX}/lib64"
+    QMAKE_LIBDIR += "$${PREFIX}/lib/x86_64-linux-gnu"
 
     rs_autologin {
         # try libsecret first since it is not limited to gnome keyring and libgnome-keyring is deprecated


G10h4ck:
Thanks for this proposal, can you point out at which line the hard-coding happens?

ghost:
In the qmake build script retroshare.pri:604

```
linux-* {
    isEmpty(PREFIX)        : PREFIX         = "/usr"
    ...
    isEmpty(RS_LIB_DIR)    : RS_LIB_DIR     = "$${PREFIX}/lib"
    ...

    QMAKE_LIBDIR *= "$$RS_LIB_DIR"
    ...
```

As a result, later on @ retroshare.pri:772  it looks for installed libupnp files at /usr/lib/ and fails to find any.

G10h4ck:
So we should probably fix it at that point, because the problem is conceptually also with other libraries too, so should we do something like use ´$${PREFIX}/lib64´ if it exists and ´$${PREFIX}/lib´ if it doesn't as default setting for  ´$$RS_LIB_DIR´, also i am almost sure that changing it the way you originally thought break at least Android build

Cheers and thanks for your contribution!

ghost:
Notice that QMAKE_LIBDIR is not a string, it's a list. If you add more paths in it, the script will search for libupnp files in that list until it finds what it looks for.

Somebody has implemented the thing you are talking about already, just for some reason the list has only 1 item, "/usr/lib". You can just add more lookup paths like I did in this PR.

Can not comment on anything Android-related, although this PR only changes things inside the 

```
linux-* {
    {...}
}
```

section @ retroshare.pri:603. The android-* section is separate and right below it.

G10h4ck:
You are right, can you reopen the PR so I can merge it please?

Thanks!

PR #2421, opened 2021-05-12
Switched video capture to QCamera and removed opencv dependency

csoler:
This can already be tested. I'll do the same for speex.


defnax:
what about when we want use voip plugin on webui ? it will require qt then
i dont know if voip will be ever recoded to get used as plugin in the webui as example.

i know webui doesnt suports yet plugins.
plugin -> js api -> webui

defnax:
by the way you can add the missed options for audio in and output selection into AudioWizard.cpp/ui?
on mumble this was present but on voip plugin is not implemented since years

- Audio Input Device Selection (Microphone)
- Audio Output Device Selection (Speaker/Headphones)

image

 i had the modifyd src/ ui for this page but seems i lost it

https://github.com/mumble-voip/mumble/archive/refs/tags/1.3.4.tar.gz
https://github.com/mumble-voip/mumble/tree/1.3.4/src/mumble

defnax:
i get crash and the camera dont show anything on windows 10, but from master voip with opencv has no issue.

- nothing showin in camera output settings
-  same in chat window always empty

csoler:
make sure you have the qtmultimedia-plugins installed. Otherwise Qt won't see the camera. I dont know to do that on windows.

felisucoibi:
+1

PR #2437, opened 2021-07-01
V0.6 tor control

csoler:
This pull request currently allows to run retroshare-service and retroshare-gui with automated Tor configuration system. It can boot an existing node and most probably can create one too (that I haven't tested myself).

Since TorControl was originally using Qt for slots, mot of the work has been to remove signals/slots and replace them by something else that is Qt-free (for instance rsEvents, callbacks, and an extra loop in the TorManager thread).

The PR is ready for testing.


felisucoibi:
+1

Kumaravinash9:
Thanks @csoler for adding Tor functionality in Retrorshare services

csoler:
Anymore comments from testers?

PR #2442, opened 2021-07-14
Emit an event when a shared file hashing complete

G10h4ck:
Properly notify when a shared file has been hashed
Deprecate event with arbitrary data packed in std::string

### DIFF
--- a/libretroshare/src/retroshare/rsfiles.h
+++ b/libretroshare/src/retroshare/rsfiles.h
@@ -194,7 +194,8 @@ enum class RsFileTransferEventCode: uint8_t {
     COMPLETED_FILES_REMOVED     = 0x02,	//
 };
 
-struct RsSharedDirectoriesEvent: RsEvent
+struct RS_DEPRECATED_FOR("Packing arbitrary data into an std::string is bad idea")
+RsSharedDirectoriesEvent: RsEvent
 {
 	RsSharedDirectoriesEvent()  : RsEvent(RsEventType::SHARED_DIRECTORIES), mEventCode(RsSharedDirectoriesEventCode::UNKNOWN) {}
 	~RsSharedDirectoriesEvent() override = default;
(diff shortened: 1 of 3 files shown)


libretroshare/src/retroshare/rsfiles.h
```diff
@@ -212,6 +213,31 @@ struct RsSharedDirectoriesEvent: RsEvent
     std::string mMessage;
 };
 
+struct RsFileHashingCompletedEvent: RsEvent
```
csoler, line 216: I'm a bit surprised by this particular choice. Shouldn't we instead use a RsFileHashingEvent, and have different event codes such as DIRECTORY_SWEEP_STARTED (means it's starting to check all shared directories), HASHING_FILE (means it's hashing a specific file), DIRECTORY_SWEEP_ENDED (means hashing is done and dir watching has completed) ? It's weird to create an event class for one specific type of hashing step.

csoler: I think that originally, I added this event with only one type as an example. It wasn't really used, but rsNotify was.
There's actually a lot of possible events in shared directories: when you sync with friends (requires some specific event type I suppose, with friend ID and type: up vs. down), directory sweep (meaning RS looks over shared dires for possible changes and adds what's to be hashed into a list), and file hashing (means it's hashing a particular file).

csoler: Also it seems to me that you do not catch the new rsEvent in the Qt UI, so it probably breaks backward compatibility.

G10h4ck: No problem in creating many struct for RsEvents we have room for 2^32 of them ;) this way a bit more work is done at compile time and less is done at runtime, while if we do events we subtypes those get switched at runtime, also non-meaningful fields for a specific event get stored in RAM/serialized/deserialized etc even if not useful. RetroShare-gui doesn't handle this particular event, but it won't break anything as the deprecated one (which is used by RetroShare-gui) keeps being emitted

PR #2450, opened 2021-08-15
V0.6 pgp handler

csoler:
Turns PGPHandler into an abstract class that is derived as OpenPGPSDKHandler to provide a specific implementation for openpgpsdk. This should make it easier to switch to any other PGP package. Also got rid of getAuthGPG(), using static members for AuthGPG instead.


libretroshare/src/gossipdiscovery/p3gossipdiscovery.cc
```diff
@@ -219,7 +219,7 @@ void p3discovery2::removeFriend(const RsPeerId &sslId)
 		std::cerr << std::endl;
 #endif
 		/* pgp peer without any ssl entries -> check if they are still a real friend */
-		if (!(AuthGPG::getAuthGPG()->isGPGAccepted(pgpId)))
+        if (!(AuthGPG::isGPGAccepted(pgpId)))
```
G10h4ck: Now that you are refactoring this please rename `isGPGAccepted` into `isPgpAccepted` note the letters changed not just the casing ;)

libretroshare/src/pgp/pgphandler.cc
```diff
@@ -33,1516 +33,82 @@
...
-#else
-	ops_memory_t *memres = ops_sign_buf(data,len,OPS_SIG_BINARY,OPS_HASH_SHA1,secret_key,ops_false,ops_false,not_raw,not_raw) ;
-#endif
+	for(std::map<RsPgpId,PGPCertificateInfo>::const_iterator it(_public_keyring_map.begin()); it != _public_keyring_map.end(); ++it)
+	{
+		std::cerr << "PGP Key: " << it->first.toStdString() << std::endl;
 
-	if(!memres)
-		return false ;
+		std::cerr << "\tName          : " <<  it->second._name << std::endl;
```
G10h4ck: Isn't better to use `RS_DBG() <<` and friends here so it works also on android without black magic?

G10h4ck (changes requested):

G10h4ck:
Beside the inline comments, shouldn't we use a global pointer + methods like we do for most services like `rsFiles->someMethod()` this way we can easily expose the functionality also to JSON API, moreover i think it is easier to integrate other PGP implementations that way instead of static methods, as another implementation may need to keep some runtime state which is more difficult to do within static stuff.

csoler:
I actually hesitated between the two options. This one has some advantages as well, but otherwise the pointer method would be more consistent with the rest of the code.

G10h4ck:
I agree on keeping it homogeneous with the rest of the code, so please switch to global pointer + methods style now that your are still in time ;)
When I started coding into RS having so different patterns and anti-patterns into different parts of code added much to the entry barrier.

libretroshare/src/gossipdiscovery/p3gossipdiscovery.cc
csoler: done

csoler:
Well actually looking at the code AuthPGP is not a structure that we may expose at the API level. So there's no problem making it static.

libretroshare/src/pgp/pgphandler.cc
G10h4ck: ping

csoler:
ok I fixed both conflicts and std::cerr problems.

PR #2459, opened 2021-08-22
Fix: additional / when combining file paths

hunbernd:
Fix downloaded files stuck in partials folder, when RS not in administrator mode

### DIFF
--- a/libretroshare/src/ft/ftcontroller.cc
+++ b/libretroshare/src/ft/ftcontroller.cc
@@ -735,8 +735,7 @@ bool ftController::completeFile(const RsFileHash& hash)
 		RsDirUtil::splitDirFromFile(fc->mDestination,dst_dir,dst_file) ;
 
 		// We use this intermediate file in case the destination directory is not available, so as to not keep the partial file name.
-
-		std::string intermediate_file_name = src_dir+'/'+dst_file ;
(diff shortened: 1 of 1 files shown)


afAtedg:
Hello,
I just install Retroshare on two friend's computers, and the problem seem to still be there.
**The downloaded file are stuck in the temporary folder.**

I haven't tried to launch it in administrator mode but I am not near the computer now, 
so I will ask my friend to try that when possible.

One computer is running on Win 11 and the other on Win 10.

And the Retroshare version is the latest one downloaded on https://retroshare.cc/downloads.html :
https://github.com/RetroShare/RetroShare/releases/download/v0.6.6/RetroShare-0.6.6-20210313-0-g751fffc30-Qt-5.15.2-x64-setup.exe

Psudonico:
Hello,
I have the same problem.
My computer is running on Win 11

hunbernd:
@Psudonico Do you use the latest version form the website? 
It is too old, and does not include the fix.

ppetr:
Would it be possible to make a release that'd include the patch?

Psudonico:
> @Psudonico Do you use the latest version form the website? It is too old, and does not include the fix.

@hunbernd 
So the question is:  "where can I find the updated version with FIX?"
thanks

PR #2472, opened 2021-09-05
Fixes crash after: (EE) asked to create an invalidated profile that a…

### LINKED ISSUE #2463: Memory corruption if you add a lot of people
If you start adding a lot of friends, the software starts to become unstable. If you add more than 30-40 person, the software crashes. Sometimes, its enough to two or three new persons (or just add the same person again) to trip the program to a segmentation fault, especially if you have more contacts than 50. The profile stays unstable even after restarting it a few times, and it will lose all the newly added contacts, because Retroshare refuses to save the newly added peoeple automatically after adding it (and it expects you to close the program normally to save your profile).

hunbernd:
Fixes crash after: (EE) asked to create an invalidated profile that already exists

createInvalidatedProfile returned with iterator.end() in some cases.
updateInternalData dereferenced the result without checking it first.
It overindexed an array, read some memory trash, then wrote it into an another array.

Closes #2463

### DIFF
--- a/retroshare-gui/src/gui/common/FriendListModel.cpp
+++ b/retroshare-gui/src/gui/common/FriendListModel.cpp
@@ -1052,7 +1052,12 @@ std::map<RsPgpId,uint32_t>::const_iterator RsFriendListModel::createInvalidatedP
 
     if(rsPeers->getGPGDetails(pgp_id,hprof.profile_info))
     {
-        std::cerr << "(EE) asked to create an invalidated profile that already exists!" << std::endl;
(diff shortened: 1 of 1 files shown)


programmingzor:
Please merge this asap. By the way it took us almost 2 weeks to hunt this bug down. Someone may laughs, and says, what takes two weeks to insert 3 lines to the code, but this is the sad reality.

csoler:
There's nothing wrong about spending 2 weeks to hunt down a bug! Thx for the fix.

PR #2496, opened 2021-10-28
Fix NotifyQt::askForPassword QInputDialog is not created in GUI Thread.

PhenomRetroShare:
To test without it, change in:
/retroshare-gui/src/gui/GenCertDialog.cpp:665
QTimer::singleShot(30000, []() { rsNotify->clearPgpPassphrase(); } );
to
QTimer::singleShot(**100**, []() { rsNotify->clearPgpPassphrase(); } );

And create a new node.

### DIFF
--- a/retroshare-gui/src/gui/MainWindow.cpp
+++ b/retroshare-gui/src/gui/MainWindow.cpp
@@ -21,6 +21,7 @@
 #include <QColorDialog>
 #include <QDesktopServices>
 #include <QIcon>
+#include <QInputDialog>
 #include <QMessageBox>
 #include <QPixmap>
 #include <QStatusBar>
@@ -1688,4 +1689,20 @@ void MainWindow::setCompactStatusMode(bool compact)
 	hashingstatus->setCompactMode(compact);
 	ratesstatus->setCompactMode(compact);
 	//opModeStatus: TODO Show only ???
+}
+
+Gui_InputDialogReturn MainWindow::guiInputDialog(const QString& windowTitle, const QString& labelText, QLineEdit::EchoMode textEchoMode, bool modal)
+{
+
+	QInputDialog dialog(this);
+	dialog.setWindowTitle(windowTitle);
+	dialog.setLabelText(labelText);
+	dialog.setTextEchoMode(textEchoMode);
+	dialog.setModal(modal);
+
+	Gui_InputDialogReturn ret;
+	ret.execReturn = dialog.exec();
+	ret.textValue = dialog.textValue();
+
+	return ret;
 }
--- a/libretroshare/src/pgp/openpgpsdkhandler.cc
+++ b/libretroshare/src/pgp/openpgpsdkhandler.cc
@@ -1322,15 +1322,19 @@ bool OpenPGPSDKHandler::decryptTextFromFile(const RsPgpId&,std::string& text,con
 
 bool OpenPGPSDKHandler::SignDataBin(const RsPgpId& id,const void *data, const uint32_t len, unsigned char *sign, unsigned int *signlen,bool use_raw_signature, std::string reason /* = "" */)
 {
-	RsStackMutex mtx(pgphandlerMtx) ;				// lock access to PGP memory structures.
 	// need to find the key and to decrypt it.
-	
-	const ops_keydata_t *key = locked_getSecretKey(id) ;
-
-	if(!key)
+	ops_keydata_t *key = nullptr;
 	{
-        RsErr() << "Cannot sign: no secret key with id " << id.toStdString() ;
-		return false ;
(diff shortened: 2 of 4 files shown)


retroshare-gui/src/gui/MainWindow.cpp
```diff
@@ -1688,3 +1689,21 @@ void MainWindow::setCompactStatusMode(bool compact)
 	ratesstatus->setCompactMode(compact);
 	//opModeStatus: TODO Show only ???
 }
+
+// Called by other thread because Widgets must be created in the GUI thread.
+Gui_InputDialogReturn MainWindow::guiInputDialog(const QString& windowTitle, const QString& labelText, QLineEdit::EchoMode textEchoMode, bool modal)
+{
+
+	RsErr(this->thread(), " <-> ", qApp->thread());
```
G10h4ck: AFAIU you should use `RsDbg` or `RsInfo` here not `RsErr`

libretroshare/src/pgp/pgphandler.cc
```diff
@@ -1381,10 +1381,12 @@ bool PGPHandler::decryptTextFromFile(const RsPgpId&,std::string& text,const std:
 
 bool PGPHandler::SignDataBin(const RsPgpId& id,const void *data, const uint32_t len, unsigned char *sign, unsigned int *signlen,bool use_raw_signature, std::string reason /* = "" */)
 {
-	RsStackMutex mtx(pgphandlerMtx) ;				// lock access to PGP memory structures.
 	// need to find the key and to decrypt it.
-	
-	const ops_keydata_t *key = locked_getSecretKey(id) ;
+	const ops_keydata_t *key = nullptr;
+	{
+		RsStackMutex mtx(pgphandlerMtx) ;				// lock access to PGP memory structures.
```
G10h4ck: better use the `RS_STACK_MUTEX` macro so we avoid compile complaining and better debug is possible

libretroshare/src/pgp/pgphandler.cc
```diff
@@ -1404,35 +1406,38 @@ bool PGPHandler::SignDataBin(const RsPgpId& id,const void *data, const uint32_t
...
+	}
+	if(!secret_key)
+	{
+		std::cerr << "Could not obtain secret key. Signature cancelled." << std::endl;
+		return false ;
+	}
 
 	// then do the signature.
 
+	RsStackMutex mtx(pgphandlerMtx) ;				// lock access to PGP memory structures.
```
G10h4ck: better use the RS_STACK_MUTEX macro so we avoid compile complaining and better debug is possible

retroshare-gui/src/gui/MainWindow.cpp
PhenomRetroShare: Done

libretroshare/src/pgp/pgphandler.cc
PhenomRetroShare: Done

libretroshare/src/pgp/pgphandler.cc
PhenomRetroShare: Done

PhenomRetroShare:
@csoler ;)

libretroshare/src/pgp/pgphandler.cc
```diff
@@ -1381,10 +1381,12 @@ bool PGPHandler::decryptTextFromFile(const RsPgpId&,std::string& text,const std:
 
 bool PGPHandler::SignDataBin(const RsPgpId& id,const void *data, const uint32_t len, unsigned char *sign, unsigned int *signlen,bool use_raw_signature, std::string reason /* = "" */)
 {
-	RsStackMutex mtx(pgphandlerMtx) ;				// lock access to PGP memory structures.
```
csoler: Moving this mutex is really not a good idea: locked_getSecretKey(id) returns a pointer in the middle of a large array containing all key structures. If anything happens to the keyring between the getSecretKey and the use of it, then an undefined behavior will happen.

libretroshare/src/pgp/pgphandler.cc
```diff
@@ -1381,10 +1381,12 @@ bool PGPHandler::decryptTextFromFile(const RsPgpId&,std::string& text,const std:
...
 	// need to find the key and to decrypt it.
-	
-	const ops_keydata_t *key = locked_getSecretKey(id) ;
+	const ops_keydata_t *key = nullptr;
+	{
+		RS_STACK_MUTEX(pgphandlerMtx); // lock access to PGP memory structures.
+		key = locked_getSecretKey(id);
```
csoler: If you really need that mutex here, then copy the struct entirely. It's costly, but that shouldn't be a problem here.

PhenomRetroShare: Done

PR #2499, opened 2021-11-03
Fix RsThread when nothing to do and run finish before start.

PhenomRetroShare:


libretroshare/src/util/rsthreads.h
```diff
@@ -248,7 +248,7 @@ class RsThread
 	 * is called, any task necessary to stop the thread (aka inducing @see run()
 	 * to return in a coherent state) should be done in the overridden version
 	 * of this method, @see JsonApiServer for an usage example. */
-	virtual void onStopRequested() {}
+	void onStopRequested() {}
```
G10h4ck: Why you removed virtual here?

PR #2501, opened 2021-11-06
Add a notification when GxsId come back in lobby.

PhenomRetroShare:
This can be setted even if GxsId is offline by using its name in
TextBrowser.

### DIFF
--- a/retroshare-gui/src/gui/chat/ChatLobbyDialog.cpp
+++ b/retroshare-gui/src/gui/chat/ChatLobbyDialog.cpp
@@ -41,7 +41,9 @@
 #include "gui/msgs/MessageComposer.h"
 #include "gui/settings/RsharePeerSettings.h"
 #include "gui/settings/rsharesettings.h"
+#include "gui/SoundManager.h"
 #include "util/HandleRichText.h"
+#include "util/misc.h"
 #include "util/QtVersion.h"
 
 #include "retroshare/rsnotify.h"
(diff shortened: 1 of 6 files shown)


defnax:
@csoler

defnax:
@G10h4ck @csoler

defnax:
please check

defnax:
@csoler @G10h4ck

defnax:
please check this

csoler:
I tried it but didn't hear any sound. I think the problem is that you use the same notification than when a new msg arrives, so if sounds for new messages are desactivated, your new functionality will not work either. If I'm correctly understanding this, then you should probably create a new notification name and allow in preference to choose which sound to use, so that it's saved in Settings where the soundManager will look for it.

PhenomRetroShare:
Updated and I've added a tooltip to tell if the sound event is unchecked.

defnax:
But only sound notify?
would be nice some toaster or notify in chat room too if user its back again

test:
sound not worked, after i come back is unchecked the notify

PhenomRetroShare:
Why add a notify (like when someone talk to you)?
If you're not around, he may have left the chat room.
If you react 1 hour later, there is no interest...

PhenomRetroShare:
@csoler ;)

defnax:
@csoler this needs to be checked

csoler:
wanted to re-try this PR but checkout pr/2501 doesn't work. Might be related to the recent split into submodules.

PhenomRetroShare:
git checkout refs/remotes/origin/pr/2501
works fine here. (detached Head)

retroshare-gui/src/gui/chat/ChatLobbyDialog.cpp
```diff
@@ -411,7 +435,7 @@ void ChatLobbyDialog::init(const ChatId &/*id*/, const QString &/*title*/)
         if(rsIdentity->getIdDetails(gxs_id,details))
             break ;
         else
-            rstime::rs_usleep(1000*300) ;
+            std::this_thread::sleep_for(std::chrono::microseconds(1000*300)) ;
```
G10h4ck: you can just use `milliseconds(300)` instead

PhenomRetroShare: Sure

defnax:
@csoler

defnax:
Another idea, dont know if possible but when chat user joins he gets notify that i mentoined him with the right text what i wrote or send him optional as mail the mention

defnax:
offline chat mention would be nice, maybe someone mentioned me but i wasnt online in the chat, when i come back online show the mention with the text

PhenomRetroShare:
You already have notifications for that. ;)

PR #2515, opened 2021-11-29
Fix BoardPostDisplayWidget when editor redraw.

PhenomRetroShare:


csoler:
When I change the sorting (Top / New / Hot) ranks at left, the widget doesn't update. It does when I scroll. Could you find a way to call some update() method when we change the sorting?

defnax:
-Switch to Card View has issues too its often slow
-after scroll the page is updated

defnax:
> When I change the sorting (Top / New / Hot) ranks at left, the widget doesn't update. It does when I scroll. Could you find a way to call some update() method when we change the sorting?

this happens only on  card view, compact view has no issue with update

PhenomRetroShare:
Updated.
All was in editor kept shown.
So I fix it in RSTreeView.

PR #2517, opened 2021-12-01
Run on Android without Qt

G10h4ck:
This pull request contribute the necessary modifications to libretroshare and retroshare-service to not depend on Qt at runtime on Android at all.

When we will switch to cmake, and double check if the python version of JSON API generator is feature pair with the C++/Qt one we will not need Qt at build time.

Moving in this direction will make RetroShare much easier to embed on any Android application


G10h4ck:
Android continuous integration report here https://gitlab.com/elRepo.io/RetroShare/-/pipelines/420345107

G10h4ck:
AppVeyor reports failure but if you look at the log it says everything is good :-?

PR #2520, opened 2021-12-06
Fix PostedCreatePostDialog when reject show double message.

PhenomRetroShare:
And stacked widget if current index is changed in ui file.


defnax:
@csoler please check

defnax:
@G10h4ck

defnax:
@csoler

PR #2521, opened 2021-12-07
added classic stylesheet

defnax:


PhenomRetroShare:
Ok for me.

defnax:
@csoler

defnax:
@csoler can you check this?

defnax:
it done for merge @csoler

PhenomRetroShare:
It's stressful to wait other to merge... ;)
Be patient.

defnax:
yes i know :)

defnax:
@csoler

csoler:
* the file should be in qss/stylesheets/ like other stylesheets. You need to embed it into the software like other built-in stylesheets. Look into images.qrc for that. Otherwise it will not show up in the list anyway (it doesn't on linux at least)
* also this file is really light as compared to other stylesheets. Is that normal? It's very surprising especially when you compare to the "default" stylesheet, which should be as small as possible given its name.
* the name is somewhat conflicting with the default stylesheet because "standard" sounds like default too.

PhenomRetroShare:
@csoler : This qss is not compatible with all system them, So I ask @defnax to put it here like old qss.
It will be embedded by installer. 
default is always loaded (like when you've choose empty in list before), so this one only add white background to some widget.
By "standard" he think "like older version"...

defnax:
or the rename to standard?

PhenomRetroShare:
I don't know where @csoler see "standard". This "retroclassic" is fine for me.

csoler:
ok then!

csoler:
btw, why is appveyor not compiling fully?

PhenomRetroShare:
Too long time:
 Failed 8 days ago in 1 hr

defnax:
when it will be merged

PR #2534, opened 2021-12-23
Fixed Darkstyle textcolors for group category & private key

defnax:
image image


defnax:
@PhenomRetroShare look this colors now better on dark style instead of gray colors

defnax:
@csoler

PR #2540, opened 2021-12-31
Fixed to view attached pulse pictures

defnax:


retroshare-gui/src/gui/TheWire/PulseMessage.cpp
```diff
@@ -148,7 +148,7 @@ void PulseMessage::setRefImageCount(uint32_t count)
 	}
 }
 
-void PulseMessage::viewPicture()
+void PulseMessage::viewPictureOne()
```
csoler, line 151: Plz use sender() to get which widgets sends the signal. This way you can choose which picture to show and avoid duplicating all the code!

PR #2543, opened 2022-01-03
Removed compiler warnings in libretroshare

thunder2:


thunder2 (commented):
Is the function getCurrentTS needed anymore?

thunder2 (commented):
The parameter req_result_count of p3turtle::performLocalSearch_generic looks like an output parameter which is not initialized.

csoler:
> The parameter req_result_count of p3turtle::performLocalSearch_generic looks like an output parameter which is not initialized.

I think that's because the generic search result item obfuscates the number of search results. I suppose it would be possible to add a virtual method in this item type that returns the correct number everytime.

csoler:
> Is the function getCurrentTS needed anymore?

There's a dozen of copies of this function throughout the code. Would be nice to clean that up, only using one from rstime.h

PR #2552, opened 2022-01-13
added a msg in console when FS is ready

csoler:


defnax:
the friend server is not compatible with windows compile issues on network.cc
https://ci.appveyor.com/project/RetroShare58622/retroshare/builds/42193935

image

PR #2568, opened 2022-01-22
Fix to follow libretroshare master

PhenomRetroShare:
RS_Gui PR should be done after libretroshare merge to follow master.


csoler:
I think this PR is a mistake. There's no change in rs-gui. Should it be deleted?

PhenomRetroShare:
No, it's not a mistake.
For now RetroShare points to 55efaf9c730859ef3fc5c6f6049595c1225f9204 libretroshare commits, not to master.
@G10h4ck can you confirm the right way to do it?

Using QtCreator, it ask you if you want to update submodule with `git submodule update`.
Opening a libretroshare file, you can see that it's not on master branch.

G10h4ck:
@PhenomRetroShare yeah this is the right way to update the submodule reference on the repository here

PR #2590, opened 2022-02-24
update MacOS Section & Install guide

defnax:


retroshare.pri
```diff
@@ -131,15 +131,9 @@ CONFIG *= no_use_dht_stunner_ext_ip
 use_dht_stunner_ext_ip:CONFIG -= no_use_dht_stunner_ext_ip
 
 # To select your MacOsX version append the following assignation to qmake
-# command line "CONFIG+=rs_macos10.11" where 10.11 depends your version
-macx:CONFIG *= rs_macos10.11
-rs_macos10.8:CONFIG -= rs_macos10.11
```
G10h4ck, line 136: Why removing all oldel MacOS versions? Are thei EOL ?

G10h4ck:
Why removing support for all older MacOS versions? Are they EOL ?

retroshare.pri
defnax: thats very outdated we need need only one entry, and on compile it shows up all versions which is not needed.
we need keep simple, else we need every time 8-9 lines edit to get compiled.

Developer need now edit only 1 or 2 lines and not 9 lines for set macos version.

G10h4ck: isn't adding 'CONFIG+=rs_macosXX.YY' to command line enough? those lines you removes just disables all the options except the one chosed

defnax: Yes that will work too, its bad that we need edit 9 lines to set macosversion for compile
We need more simple

PhenomRetroShare: Qt warns that it's only been tested with version 10.15 of the platform SDK when using 11.1.

PR #2593, opened 2022-03-08
added list of already received peers from FS so that we do not request them again

csoler:


defnax:
hope this is windows compatible, macos always broken

PR #2602, opened 2022-03-25
added cmake list for files in GUI

csoler:
I did the easy part first ;-)


defnax:
samv3 is a issue on macos compile, that broked the macos platform

csoler:
I'm merging this although it doesn't compile yet with cmake. This is going to help improving it.

PR #2605, opened 2022-04-01
[WIP] Improvements to FriendsDialog

csoler:


defnax:
i dont like this idea , but when you changing why not generate for every friend a own colors?
every friend blue looks boring for me

csoler:
what do you mean? Having plenty of colors everywhere makes the software look like Disneyworld. That's not necessarily a good idea.

csoler:
ah I see. Yes we talked about using the avatar generator of GxsId for nodes (that makes sense), but we can only do this when we have some avatar generator for GxsId to replace (we cannot use the same!). I could not find any C/C++ implementation for now.

csoler:
This PR only replaces the "man" default icon with some "network node" default icon.  But it's a temporary thing if you consider the above changes.

defnax:
whats about the forum combobox, i sended you the svg icons

csoler:
For the forum combobox I dont know how to use 3 icons in a button. We have 2 possibilities: use 3 buttons with pressed/unpressed and make sure that when you press one, the others will deselect, or use a 3-states button. What's best?

PR #2611, opened 2022-04-06
Added banned item

defnax:


csoler:
I don't see the point in this PR: banned items are already shown, and if you want to see them easily, just sort w.r.t. the reputation column. Originally I thought that this PR was intended to display the items that are kept banned by the reputation system so as to not accept them again, but this is just a list of GxsIds, so you cannot display info about them anyway.

PR #2612, opened 2022-04-09
Fix MacOS build

PhenomRetroShare:
Need to use:
CONFIG+=rs_macos11.1 CONFIG+=no_rs_sam3 CONFIG+=no_rs_sam3_libsam3
CONFIG+=c++14


G10h4ck:
Seems good to me

PR #2619, opened 2022-05-03
gui: chat: add auto shrink chattextedit

chelovechishko:
Didn't check it with other types than usual chat itself.
Increased minimum size for chattextedit 'cause it's not big enough for my letters.


chelovechishko:
I didn't check distant stuff.

chelovechishko:
No, i see nothing wrong with this commit in this pr.
Overall, this is revert of #2567 .

Issue #30, opened 2015-08-14
 Create a simple Skype-like alternative Qt5 QML Gui 


defnax:
we propose to make a simple Skype-like alternative to retroshare-gui. only Chat, Lobbies and Voip/Video. Not sure what to call it - but we will probably use a different brand from retroshare. 
Here you can find the example qml model:
https://github.com/drbob/rsqml-models

make a easy clean gui like skype:
https://skypeblogs.files.wordpress.com/2014/10/mac1.png

or look example project:
https://ring.cx/

QML retroshare client:
https://github.com/Emotyco

ohhmm:
consider plug in skins system

defnax:
QML gui by @kdebiec needs to be continoued
https://github.com/Emotyco

PR #2621, opened 2022-05-21
Raise max simultaneous CRC requests to 500

jolavillette:
Raise max simultaneous CRC requests to 500 to allow handling of download speed up to 100 MB/s


csoler:
You made a PR to the wrong repository (Retroshare instead of libretroshare), so your PR contains lots of changes that should probably not not there. I suppose I can delete it. 
Also, I had doen a PR doing the same change in libretroshare 2 days ago, waiting for merge (#30), and I just merged it.

PR #2623, opened 2022-06-22
V0.6 token system

csoler:
Getting rid of the Token system in UI


defnax:
hope by the way you can fix token stuff for wire and wiki service too?

csoler:
> hope by the way you can fix token stuff for wire and wiki service too?

Try to do it yourself. See how it's done in other services.

PR #2657, opened 2022-11-21
[Messages] Feature to expand To field 

defnax:
i has created this changes to view the mass mail recipients to resize the To field with recipients counting.
recreated old PR #1951

image
image

### DIFF
--- a/retroshare-gui/src/gui/msgs/MessageWidget.cpp
+++ b/retroshare-gui/src/gui/msgs/MessageWidget.cpp
@@ -135,6 +135,8 @@ MessageWidget::MessageWidget(bool controlled, QWidget *parent, Qt::WindowFlags f
 	isWindow = false;
 	currMsgFlags = 0;
 	expandFiles = false;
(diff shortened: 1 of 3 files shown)


csoler:
when you have many recipients, it seems that it makes the place very messy. Couldn't we use a tooltip instead?

defnax:
it was only idea, i dont know how email clients handle it with lots of recipients for display.
Not sure if tooltip is a good idea
i know only from work, when get a lots of recipeints you can collapse the rest of recipients to view it.

csoler:
They do the same as you did. The problem we have is that there's not much space to expand a multi-line widget.

csoler:
Could you solve the conflicts? We could potentially merge this and improve it later on. Still, I'm comparing with what master branch currently does, and it already uses a widget with a slider to display large amounts of recipients at a time.

defnax:
i can have a look tonight, now at work. yes it can be more improved.

defnax:
#2660

PR #2658, opened 2022-12-01
Full rework on Mail system

csoler:
I had to make a full pass over the mail system since some information was actually not stored, which caused most of the bugs in the UI.
This needs testing.

*WARNING*
- the PR allows to load mails from before and convert them to the new format.  This goes only one way, consequently, for testing, making a backup of old data/ directory is essential
- pending messages are not restored, mostly because some information wasn't actually saved (destination of each pending copy).
- This PR needs PR 59 from libretroshare

### DIFF
--- a/retroshare-gui/src/gui/common/FriendSelectionWidget.cpp
+++ b/retroshare-gui/src/gui/common/FriendSelectionWidget.cpp
@@ -127,11 +127,45 @@ FriendSelectionWidget::FriendSelectionWidget(QWidget *parent)
 
 	/* Refresh style to have the correct text color */
 	Rshare::refreshStyleSheet(this, false);
+
+    mEventHandlerId_identities = 0;
+    rsEvents->registerEventsHandler( [this](std::shared_ptr<const RsEvent> event) {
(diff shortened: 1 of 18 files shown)


defnax:
By the was when you are working on the mail and can you add group posibilty to send mail to a group which
like Contacts and dynamicly generated groups of gxs contacts
on old Peer messaging we had possible to send a mail to group of peer contacts (friends,coworkers, favorites)

maybe this idea can be used on both places on people & mails.

[ Mail ] Add a Button for Generate a new GXS Contacts Group(AddressBook), and a Dialog where you manage your Contacts List(Add/Store/Load/Edit).
[ Mail ] [Composer] Show the list of GXS Group Contacts which are dynamicly generated by myself [ RS Dev Group, RS Group Testers, WIRETESTERS, Favorites, Coworkers,....] for selection
[ People ] Generate dynamicly list of gxs contacts groups of your favorite Contacts for example [ RS Dev Group, RS Testers, Favorites, Coworkers,....]

similiar like this but use it for people&mail with gxs identities
image

defnax:
here the old node groups in composer rs 0.6.0 , where it showed up on composer, similiar needs with gxs groups
image

PR #2660, opened 2022-12-13
[Messages] Feature to expand To field 

defnax:


defnax:
@csoler look if you like it.
i can remove else the icon from the button, or add the counting to a extra label like before.

defnax:
#2661

PR #2668, opened 2023-01-13
Changed Moderators list to use with a Label

defnax:
Changed Moderators list to use with a Label, less space is used then before


defnax:
@csoler

PR #2683, opened 2023-02-16
[WIP] added help for commandline params

csoler:


csoler:
this problem is already handled in more recent PR #2844.

PR #2689, opened 2023-03-12
added RSNewWebUI as a submodule so as to help packaging

csoler:


G10h4ck:
looks good to me, this help mainly with QMake, once we rollout CMake we don't even need to use submodules anymore ;-)

PR #2693, opened 2023-03-27
Moved settings check for popupflags to AddToaster

defnax:


defnax:
i need to update tihs this weekend

PR #2712, opened 2023-04-17
Fix removed wordrap from Title label and use ElidedLabel

defnax:


defnax:
There is a issue with wordwrap and QLabel
https://forum.qt.io/topic/128286/qlabel-with-wordwrap-fixed-vertical-policy-and-stretched-for-the-whole-qdialog-s-width-fails-to-resize-height/4

PR #2751, opened 2023-07-04
router statistics - sort tunnels by traffic (gui)

RetroPooh:


defnax:
was nice if we can have more stats for downloaded/uploaded/traffic rs shows only speed in stats 

Downloaded / Uploaded Column per friend in MB/GB/TB
Downloaded / Uploaded traffic per Service  in MB/GB/TB

example here needs show the traffic in in MB/GB/TB
image

csoler:
does it need another PR? because it crashes.

defnax:
@RetroPooh update?

defnax:
@csoler close we implemented always sorting by traffic works

PR #2778, opened 2023-10-24
Added colors for the service output

defnax:


G10h4ck:
Do this work also on non-windows platforms?

defnax:
> Do this work also on non-windows platforms?

needs to be tested i get this thing from here, normaly cross platform feature
https://stackoverflow.com/questions/5947742/how-to-change-the-output-color-of-echo-in-linux

defnax:
image

csoler:
I just tested on linux. It works and for sure it makes the output more clear.

csoler:
in the future it would be better to encapsulate the coloring into some function, because it puts some crap in the strings everywhere we use it.

PR #2780, opened 2023-10-26
update output colors

defnax:


defnax:
@csoler i has seperated the color strings from the text, maybe now codes looks better.
`"\033[0;32m" "Please enter your PGP password for key:\n    " "\033[0m"`

csoler:
It's better. In the future it will be nice to put this into a method such as  std::string rs_print::colored(RsTextColor, std::string). Not a priority now, but this will avoid having these weird strings everywhere.

PR #2799, opened 2023-11-30
Allow to override openssl security level

hunbernd:
Let RS to load old profiles if started with -ssllevel 0

This PR also needed: https://github.com/RetroShare/libretroshare/pull/128


csoler:
This was fixed by #141, so it's irrelevant now.

PR #2805, opened 2023-12-17
Added a dock menu for macOS

defnax:
image

### DIFF
--- a/retroshare-gui/src/gui/MainWindow.cpp
+++ b/retroshare-gui/src/gui/MainWindow.cpp
@@ -28,6 +28,7 @@
 #include <QString>
 #include <QUrl>
 #include <QtDebug>
+#include <QMenuBar>
 
 #include <retroshare/rsplugin.h>
 #include <retroshare/rsconfig.h>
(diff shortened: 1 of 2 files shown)


retroshare-gui/src/gui/MainWindow.h
```diff
@@ -324,6 +326,10 @@ private slots:
 
     QString nameAndLocation;
 
+    /** The menubar (Mac OS X only). */
```
csoler, line 329: since this is MacOS code only, you should use the #ifdef here as well. No need to leave an uninitialized/unused pointer on other systems

retroshare-gui/src/gui/MainWindow.cpp
```diff
@@ -651,10 +653,42 @@ void MainWindow::createTrayIcon()
...
 
+#if defined(Q_OS_DARWIN)
+    createMenuBar();
+#endif
+
     connect(trayIcon, SIGNAL(activated(QSystemTrayIcon::ActivationReason)), this, SLOT(toggleVisibility(QSystemTrayIcon::ActivationReason)));
     trayIcon->show();
 }
 
+/** Creates a new menubar for macOS */
```
csoler, line 664: same here: ifdef missing, unless this is used on other systems too but up to know we lived without it.

defnax: i will create a new fresh pr

PR #2806, opened 2023-12-18
Added logo for retroshare-service on macOS

defnax:


defnax:
new dock menu code added here @csoler

PR #2827, opened 2024-01-12
[macOS] Fixed some font issues on macos platform

defnax:


defnax:
some widgets in rs does not use the default fonts on macOS, it gets very tiny on mac,
i find a temporary solution with 
https://forum.qt.io/topic/151313/font-sizes-too-small-on-macos

```
#if defined(Q_OS_DARWIN)
    QFont font = QFontDatabase::systemFont(QFontDatabase::GeneralFont);
    ui.lobbyTreeWidget->setFont(font);
#endif
```

this classes has issues to get default fonts:
Ui's using RSTreeView
Ui's using RSTreeWidget
QListWidget's
QLabel's

csoler:
is this a Qt bug, or do we have some specific fonts for these widgets in e.g. default.css ? That would also explain that they do not use system fonts.

defnax:
there no fonts more used in default.css
rs get fonts from ui files
```
/* Default stylesheet
   This file is used as default for all stylesheets and can be overloaded
   Don't use font-size or font-weight, they are defined in .ui files.
   Icon size are defined in .cpp files.
```

PR #2832, opened 2024-01-16
Added export friendslist to profile page

defnax:


csoler:
The same code already exists in FriendList.cpp. That means duplication of code and duplication of functionality. I agree that the export function could be in node preferences, but then can you use FriendList::exportFriends() so that the same code is used in both places?

defnax:
done #2847

PR #2836, opened 2024-01-29
Removed "chat rooms" string from tree item

defnax:


defnax:
@csoler you forget this this was for translations too, but you merge th e other one first

PR #2844, opened 2024-02-13
Cleaned up command line parameter handling (to be used with #143)

csoler:


csoler:
This PR rewrites the handling of parameters so they can be parsed using the argstream class (as they are in retroshare-service), and so that all options are displayed with --help. This needs to be tested on Windows and MacOS, especially passing options to a running RS server such as openning a rscollection file from the system file manager. Needs PR 143.

retroshare-gui/src/main.cpp
```diff
@@ -233,28 +223,57 @@ feenableexcept(FE_INVALID | FE_DIVBYZERO);
...
 #endif // ifdef RS_AUTOLOGIN
-
+    as 		>> help('h',"help",QObject::tr("Display this help").toStdString().c_str());
+
+    if(as.helpRequested())
+    {
+        RsInfo() << "\n" <<
+            "+================================================================+\n"
+            "|     o---o                                             o        |\n"
+            "|      \\ /               - Retroshare GUI -            / \\       |\n"
```
chelovechishko: Does it intended to be misaligned (on the right side of the icon), or is it a stylish style?

csoler:
"\\" counts as one character.

retroshare-gui/src/main.cpp
csoler: "\\\\" counts as one character.

defnax:
image
its does not compile here


and fix in main.cpp line 708
```
#ifdef __APPLE__
        /* For OSX, we set the default to "cleanlooks", as the AQUA style hides some input boxes
         * only on the first run - as the user might want to change it ;)
         */
        QString osx_style("cleanlooks");
        RsApplication::setStyle(osx_style);
        Settings->setInterfaceStyle(osx_style);
#endif
```

defnax:
this not solves the issue for me get the different issue at last
```
  static QStringList _links;           /**< List of links passed by arguments. */
  static QStringList _files;           /**< List of files passed by arguments. */
```

i has disabled this code not sure if we need it, but now compile on mac is fixed

```
#ifdef __APPLE__
bool RsApplication::event(QEvent *event)
{
  /*switch (event->type()) {
    case QEvent::FileOpen:{
      QFileOpenEvent* fileOE = static_cast<QFileOpenEvent *>(event);
      QStringList args;
      if (! fileOE->file().isEmpty()) {
        _files.append(fileOE->file());
        emit newArgsReceived(QStringList());
        return true;
      } else if (! fileOE->url().isEmpty()) {
        _links.append(fileOE->url().toString());
        emit newArgsReceived(QStringList());
        return true;
      }
    }
    default:
    return QApplication::event(event);
  }*/
}
#endif
```

defnax:
maybe you can fix this?
i saw it was added by @PhenomRetroShare 

```
#ifdef __APPLE__
bool Rshare::event(QEvent *event)
```

- https://github.com/RetroShare/RetroShare/commit/10bb542e835c61118eca3c021a5616a07a2d0d3f#diff-8ca63a9c854845df4ec6607fbb62ae208b7ba6a40ba81a91c794c2736e6353df

```
Allow only one instance for useLocalServer option activated. If exists,
pass to it arguments.

For now, only allow to pass RsCollection file and retroshare://
protocol.
```
- https://github.com/RetroShare/RetroShare/commit/10bb542e835c61118eca3c021a5616a07a2d0d3f#

defnax:
maybe this is needed for macos added by phenom and you removed but not tested yet, last night not had time so much for test

```
  /** Returns links passed by arguments. */
  static QStringList* links() { return &_links; }
  /** Returns files passed by arguments. */
  static QStringList* files()  {return &_files; }
```

```
  static QStringList _links;           /**< List of links passed by arguments. */
  static QStringList _files;           /**< List of files passed by arguments. */
```

csoler:
> this not solves the issue for me get the different issue at last
> 
> ```
>   static QStringList _links;           /**< List of links passed by arguments. */
>   static QStringList _files;           /**< List of files passed by arguments. */
> ```
> 
> i has disabled this code not sure if we need it, but now compile on mac is fixed
> 
> ```
> #ifdef __APPLE__
> bool RsApplication::event(QEvent *event)
> {
>   /*switch (event->type()) {
>     case QEvent::FileOpen:{
>       QFileOpenEvent* fileOE = static_cast<QFileOpenEvent *>(event);
>       QStringList args;
>       if (! fileOE->file().isEmpty()) {
>         _files.append(fileOE->file());
>         emit newArgsReceived(QStringList());
>         return true;
>       } else if (! fileOE->url().isEmpty()) {
>         _links.append(fileOE->url().toString());
>         emit newArgsReceived(QStringList());
>         return true;
>       }
>     }
>     default:
>     return QApplication::event(event);
>   }*/
> }
> #endif
> ```

i believe that this was added in order to capture when the system opens a rscollection or a retroshare link. Normally Qt does it cross-systems using the mechanism that is implemented outside of the __MACOS__ section. I'm not sure why MacOS needs this specific section. Maybe Phenom can tell us more about this.

csoler:
> maybe this is needed for macos added by phenom and you removed but not tested yet, last night not had time so much for test
> 
> ```
>   /** Returns links passed by arguments. */
>   static QStringList* links() { return &_links; }
>   /** Returns files passed by arguments. */
>   static QStringList* files()  {return &_files; }
> ```
> 
> ```
>   static QStringList _links;           /**< List of links passed by arguments. */
>   static QStringList _files;           /**< List of files passed by arguments. */
> ```

if that is needed only on macOS, then you need to add a ifdef around it.

defnax:
new pr

- https://github.com/csoler/RetroShare/pull/34

PR #2847, opened 2024-02-26
Added export Friendslist to Profile Page

defnax:
- updated version of https://github.com/RetroShare/RetroShare/pull/2847

### DIFF
--- a/retroshare-gui/src/gui/settings/CryptoPage.cpp
+++ b/retroshare-gui/src/gui/settings/CryptoPage.cpp
@@ -34,6 +34,7 @@
 #include <gui/connect/ConfCertDialog.h>
 #include <gui/profile/ProfileManager.h>
 #include <gui/statistics/StatisticsWindow.h>
+#include <gui/common/NewFriendList.h>
 
 #include <retroshare/rspeers.h> //for rsPeers variable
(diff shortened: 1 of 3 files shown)


defnax:
@csoler

csoler:
There's no point in allocating a NewFriendList (which takes quite a lot of RAM) since there is one already in the Network page. Here you're basically allocating memory that is most probably not going to be used. There's two good ways to do what you need:
1 - use a slot in CryptoPage that does NewFriendList().exportFriendListClicked() (this allocates a NewFriendList only temporarily and then destroys it)
or better:
2 - make NewFriendList::exportFriendList() static, and from CryptoPage, just call this method as a static method.

defnax:
hi i doesnt get it worked

```
connect(ui.exportfriendslist,SIGNAL(clicked()), this, SLOT(exportFriendsList()) );


void CryptoPage::exportFriendsList()
{
    NewFriendList::exportFriendList();
}

```

and in NewFriendList.h
```
Public:

	static  bool exportFriendlist(QString &fileName);
```

its public im not sure why it does not find it?

image

defnax:
- new version https://github.com/RetroShare/RetroShare/pull/2849

PR #2849, opened 2024-03-15
Added export Friendslist to Profile Page

defnax:


csoler:
could you make 1 button "export friends list" and one "export profile"? There's no point in having text + button "export". That's not very important of course. 
I also noticed something: when exporting, locations for nodes with the current profile do not show up in the exported file. Is that intended?

defnax:
> I also noticed something: when exporting, locations for nodes with the current profile do not show up in the exported file. Is that intended?

I dont remember more about locations how it was , but Importing Friendslist with Groups not works more since few years

defnax:
renamed the buttons @csoler

PR #2850, opened 2024-03-17
Added pinned column for forums

defnax:


defnax:
@csoler

csoler:
I haven't tried it yet, but what is the idea? We have a pinning system for posts already, that is controlled by the forum owner. Pinned posts display always on top.

defnax:
you did coded pinned in a wrong column, you coded a dirty to use same column for read/unread
this kind i never do like how you did.
i maked it now more clear looks better when its have own column instead use the hybrid mode
i separeted the pinned from the read column

defnax:
else i can show tonight how it looks before and after

csoler:
But then you add a column that is unused for 99.99% of the posts ?

defnax:
i forget this, then maybe we can add a option hide column when nothing pinned

PR #2867, opened 2024-07-09
Circle Avatars feature

defnax:


csoler:
I dont' understand how you store the circle avatar: it should normally be in the data part of the circle (but there's no place for it now). So it's a local information? But this doesn't make sense since other members in the same network will see a different avatar.

PR #2880, opened 2024-09-12
Added additional chat description field for add rules, links & infos

defnax:
Screenshots
image

image
image


defnax:
@csoler

csoler:
adding a description to chat rooms is a good idea. However it needs to be added as a new field to the relevant chat room item. If you hack the description into the title of the chat room this is not going to be backward compatible.

defnax:
what you mean with "However it needs to be added as a new field to the relevant chat room item" you dont mean in libretroshare? 
See screenshots seems you dont test my prs

csoler:
I mean that chat rooms are advertised to friends using a specific RsItem. If you want to add an extra field, you shouldn't try to smuggle the new field inside an existing one, because it's going to be a problem for non patched peers in addition to not being a proper way to do it. Of course for that you need a new RsItem (or make sure that the extra field is not a problem for old peers).

PR #2882, opened 2024-09-25
Fixes Chat notify & Chat room improvements

defnax:


JohannesKleine:
Tested between master+pr and obs on Debian 12. No different behavior noted. What i am supposed to see? Maybe it was not broken on Debian.

defnax:
cyril doesnt wanted this but its a fix

PR #2907, opened 2025-01-04
a few small fixes

### LINKED ISSUE #2889: UI not updating i2p port numbers
Arch Linux with RetroShare 0.6.7a (from AUR)

===

Preferences > Hidden Service Configuration > Incoming

The I2P Address has no port box. A port box appears if a port is added to the end of the address. But this port box duplicates the port at the end of the address and doesn't work, maybe because it results in an address of the form ADDRESS.b32.i2p:#####:#####

Where it says "For I2P: See http://127.0.0.1:7657/i2ptunnelmgr" this is wrongly assuming the I2P router is set up on the localhost.

===

Preferences > Hidden Service Configuration > Automatic

All options are greyed out. Selecting advanced mode doesn't activate them. Selecting advanced mode isn't persistent after restart - the box just becomes unchecked again.

===

Preferences > Hidden Service Configuration > Outgoing Manual Tor/I2P

Changing the port on the I2P Socks Proxy turns the light green but the information immediately reverts to the (wrong) default. 

===

All of these settings should be in a conf file. If they already are, then the UI should say where it's saving the settings.

Some of these problems look like they might be permissions issues, but a normal user can't access a Node created as root so usually we want to edit the conf file with sudo privileges and then run the UI as the normal user.

dbear496:
Fixes the manual outgoing I2P port settings UI. Previously, the outgoing proxy port could not be changed when in manual I2P mode -- the entry box would revert to the previous value as soon as I click away. This PR fixes the issue by properly calling the `saveAddresses` method when the port value is edited. I believe this (at least partially) fixes #2889 .

Fixes removing JSON API tokens. The API token map is keyed by the username, not the whole string, so the username portion must be isolated before removing.

Allows voting for one's own identity. This addresses an anonymity issue: If it is not possible to vote for one's own identity, then it is possible for someone to prove that they don't hold a pseudonymous identity by casting a vote for it. Though this is only a small loss of anonymity, it is nonetheless a loss of anonymity, so I believe it is important to be able to vote for one's own identity. This change was suggested by radium-226 in this forum post: retroshare://forum?name=voting%20for%20own%20identity&id=cdbc2cb6881da77018e1bd86f90711e4&msgid=c395d8ce1260a3e781b65a93f873e9080de2bde3 .

Removes the trailing newline when writing single links to the clipboard. This is very useful for pasting RS links to a terminal. This change was suggested by radium-226 in this forum post: retroshare://forum?name=Removing%20the%20newline%20from%20copied%20links&id=cdbc2cb6881da77018e1bd86f90711e4&msgid=afbafd9371b7a82e735a648318c6871167b104f5 .

### DIFF
--- a/retroshare-gui/src/gui/Identity/IdDialog.cpp
+++ b/retroshare-gui/src/gui/Identity/IdDialog.cpp
@@ -50,7 +50,7 @@
 #include "util/rsdebug.h"
 
 #include "retroshare/rsgxsflags.h"
-#include "retroshare/rsmsgs.h" 
+#include "retroshare/rsmsgs.h"
 #include "retroshare/rspeers.h"
 #include "retroshare/rsservicecontrol.h"
 
@@ -167,7 +167,7 @@ IdDialog::IdDialog(QWidget *parent)
 	// This is used to grab the broadcast of changes from p3GxsCircles, which is discarded by the current dialog, since it expects data for p3Identity only.
 	//mCirclesBroadcastBase = new RsGxsUpdateBroadcastBase(rsGxsCircles, this);
 	//connect(mCirclesBroadcastBase, SIGNAL(fillDisplay(bool)), this, SLOT(updateCirclesDisplay(bool)));
-    
+
 	ownItem = new QTreeWidgetItem();
 	ownItem->setText(RSID_COL_NICKNAME, tr("My own identities"));
 	ownItem->setFont(RSID_COL_NICKNAME, ui->idTreeWidget->font());
@@ -253,7 +253,7 @@ IdDialog::IdDialog(QWidget *parent)
 
 	connect(ui->filterLineEdit, SIGNAL(textChanged(QString)), this, SLOT(filterChanged(QString)));
 	connect(ui->ownOpinion_CB, SIGNAL(currentIndexChanged(int)), this, SLOT(modifyReputation()));
-	
+
 	connect(ui->inviteButton, SIGNAL(clicked()), this, SLOT(sendInvite()));
 	connect(ui->editButton, SIGNAL(clicked()), this, SLOT(editIdentity()));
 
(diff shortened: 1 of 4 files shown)


retroshare-gui/src/gui/Identity/IdDialog.cpp
```diff
@@ -2266,29 +2263,29 @@ void IdDialog::IdListCustomPopupMenu( QPoint )
...
+      contextMenu->addAction(FilesDefs::getIconFromQtResourcePath(":/icons/png/thumbs-neutral.png"), tr("Set neutral opinion"), this, SLOT(neutralPerson()));
+
+		if(n_negative_reputations == 0)
+      contextMenu->addAction(FilesDefs::getIconFromQtResourcePath(":/icons/png/thumbs-down.png"), tr("Set negative opinion"), this, SLOT(negativePerson()));
 
 		if(one_item_owned_by_you && n_selected_items==1)
 		{
 			contextMenu->addSeparator();
 
-			contextMenu->addAction(QIcon(""),tr("Copy identity to clipboard"),this,SLOT(copyRetroshareLink())) ;
```
G10h4ck, line 2291: Is this going away because already done at line 2271 for all identities ?

G10h4ck (approved):
It seems good to me, just one inline question for clarification ;-)

retroshare-gui/src/gui/Identity/IdDialog.cpp
dbear496: Previously, this was done separately in two places: on line 2270 for non-own identities and on line 2291 for own identities. Now it is done only once for both cases on line 2272.

G10h4ck: Perfect

csoler (approved):

PR #2924, opened 2025-03-16
Fix distant chat status

defnax:


defnax:
@csoler fix for status message was wrong till my pr fixes it

csoler:
I don't understand why: when the friend is offline then the messages will stay on your node and will be send when it gets online. This is true for chat and distant chat.

PR #2933, opened 2025-04-13
Fixed icon to use png

defnax:
issue resize with svg does not work


retroshare-gui/src/gui/gxs/GxsIdTreeWidgetItem.cpp
```diff
@@ -201,7 +201,7 @@ void GxsIdTreeItemDelegate::paint(QPainter *painter, const QStyleOptionViewItem
     // additional Qt role only to determine the number of destinations.
 
     if(dt == "")
-        ownOption.icon = FilesDefs::getIconFromQtResourcePath(":/icons/svg/people2.svg");
+        ownOption.icon = FilesDefs::getIconFromQtResourcePath(":/icons/png/people2.png");
```
csoler, line 204: Why this change? svg is most of the time much smaller than png and we already have memory problems.

defnax: svg resize not works on a mail item bug, but png works

defnax: ins mails columns is this bug
image

defnax: as png version 5,23 kb
in svg 3.38kb

JohannesKleine:
I believe the bug lays some where else. I got avatar if it is from foo@id and foo(my pc)@id shows no avatar. So maybe the way it processes the two different mail types has something to do with it. Notice mouse over gives you on avatars profile information and on icons from system masseges and foo(my pc)@id mails does mouse over not work.

defnax:
@csoler @JohannesKleine  u think svgs are not resizing, then the rest of column get affected one bug make buggy on all items

PR #2936, opened 2025-04-25
Updates for macos sequoia to run retroshare-service

redDwarf03:


retroshare.pri
```diff
@@ -150,37 +150,37 @@ rs_jsonapi:CONFIG -= no_rs_jsonapi
 # To enable forums indexing append the following assignation to qmake command
 # line "CONFIG+=rs_deep_forums_index"
 CONFIG *= no_rs_deep_forums_index
-rs_deep_forums_index:CONFIG -= no_rs_deep_forums_index
+rs_deep_forums_index:DEFINES *= RS_DEEP_FORUMS_INDEX
```
csoler: This and the other ones like this one is incorrect. 
also was the previous version blocking compilation??

redDwarf03: i try with old values and it's ok. i will rollback

defnax:
hi but this place i need for retroshare gui compile else i would get trouble to find macos version compiling retroshare 
image
i think you has not tested compile retroshare if its broken now with this

redDwarf03:
> hi but this place i need for retroshare gui compile else i would get trouble to find macos version compiling retroshare image i think you has not tested compile retroshare if its broken now with this

Yeah, i retested without this update. it works. i rollback

defnax:
@csoler check?

retroshare-service/CMakeLists.txt
```diff
@@ -113,6 +113,23 @@ target_link_libraries(${PROJECT_NAME} PRIVATE retroshare)
...
+target_link_directories(${PROJECT_NAME} PRIVATE
+    ${PROJECT_SOURCE_DIR}/../supportlibs/librnp/Build/src/lib
+    ${PROJECT_SOURCE_DIR}/../supportlibs/librnp/Build/src/libsexpp
+)
+
+# Link libraries by name, including RNP libs and dependencies first
+target_link_libraries(${PROJECT_NAME} PRIVATE
+    rnp             # Link by name
+    sexpp
+    botan-3
```
JohannesKleine, line 126: -- Looking for BOTAN_HAS_DL_PUBLIC_KEY_FAMILY - not found
CMake Error at src/lib/CMakeLists.txt:178 (message):
  A required botan feature is missing: DL_PUBLIC_KEY_FAMILY

if compiled with qmake normal gui version

JohannesKleine: botan is installed via brew botan (version 3.)

JohannesKleine: i was able to compile it, but only with botan-2

PR #2937, opened 2025-04-27
Added retroshare links for group author in usage stats

defnax:


JohannesKleine:
sum should only show one line on the left site, when user is selected

showing errors for unfilled data things when a service is selected

JohannesKleine:
Just omit errors and label it correctly like when a service is selected not just item #01 etc..

defnax:
Hi im not coded this stats, we need help to improve the statistics pages you are welcome to help😅
I added only retroshare links in the people->usage statistics when owner of a identity has board/forum/channel to link as rs link

defnax:
I think your comment has nothing todo with my added changes 😊

defnax:
@csoler

csoler:
I cannot reproduce the bug that johannesKleine reports.

retroshare-gui/src/gui/Identity/IdDialog.cpp
```diff
@@ -1889,8 +1889,18 @@ QString IdDialog::createUsageString(const RsIdentityUsage& u) const
...
-        	return tr("Group author for group %1 in service %2").arg(QString::fromStdString(u.mGrpId.toStdString()),service_name);
-        break ;
+	{
+		RetroShareLink link;
+
+		if(service_type == RetroShareLink::TYPE_CHANNEL )
+			link = RetroShareLink::createGxsGroupLink(service_type,u.mGrpId,tr("Channel"));
+		else if(service_type == RetroShareLink::TYPE_POSTED )
+			link = RetroShareLink::createGxsGroupLink(service_type,u.mGrpId,tr("Board"));
+		else
```
csoler, line 1899: This is invalid because if we add a new service (not in the switch) for instance adding a plugin, then it will be reported as a forum. If you don't know how to grab the RetroShareLink type for a plugin (may not even be possible) then dont create a link in this case.

csoler (changes requested):

PR #2942, opened 2025-05-28
fix typo Stared -> Starred

### LINKED ISSUE #2895: Misspelling of "starred" in Mail system
image
I'm going to assume the left circle is a typo since the right one is spelled correctly.

It would be in [this file.](https://github.com/RetroShare/RetroShare/blob/38e82adaeaadb5af6d542beb65cf3c0026d739ac/retroshare-gui/src/gui/msgs/MessagesDialog.cpp#L514)

JohannesKleine:
fixes #2895


JohannesKleine:
I assume the binary translation files will be upgraded through transiflex. I compiled it with this change and it works so far that it just needs those binary updates to show correctly in non English, otherwise it will just display "Starred".

defnax:
@csoler

PR #2949, opened 2025-06-01
.desktop file remove absolute path; fixes #2925

### LINKED ISSUE #2925: retroshare.desktop hardcodes install prefix
I have installed RetroShare to a custom prefix, but the installed desktop file assumes the retroshare binary is at `/usr/bin/retroshare` and similarly for the icon.

This is caused because the desktop file in the source code is not modified at build time to account for the install prefix.

JohannesKleine:


defnax:
@csoler

build_scripts/RedHat+Fedora/data/retroshare.desktop
```diff
@@ -3,8 +3,8 @@ Encoding=UTF-8
 Version=1.0
 Name=RetroShare
 Comment=Securely communicate with your friends
-Exec=/usr/bin/retroshare %U
-Icon=/usr/share/pixmaps/retroshare.xpm
```
csoler, line 7: Is this standard to not use an explicit path? It seems pretty risky to me: the executable that will be launched may not be the one you think.

JohannesKleine: It is good practice to use the absolute path, to be on the safe side, it is possible to use just the name.

PR #2956, opened 2025-06-17
Added for dark style different link color

defnax:


defnax:
this will not work for system dark theme on mac & linux
Else we can set by default light blue for links

PR #2957, opened 2025-06-18
Added alternative link color for retroshare links

defnax:
some places not implemented, its hardcoded there need to find a way for QTextBrowser & RSTextBrowser classes


defnax:
This my first version wasnt good

PR #2961, opened 2025-06-30
Added system dark mode for Windows 10

defnax:
require a restart or you need switch to other styles and back to default (built-in)

image

its for the default (built-in) 
image
image
image


JohannesKleine (commented):
It does Work on Windows 11, no negative side effects observed for Debian 12.

PR #2962, opened 2025-07-02
Added for Forums fonts settings

defnax:


JohannesKleine (approved):
Tested with Debian 12, zoom in and zoom out does work.

defnax:
Main improvement is store font size in Settings & Load in forum text

PR #2963, opened 2025-07-03
Fix to load the right logo without restart too

defnax:


JohannesKleine (approved):
Tested on Debian 12 with latest master and it does change the logo on the home screen according to different style sheets without restart.

retroshare-gui/src/gui/HomePage.cpp
```diff
@@ -382,6 +382,13 @@ void HomePage::openWebHelp()
     QDesktopServices::openUrl(QUrl(QString("https://retrosharedocs.readthedocs.io/en/latest/")));
 }
 
+void HomePage::showEvent(QShowEvent *event)
+{
+	if (!event->spontaneous()) {
+		updateHomeLogo();
+	}
```
csoler, line 389: don't you also need to call the parent class showEvent() ?

retroshare-gui/src/gui/HomePage.h
```diff
@@ -51,6 +51,8 @@ class HomePage : public MainPage
     void getOwnCert(QString& invite,QString& description) const;
     RetroshareInviteFlags currentInviteFlags() const ;
 
+    virtual void showEvent(QShowEvent *) ;
```
csoler: override is missing

retroshare-gui/src/gui/HomePage.cpp
defnax: no i doesnt need it works like this

JohannesKleine (approved):

retroshare-gui/src/gui/HomePage.cpp
JohannesKleine: In the code base it is usually called. So i can add ```QWidget::showEvent(event);``` to line 390 in another pull request just to be consistent with the rest of the code. It should work without it, but when qt changes something or by switching from qt5 to qt6 it might break without it. I tested it both.

PR #2965, opened 2025-07-05
Fixed the QComboBox issues for light and dark style

defnax:


JohannesKleine:
image
For testing it fixes this issue of the numbers are not properly shown for light and dark style.

JohannesKleine:
Tested on Debian 12, no negative side effects observed.

defnax:
@csoler check

retroshare-gui/src/gui/qss/stylesheet/Standard_Dark.qss
```diff
@@ -1243,6 +1243,7 @@ QComboBox {
   /* changed to 4px to fix #239 */
   /* Fixes #103, #111 */
   min-height: 1.5em;
+  min-width: 35px;
```
csoler: you should probably use em instead of px: em is the font size in pixels, so it is compatible with HDPI screens. px instead is an absolute value which will have different sizes depending on the Qt scale factor.

defnax:
done @csoler

PR #2968, opened 2025-07-07
Fix for Standard Dark

defnax:


csoler:
This works but i'm wondering if it wouldn't be better to have some generic method to get colors for dark vs. white styles.

defnax:
i think with qt 5.15 not easy on all platforms detect if light or dark mode
on windows 10/11 with qt 5.15.x we can use to detect dark/light mode and set useable color or for standard dark
```
QSettings settings("HKEY_CURRENT_USER\\Software\\Microsoft\\Windows\\CurrentVersion\\Themes\\Personalize",QSettings::NativeFormat);
if(settings.value("AppsUseLightTheme")==0){
    painter->setBrush(Qt::black);
}
```

but on qt 6.5/6.9 its get more easier detect dark/light mode
https://www.qt.io/blog/dark-mode-on-windows-11-with-qt-6.5

`QStyleHints has a new property, colorScheme, that holds a Qt::ColorScheme enum value, either Qt::ColorScheme::Light or Qt::ColorScheme::Dark (or Qt::ColorScheme::Unknown on systems where there is no such system color scheme). This allows applications to determine whether a hand-crafted application palette should be dark or light, respecting the user's system preference. And like all good properties, it comes with a change notification signal QStyleHints::colorSchemeChanged that allows applications to respond to changes in the system color scheme.`

JohannesKleine:
Tested on Debian 12, it works. Good to be merged. After RetroShare uses qt6 than it is time to fix it.

PR #2969, opened 2025-07-08
Added for login dialog switch background image

defnax:


retroshare-gui/src/gui/StartDialog.cpp
```diff
@@ -90,6 +91,9 @@ StartDialog::StartDialog(QWidget *parent)
 	}
 
 	ui.password_input->setFocus();
+
+	if (Settings->getSheetName() == ":Standard_Dark")
+		ui.rsTopLogo->setPixmap(FilesDefs::getPixmapFromQtResourcePath(":images/logo/logo_web_nobackground_black.png"));
```
JohannesKleine: recommend changes by adding the / after :, it prevents [QDir::addSearchPath](https://doc.qt.io/archives/qt-5.15/qdir.html#addSearchPath)() getting called, thince you are using the full path to the file anyways.
":images/logo/… to ":/images/logo/…

JohannesKleine (approved):
Tested on Debian 12.

defnax:
@csoler check

retroshare-gui/src/gui/StartDialog.cpp
```diff
@@ -90,6 +91,9 @@ StartDialog::StartDialog(QWidget *parent)
 	}
 
 	ui.password_input->setFocus();
+
+	if (Settings->getSheetName() == ":Standard_Dark")
+		ui.rsTopLogo->setPixmap(FilesDefs::getPixmapFromQtResourcePath(":/images/logo/logo_web_nobackground_black.png"));
```
csoler, line 94: Isn't this something you should better do in the qss file? It's weird to hack an image here depending on the specific style RS is using.

PR #2974, opened 2025-07-12
Fixed hardcoded background color

defnax:
* disabled hardcoded white background color

### DIFF
--- a/retroshare-gui/src/gui/settings/RSPermissionMatrixWidget.h
+++ b/retroshare-gui/src/gui/settings/RSPermissionMatrixWidget.h
@@ -34,7 +34,7 @@
 #define SCALE_WIDTH   75  /** Width of the scale */
 
 #define BACK_COLOR       Qt::white
-#define FOREGROUND_COLOR Qt::black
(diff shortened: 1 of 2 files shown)


defnax:
@csoler check

defnax:
@JohannesKleine

retroshare-gui/src/gui/settings/RSPermissionMatrixWidget.h
```diff
@@ -34,7 +34,7 @@
 #define SCALE_WIDTH   75  /** Width of the scale */
 
 #define BACK_COLOR       Qt::white
-#define FOREGROUND_COLOR Qt::black
+#define FOREGROUND_COLOR Qt::gray
```
JohannesKleine, line 37: I do not like it. With built-in in light mode and standard light, it is not easy readable and gives me eye cancer. Built-in in dark mode looks good, the same in standard dark.

JohannesKleine (changes requested):
A more consolidated approach, where those colors are changed in one place.

PR #2991, opened 2025-07-20
Replaced deprecated QString::sprintf by QString::asprintf

thunder2:


JohannesKleine:
Quick test on Debian 12 with Qt5 no negative side effects observed.

PR #2995, opened 2025-07-21
Fixed set of default language in LanguageSupport::localize. Use value…

thunder2:
… of language map instead of key.


JohannesKleine:
Tested on Debian 12 qt5, language switching still works.

PR #3002, opened 2025-07-23
Qt6

thunder2:

PR #3003, opened 2025-07-25
Added Qt6 workflow

defnax:
require #3002


defnax:
@csoler as next need this workflow

PR #3014, opened 2025-08-04
Macos compatibility

JohannesKleine:
requires https://github.com/RetroShare/libretroshare/pull/190
it includes https://github.com/RetroShare/RetroShare/pull/2936 and my improvements

### DIFF
--- a/retroshare.pri
+++ b/retroshare.pri
@@ -860,14 +860,14 @@ macx-* {
 	}
 	rs_macos14.5 {
 		message(***retroshare.pri: Set Target and SDK to MacOS 14.5 )
-		QMAKE_MACOSX_DEPLOYMENT_TARGET=14.5
+		QMAKE_MACOSX_DEPLOYMENT_TARGET=10.13
 		QMAKE_MAC_SDK = macosx10.13
 		QMAKE_CXXFLAGS += -Wno-nullability-completeness
 		QMAKE_CFLAGS += -Wno-nullability-completeness
 	}
 	rs_macos15.2 {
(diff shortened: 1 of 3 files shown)


defnax:
thank you 🚀

defnax:
This worked compile on macos & github actions?

defnax:
@csoler

csoler:
I'm not an expert on cmake. Anyone can validate this?

JohannesKleine:
@csoler i made "git squash" on pr #2936 and added my stuff, so the review of it should still be valid

defnax:
@G10h4ck

retroshare.pri
```diff
@@ -836,25 +804,46 @@ macx-* {
 	}
 	rs_macos10.14 {
 		message(***retroshare.pri: Set Target and SDK to MacOS 10.14 )
-		QMAKE_MACOSX_DEPLOYMENT_TARGET=10.14
+		QMAKE_MACOSX_DEPLOYMENT_TARGET=10.13
```
csoler, line 807: when rs_macos10.14 is asked, you should use 10.14. Why did you change all the versions (here and below??)

JohannesKleine:
QMAKE_MACOSX_DEPLOYMENT_TARGET is the minimum MacOS Version it can be compiled for, whereas QMAKE_MAC_SDK is the sdk version available on the MacOs. I made that change on purpose to be able to compile for older MacOS from any supported MacOs version. The effective minimum MacOS version will be 10.13 or higher depending on what the minimum MacOS version of the installed libraries is. It will be displayed in an info message while linking if it is not 10.13.

PR #3034, opened 2025-10-05
Update Linux_InstallGuide.md

sammy2k8-lab:
added "doxygen" as additional dependence because of compile error (doxygen not found) with clean ubuntu 24.04/25.04 install.


csoler:
This has been done in another PR apparently (as conflicts already report)

PR #3057, opened 2025-12-01
Fix compilation error 'QDateTime::fromTime_t' not found when building with Qt6

Passw:
QDateTime::fromTime_t
QDateTime::fromSecsSinceEpoch


thunder2:
@csoler There is duplicate source code in FriendsDialog::chatMessageReceived. Before and inside "if(ui.chatWidget->isActive())"
Inside the if statement is the correct usage of DateTime::DateTimeFromTime_t for Qt 6 instead of QDateTime::fromTime_t.

defnax:
@csoler fix qt6 compile

JohannesKleine:
Replaced by #3065

PR #3071, opened 2025-12-14
Fixed Layout issue

defnax:
* Fixed to have same Thumbnail dimensions


csoler:
I'm not sure that using explicit sizes like this will sort this problem for all systems. Let's see.

PR #3072, opened 2025-12-16
Remove recursive singleShot calls on feed items

jolavillette:
Remove recursive singleShot calls on feed items, that were creating thousands of timers and causing massive cpu overhead.
The functionality is replaced by a single, controlled member QTimer started in the constructor/setup method.

### DIFF
--- a/retroshare-gui/src/gui/feeds/ChatMsgItem.cpp
+++ b/retroshare-gui/src/gui/feeds/ChatMsgItem.cpp
@@ -67,6 +67,11 @@ ChatMsgItem::ChatMsgItem(FeedHolder *parent, uint32_t feedId, const RsPeerId &pe
     updateItemStatic();
     updateItem();
     insertChat(message);
+
+    m_updateTimer = new QTimer(this);
+    m_updateTimer->setSingleShot(false);
+    connect(m_updateTimer, SIGNAL(timeout()), this, SLOT(updateItem()));
+    m_updateTimer->start(1000);
 }
 
 void ChatMsgItem::updateItemStatic()
@@ -122,11 +127,6 @@ void ChatMsgItem::updateItem()
             msgButton->setEnabled(false);
         }
     }
-    
-    /* slow Tick  */
-    int msec_rate = 10129;
-
-    QTimer::singleShot( msec_rate, this, SLOT(updateItem( void ) ));
     return;
 }
 
--- a/retroshare-gui/src/gui/feeds/ChatMsgItem.cpp
+++ b/retroshare-gui/src/gui/feeds/ChatMsgItem.cpp
@@ -36,6 +36,9 @@
 
 #include "gui/msgs/MessageInterface.h"
 
+#include "util/qtthreadsutils.h"
+#include <retroshare/rsevents.h>
+
 /*****
  * #define DEBUG_ITEM 1
  ****/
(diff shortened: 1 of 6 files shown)


csoler:
These timers shouldn't exist at all: the security item should have its own event handler (see PeetItem.cpp) so as to capture whatever change is needed. In summary: modify your PR so that (1) the singleShot calls are just removed and no timer is created, and (2) add a event handler to capture the event that update the item is needed.

retroshare-gui/src/gui/feeds/ChatMsgItem.cpp
```diff
@@ -36,6 +36,9 @@
 
 #include "gui/msgs/MessageInterface.h"
 
+#include "util/qtthreadsutils.h"
+#include <retroshare/rsevents.h>
```
csoler: <> is for system includes, and "" for local includes. Not a problem, but just for you to know.

retroshare-gui/src/gui/feeds/SecurityIpItem.h
```diff
@@ -60,12 +63,13 @@ private slots:
 	void banIpListChanged(const QString &ipAddress);
 
 private:
-    RsFeedTypeFlags mType;
+	RsFeedTypeFlags mType;
```
csoler, line 65: is that used anywhere?

retroshare-gui/src/gui/feeds/SecurityIpItem.cpp
```diff
@@ -76,6 +83,17 @@ void SecurityIpItem::setup()
 
 	updateItemStatic();
 	updateItem();
+
+	mEventHandlerId = 0;
+
+	rsEvents->registerEventsHandler( [this](std::shared_ptr<const RsEvent> e)
+	{
+	        RsQThreadUtils::postToObject([=]()
```
csoler, line 91: Is there a way to somehow only call updateItem() when it is relevant? Not critical but that would reduce the load.

retroshare-gui/src/gui/feeds/ChatMsgItem.cpp
jolavillette: I will fix it and another one in SecurityIpitem.h

retroshare-gui/src/gui/feeds/SecurityIpItem.h
jolavillette: Apparently it is used to discriminate between banned IPs and wrong external IPs

retroshare-gui/src/gui/feeds/SecurityIpItem.cpp
jolavillette: like this ?

        rsEvents->registerEventsHandler( [this](std::shared_ptr<const RsEvent> e)
        {
                RsQThreadUtils::postToObject([=]()
                {
                        // Filter events to only update relevant items.
                        auto fe = dynamic_cast<const RsFriendListEvent*>(e.get());
                        if(fe && fe->mSslId != mSslId)
                                return;
                        UpdateItem();
                }
                , this );
        }, mEventHandlerId, RsEventType::FRIEND_LIST );

csoler: yes. Perfect.

retroshare-gui/src/gui/feeds/SecurityIpItem.cpp
```diff
@@ -76,6 +83,21 @@ void SecurityIpItem::setup()
...
+        	{
+			// Filter events to only update relevant items.
+			auto fe = dynamic_cast<const RsFriendListEvent*>(e.get());
+			if(fe && fe->mSslId != mSslId)
```
csoler: actually it should be if(!fe || fe->mSslId != mSslId) return; otherwise you have have fe=nullptr and not return ;-)

jolavillette: done

PR #3084, opened 2025-12-20
Replace RTT graph with a sortable table view: Name, RTT, ID, IP

jolavillette:
Replace RTT graph with a sortable table view: Name, RTT, ID, IP.

The current RTT statistics view is a moving graph, that can't be sorted, and that can't fit in the window when there are many friends.

It replaced it with a table that provide Peer Name, RTT (ms), Node ID, and IP Adress. The table is entirely sortable.

Note that this view also allows easy identification of a friend by their ID.

### DIFF
--- a/retroshare-gui/src/gui/statistics/RttStatistics.cpp
+++ b/retroshare-gui/src/gui/statistics/RttStatistics.cpp
@@ -18,124 +18,195 @@
  *                                                                             *
  *******************************************************************************/
 
+#include "RttStatistics.h"
 #include <iostream>
-#include <QTimer>
-#include <QObject>
-
-#include <QPainter>
-#include <QStylePainter>
-
+#include <limits>
+#include <QHeaderView>
+#include <QHostAddress>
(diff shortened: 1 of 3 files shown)


csoler:
The way this is implemented (using explicit item insertion in a table) may get super costly when you have many items. Using an abstract item model would be much more efficient. This is however much more difficult to implement.

About the idea of showing RTT as numbers vs. curves, this has pros and cons. Curves allow you to see time-related changes and globally sort RTT over a longer time period, which can be sometimes useful. Numbers are more accurate and as you say, more easy to sort. It would be probably good to have this table as an option.

jolavillette:
See new PR 3087
https://github.com/RetroShare/RetroShare/pull/3087
Graph view restored, we now have 2 tabs
Complexity now O(N) as we use QHash lookup for the table view
Closing pr/3084

jolavillette:
Closing pr/3084 in favor of pr/3087
https://github.com/RetroShare/RetroShare/pull/3087

PR #3087, opened 2025-12-21
Add table view in rtt statistics

jolavillette:
Add a sortable table view: Name, RTT, ID, and IP

The current RTT statistics view is a moving graph, that can't be sorted, and that can't fit in the window when there are many friends

This PR add a sortable table that provide Peer Name, RTT (ms), Node ID, and IP Address

Note that this view also allows easy identification of a friend by their ID.

### DIFF
--- a/retroshare-gui/src/gui/statistics/RttStatistics.cpp
+++ b/retroshare-gui/src/gui/statistics/RttStatistics.cpp
@@ -136,8 +136,8 @@ void RttStatistics::processSettings(bool bLoad)
 // --- Table Update Logic (O(N) Optimized) ---
 void RttStatistics::updateRttValues()
 {
-    // Optimize: Only update if Table tab is visible (Index 1)
-    if (tabWidget->currentIndex() != 1) return;
+    // Only update if the rtt tab is visible and the table view is selected
+    if (!isVisible() || tabWidget->currentIndex() != 1) return;
 
     std::list<RsPeerId> idList;
     if (!rsPeers) return;
--- a/retroshare-gui/src/gui/statistics/RttStatistics.cpp
+++ b/retroshare-gui/src/gui/statistics/RttStatistics.cpp
@@ -18,123 +18,194 @@
  *                                                                             *
  *******************************************************************************/
 
+#include "RttStatistics.h"
 #include <iostream>
-#include <QTimer>
-#include <QObject>
-
-#include <QPainter>
-#include <QStylePainter>
-
+#include <limits>
+#include <QHeaderView>
+#include <QHostAddress>
 #include <retroshare/rsrtt.h>
 #include <retroshare/rspeers.h>
-#include "RttStatistics.h"
-#include "time.h"
-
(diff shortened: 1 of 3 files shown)


retroshare-gui/src/gui/statistics/RttStatistics.cpp
```diff
@@ -66,23 +118,75 @@ void RttStatistics::processSettings(bool bLoad)
...
-    
+
     m_bProcessSettings = false;
+}
 
+// --- Table Update Logic (O(N) Optimized) ---
+void RttStatistics::updateRttValues()
+{
+    // Optimize: Only update if Table tab is visible (Index 1)
+    if (tabWidget->currentIndex() != 1) return;
```
csoler: I would prefer if(!isVisible()) return; since the tab may be chosen but you're actually in a different part of the UI.

jolavillette: done

JohannesKleine (approved):
Tested on Debian 13, it is a nice table to debug bad connections or find peers by hash.

retroshare-gui/src/gui/statistics/RttStatistics.cpp
```diff
@@ -32,28 +38,74 @@
...
+// --- Custom Item for Sorting (Table) ---
+class RttTreeItem : public QTreeWidgetItem {
+public:
+    using QTreeWidgetItem::QTreeWidgetItem;
+
+    bool operator<(const QTreeWidgetItem &other) const {
+        int sortCol = treeWidget() ? treeWidget()->sortColumn() : 0;
+
+        // RTT Sort
+        if (sortCol == 1) {
```
csoler: could you plz use names here? Can be an enum, or a #define, so that we know what column corresponds to each case.

retroshare-gui/src/gui/statistics/RttStatistics.cpp
```diff
@@ -32,28 +38,74 @@
...
+    m_timer->start(1000);
+
 	// load settings
     processSettings(true);
 }
 
 RttStatistics::~RttStatistics()
 {
-
+    if(m_timer) m_timer->stop();
```
csoler: don't forget to also delete the timer.

retroshare-gui/src/gui/statistics/RttStatistics.cpp
```diff
@@ -66,23 +118,75 @@ void RttStatistics::processSettings(bool bLoad)
...
+        if (!item) {
+            item = new RttTreeItem(treeWidget);
+            item->setData(0, Qt::UserRole, qPeerId);
+        }
+
+        item->setText(0, peerName);
+        if (rttInMs != -1) item->setText(1, QString::number(rttInMs));
+        else item->setText(1, "?");
+        item->setText(2, qPeerId);
+        item->setText(3, ipAddress);
```
csoler: You may have forgotten to also remove (and delete) the items for peers that are not online anymore, or removed.

PR #3090, opened 2025-12-22
Splitting rsMsgs into rsMail and rsChats

csoler:


csoler:
To be tested. No changes expected. It's just the 1st step in a long road of cleaning up the chat system.

PR #3093, opened 2025-12-24
[WIP] Improve upload statistics

jolavillette:
attn pr/3093 now requires libretroshare pr/262

improve upload statistics

code by Claude and Gemini

### DIFF
--- a/retroshare-gui/src/gui/FileTransfer/SharedFilesDialog.cpp
+++ b/retroshare-gui/src/gui/FileTransfer/SharedFilesDialog.cpp
@@ -145,8 +145,7 @@ public:
     void setUploadedOnly(bool val) {
         if (m_uploadedOnly != val) {
              m_uploadedOnly = val;
-             // invalidateFilter(); // CRASH FIX: Do not invalidate incrementally. 
-                                  // The dialog calls model->update() which triggers a full Reset.
+             invalidateFilter(); 
         }
     }
 
@@ -229,7 +228,7 @@ SharedFilesDialog::~SharedFilesDialog()
  * Constructor for the base SharedFilesDialog.
  */
 SharedFilesDialog::SharedFilesDialog(bool remote_mode, QWidget *parent)
-  : RsAutoUpdatePage(1000,parent), model(NULL), uploadedOnly_CB(NULL)
+  : RsAutoUpdatePage(1000,parent), model(NULL)
 {
     /* Invoke the Qt Designer generated object setup routine */
     ui.setupUi(this);
@@ -350,17 +349,7 @@ SharedFilesDialog::SharedFilesDialog(bool remote_mode, QWidget *parent)
 LocalSharedFilesDialog::LocalSharedFilesDialog(QWidget *parent)
     : SharedFilesDialog(false,parent)
 {
-    // Label updated to "Popular files"
-    uploadedOnly_CB = new QCheckBox(tr("Popular files"), this);
-    
-    // Updated tooltip to be more descriptive of the "Popularity" (upload activity)
-    uploadedOnly_CB->setToolTip(tr("Show only files and folders that have been successfully uploaded to other peers"));
-    
-    // Positioning the checkbox next to the view selector
-    int cbIndex = ui.horizontalLayout_2->indexOf(ui.viewType_CB);
-    ui.horizontalLayout_2->insertWidget(cbIndex + 1, uploadedOnly_CB);
-    
-    connect(uploadedOnly_CB, SIGNAL(toggled(bool)), this, SLOT(filterUploadedOnlyToggled(bool)));
+    connect(ui.uploadedOnly_CB, SIGNAL(toggled(bool)), this, SLOT(filterUploadedOnlyToggled(bool)));
 
     // Ensure proper columns are visible for local sharing
     ui.dirTreeView->setColumnHidden(SHARED_FILES_DIALOG_COLUMN_WN_VISU_DIR, false) ;
@@ -400,6 +389,7 @@ RemoteSharedFilesDialog::RemoteSharedFilesDialog(QWidget *parent)
     //
     changeCurrentViewModel(ui.viewType_CB->currentIndex()) ;
 
+    ui.uploadedOnly_CB->hide();
     ui.addShares_PB->hide() ;
 }
 
@@ -524,9 +514,9 @@ void SharedFilesDialog::changeCurrentViewModel(int viewTypeIndex)
         proxyModel = flat_proxyModel ;
     }
 
-    if(uploadedOnly_CB) {
-        if(tree_proxyModel) tree_proxyModel->setUploadedOnly(uploadedOnly_CB->isChecked());
-        if(flat_proxyModel) flat_proxyModel->setUploadedOnly(uploadedOnly_CB->isChecked());
+    if(ui.uploadedOnly_CB) {
+        if(tree_proxyModel) tree_proxyModel->setUploadedOnly(ui.uploadedOnly_CB->isChecked());
+        if(flat_proxyModel) flat_proxyModel->setUploadedOnly(ui.uploadedOnly_CB->isChecked());
     }
 
     showProperColumns() ;
@@ -561,7 +551,7 @@ void SharedFilesDialog::changeCurrentViewModel(int viewTypeIndex)
     FilterItems();
 
     // MODIFICATION: Expand tree if "Uploaded Only" is active, otherwise items remain hidden in collapsed folders.
-    if(viewTypeIndex==VIEW_TYPE_TREE && uploadedOnly_CB && uploadedOnly_CB->isChecked()) {
+    if(viewTypeIndex==VIEW_TYPE_TREE && ui.uploadedOnly_CB && ui.uploadedOnly_CB->isChecked()) {
         expandAll();
     }
 }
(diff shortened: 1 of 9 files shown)


retroshare-gui/src/gui/FileTransfer/SharedFilesDialog.cpp
```diff
@@ -292,17 +343,31 @@ SharedFilesDialog::SharedFilesDialog(bool remote_mode, QWidget *parent)
...
+/**
+ * Constructor for LocalSharedFilesDialog.
+ * Sets the new "Popular files" label and a more descriptive tooltip.
+ */
 LocalSharedFilesDialog::LocalSharedFilesDialog(QWidget *parent)
     : SharedFilesDialog(false,parent)
 {
-    // Hide columns after loading the settings
+    // Label updated to "Popular files"
+    uploadedOnly_CB = new QCheckBox(tr("Popular files"), this);
```
csoler: This is surprising that you manually set that widget. It should be in the .ui file (possibly with that same name) so that you do not have to deal with layouts here. Also if someone modifies the ui file the layouts names may change which will break your code.

csoler: also if the CB is in the ui file you need to hide/show it depending on the current class.

retroshare-gui/src/gui/RemoteDirModel.cpp
```diff
@@ -148,8 +148,113 @@ void RetroshareDirModel::treeStyle()
...
+    if(RemoteMode)
+        return;
+
+    // Start recursion from Root (NULL)
+    collectStatsRecursive(NULL);
+}
+
+TreeStyle_RDM::FolderStats TreeStyle_RDM::collectStatsRecursive(void* ref)
+{
+    FolderStats stats;
```
csoler, line 167: add:
- a constructor to FolderStats that inits the values to 0. This will fix the uninitialized memory read when ref points to a directory with only subdirectories (see the code below)
- an operator+=(const FolderStats& s), so that the code below simplifies (stats += childstats)

retroshare-gui/src/gui/RemoteDirModel.cpp
```diff
@@ -148,8 +148,113 @@ void RetroshareDirModel::treeStyle()
...
+    if(details.type == DIR_TYPE_FILE || details.type == DIR_TYPE_EXTRA_FILE)
+    {
+        stats.size = details.size;
+        stats.count = 1;
+        stats.uploads = rsFiles->getCumulativeUpload(details.hash);
+
+        // Special handling for Extra Files (virtual root)
+        if (details.type == DIR_TYPE_EXTRA_FILE) 
+        {
+            m_folderUploadTotals["!!RS_EXTRA_FILES_ROOT!!"] += stats.uploads;
```
csoler: It seems to me that these 3 maps actually contain each member of FolderStats. It would be better (and faster, since it's 1 search instead of 3) to use a std::map< , FolderStats> in the first place.

retroshare-gui/src/gui/RemoteDirModel.cpp
```diff
@@ -148,8 +148,113 @@ void RetroshareDirModel::treeStyle()
...
+    }
+
+    return true; 
+}
+
 void TreeStyle_RDM::update()
 {
+
+    // Recalculate totals before notifying view update
+    recalculateDirectoryTotals();
```
csoler, line 248: I am not 100% sure, but shouldn't preMods() be called before doing anything here?

retroshare-gui/src/gui/RemoteDirModel.cpp
```diff
@@ -312,15 +419,10 @@ int FlatStyle_RDM::columnCount(const QModelIndex &/*parent*/) const
...
+QString RetroshareDirModel::getFlagsString(FileStorageFlags /*flags*/)
 {
-	char str[11] = "-  -  -" ;
-
-    if(flags & DIR_FLAGS_BROWSABLE) 	     str[0] = 'B' ;
-    if(flags & DIR_FLAGS_ANONYMOUS_SEARCH) 	 str[3] = 'S' ;
-	if(flags & DIR_FLAGS_ANONYMOUS_DOWNLOAD) str[6] = 'N' ;
-
-	return QString(str) ;
+	return QString("") ;
```
csoler, line 416: Is the  ItemDelegate handling this now?

retroshare-gui/src/gui/RemoteDirModel.cpp
```diff
@@ -675,8 +633,36 @@ QString FlatStyle_RDM::computeDirectoryPath(const DirDetails& details) const
...
+    if (!index.isValid())
+        return QVariant();
+
+    if (role == Qt::DisplayRole)
+    {
+        void *ref = index.internalPointer();
+        
+        // Try Lock-less read first? No, use lock for safety matching updateRefs
+        RS_STACK_MUTEX(_ref_mutex);
+        if (m_cache.contains(ref))
```
csoler: you can reduce the cost by a factor of 2 by doing:
auto it = m_cache.find(ref);

if(it != null)
    do_something(it->second)

since in your code both contains() and m_cache[ref] do a log search.

retroshare-gui/src/gui/RemoteDirModel.cpp
```diff
@@ -1219,8 +1212,9 @@ void FlatStyle_RDM::postMods()
 {
     time_t now = time(NULL);
 
-    if(_last_update + FLAT_VIEW_MIN_DELAY_BETWEEN_UPDATES > now)
-        return ;
+    // MODIFICATION: Removed throttling because it caused stuck model state (beginReset without endReset)
```
csoler: well that's 2 different problems. The throttling was needed here becasue you may receive may updates from multiple friends and re-updating the model everytime is very costly.

retroshare-gui/src/gui/RemoteDirModel.cpp
```diff
@@ -1524,71 +1518,213 @@ void RetroshareDirModel::getFilePaths(const QModelIndexList &list, std::list<std
...
+        return ;
+    }
 
-	if(result_list.empty())	// in this case we dont clear the list of filtered items, so that we can keep the old filter list
-		return ;
+    mFilteredPointers.clear();
 
-	mFilteredPointers.clear();
+    // Mark matching pointers and their hierarchy as visible
+    for(auto it(result_list.begin()); it != result_list.end(); ++it)
```
csoler, line 1594: I see no mutex protection here, and it seems that result_list may contain refs that do not exist anymore in a situation. This is fine if that class is only called from the GUI thread. I think it's the case.

retroshare-gui/src/gui/RemoteDirModel.h
```diff
@@ -203,12 +210,34 @@ class TreeStyle_RDM: public RetroshareDirModel
...
 		void showEmpty(const bool value);
 
 	private:
 		QAction *_showEmptyAct;
 		bool _showEmpty;
+
+        // Helper to calculate total recursive statistics per directory (Files count, Size, Uploads)
+        void recalculateDirectoryTotals();
+        
+        QHash<QString, uint64_t> m_folderUploadTotals;
```
csoler: use a single std::map<QString,FolderStats>

csoler (commented):
Very long PR! (github has issues displaying it).

jolavillette:
# Shared Files Improvements & Refactoring

This update addresses developer feedback, optimizes performance for large file lists, and fixes search responsiveness issues in Flat View.

### 1. Developer Feedback Compliance
*   **UI Integration**: Moved its manual creation of `uploadedOnly_CB` (Popular Files) into the `.ui` file and managed visibility dynamically.
*   **Data Structures**: Added a proper constructor and `operator+=` to [FolderStats] for cleaner recursive aggregation.
*   **Map Consolidation**: Replaced three separate statistical maps with a single `std::map<QString, FolderStats> m_folderTotals` to reduce lookup overhead.
*   **Flags Display**: Optimized [getFlagsString] to return an empty string, letting the [ItemDelegate] handle the graphical icon display.

### 2. Performance Optimizations
*   **Caching**: Optimized cache access in `FlatStyle_RDM::data` by using `m_cache.find(ref)` to avoid redundant double-lookups.
*   **Update Flow**: Ensured [preMods()] is correctly called before calculating directory totals.
*   **Batching**: Tuned batch loading speeds for the Flat View to provide a smoother UI experience.

### 3. Search & Responsiveness Fixes
*   **Throttle Bypass**: Modified the 120-second update throttle to be bypassed during search operations and filter clearing, ensuring immediate UI feedback.
*   **Update Guard Fix**: Resolved an issue where rapid typing in the search box would cause updates to be ignored if a previous background crawl was still active.
*   **Hierarchical Visibility**: Improved the search engine climbing logic to ensure shared folder roots (DIR_TYPE_ROOT) are correctly marked as visible in Tree View search results.

retroshare-gui/src/gui/FileTransfer/SharedFilesDialog.cpp
```diff
@@ -52,9 +52,36 @@
 #include <QString>
 #include <QStyledItemDelegate>
 #include <QTreeView>
+#include <QCheckBox>
+#include <QHBoxLayout>
 
 #include <set>
 
+/**
+ * Helper class to block signals and show wait cursor during UI updates.
```
csoler: I think you can move that to utils/misc.h (where is the whileBlocking() function) since it can be used for many other contexts.

retroshare-gui/src/gui/RemoteDirModel.h
```diff
@@ -253,6 +294,19 @@ class FlatStyle_RDM: public RetroshareDirModel
...
+            QString name;
+            QString sizeStr;
+            QString ageStr;
+            QString uploadStr;
+            uint64_t size;
+            uint32_t mtime;
+            uint64_t uploads;
+            bool hasUploads;
+        };
+        QHash<void*, CachedFileDetails> m_cache;
```
csoler: This looks dangerous: nothing guaranties that two different void* will not have the same hash, unless you provide a specific function for this such as in 

QHash<void *, QString> hash(0, [](void*p)->uint64_t { return (uint64_t)p ; });

(I'm not sure that actually compiles). If not, try a std::map instead.

csoler: Also, during the life of a TreeModel it's very likely that the same ref will be used multiple times and possibly for different content. How can you ensure that the cache item is invalidated then?

retroshare-gui/src/gui/RemoteDirModel.cpp
csoler: Actually it is since there are small images. It's been ignored for a while.

retroshare-gui/src/gui/RemoteDirModel.h
csoler: also I don't see the advantage of the m_cache structure (as a cache) since requestDirDetails(void *) doesn't cost much as compared to how complicated it will be to keep the m_cache consistently updated.

PR #3097, opened 2025-12-28
Improved File transfer Tunnels display

defnax:


csoler:
For a widget that needs updating every second, it's much more preferable to use an abstract item model. In the PR you constantly add/remove items from a long list, which is going to cost a lot. The same should be done for GXS sync tunnels.

JohannesKleine:
Works on Debian 13.

defnax:
New review? @csoler

retroshare-gui/src/gui/statistics/GxsNetTunnelsDialog.cpp
```diff
@@ -0,0 +1,376 @@
...
+class NetTunnelsListDelegate: public QAbstractItemDelegate
+{
+public:
+    NetTunnelsListDelegate(QObject *parent=0);
+    virtual ~NetTunnelsListDelegate();
+    void paint(QPainter * painter, const QStyleOptionViewItem & option, const QModelIndex & index) const;
+    QSize sizeHint(const QStyleOptionViewItem &, const QModelIndex &) const;
+};
+
+NetTunnelsListDelegate::NetTunnelsListDelegate(QObject *parent) : QAbstractItemDelegate(parent)
```
csoler, line 59: maybe you can move that into the class since we're already in the .cpp.

retroshare-gui/src/gui/statistics/GxsNetTunnelsDialog.cpp
```diff
@@ -0,0 +1,376 @@
...
+	processSettings(true);
+}
+
+GxsNetTunnelsDialog::~GxsNetTunnelsDialog()
+{
+    // save settings
+    processSettings(false);
+}
+
+void GxsNetTunnelsDialog::processSettings(bool bLoad)
```
csoler, line 168: this function is doing nothing. Is that intended?

retroshare-gui/src/gui/statistics/GxsNetTunnelsDialog.h
```diff
@@ -0,0 +1,63 @@
...
+
+#include <retroshare/rsturtle.h>
+#include <retroshare/rstypes.h>
+
+#include <retroshare-gui/RsAutoUpdatePage.h>
+
+#include "ui_GxsNetTunnelsDialog.h"
+
+// Defines for download list list columns
+#define COLUMN_SERVICE           0
```
csoler: if possible, put the defines in the .cpp (I dont' think they are used in the .h) because it may cause conflicts when including this .h from elsewhere.

PR #3098, opened 2025-12-30
Fix Windows freeze in RsCollectionDialog Download mode

jolavillette:
Fix Windows freeze in RsCollectionDialog Download mode by making the local file system model initialization conditional

### DIFF
--- a/retroshare-gui/src/gui/common/RsCollectionDialog.cpp
+++ b/retroshare-gui/src/gui/common/RsCollectionDialog.cpp
@@ -1,21 +1,21 @@
 /*******************************************************************************
  * retroshare-gui/src/gui/common/RsCollectionDialog.cpp                        *
- *                                                                             *
+ * *
  * Copyright (C) 2011, Retroshare Team <retroshare.project@gmail.com>          *
- *                                                                             *
+ * *
  * This program is free software: you can redistribute it and/or modify        *
  * it under the terms of the GNU Affero General Public License as              *
  * published by the Free Software Foundation, either version 3 of the          *
  * License, or (at your option) any later version.                             *
- *                                                                             *
+ * *
  * This program is distributed in the hope that it will be useful,             *
  * but WITHOUT ANY WARRANTY; without even the implied warranty of              *
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                *
  * GNU Affero General Public License for more details.                         *
- *                                                                             *
+ * *
  * You should have received a copy of the GNU Affero General Public License    *
  * along with this program. If not, see <https://www.gnu.org/licenses/>.       *
- *                                                                             *
+ * *
  *******************************************************************************/
 
 #include "gui/common/FilesDefs.h"
@@ -36,6 +36,7 @@
 #include <QMenu>
 #include <QTextEdit>
 #include <QTreeView>
+#include <QFileDialog>
 
 #define COLUMN_FILE     0
 #define COLUMN_FILEPATH 1
@@ -125,7 +126,7 @@ protected:
  * @param readOnly: Open dialog for RsColl as ReadOnly
  */
 RsCollectionDialog::RsCollectionDialog(const QString& collectionFileName, RsCollectionDialogMode mode)
-  : _mode(mode)
+  : _mode(mode), _dirModel(0), _tree_proxyModel(0), _selectionProxy(0)
 {
     RsCollection::RsCollectionErrorCode err_code;
     mCollection = new RsCollection(collectionFileName,err_code);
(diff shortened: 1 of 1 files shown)


jolavillette:
Whats Gemini says about the fix:

Identified the Root Cause: Found that initializing QFileSystemModel with setRootPath("/") causes a blocking enumeration of all drives (network, local, and removable) on Windows, leading to the UI freeze.

Implemented Conditional Initialization: Modified the init() function to only instantiate the file system model when the dialog is in EDIT mode.

Optimized Download Mode: In DOWNLOAD mode, the disk scan is now entirely skipped, making the dialog open instantly without any interface hang.

Ensured Stability: Explicitly set file system pointers to 0 (null) in Download mode to prevent any unnecessary background disk activity or potential crashes.

retroshare-gui/src/gui/common/RsCollectionDialog.cpp
```diff
@@ -1,21 +1,21 @@
 /*******************************************************************************
  * retroshare-gui/src/gui/common/RsCollectionDialog.cpp                        *
- *                                                                             *
+ * *
```
csoler: something wrong happened here.

retroshare-gui/src/gui/common/RsCollectionDialog.cpp
```diff
@@ -125,7 +126,7 @@ class FSMSortFilterProxyModel : public QSortFilterProxyModel
  * @param readOnly: Open dialog for RsColl as ReadOnly
  */
 RsCollectionDialog::RsCollectionDialog(const QString& collectionFileName, RsCollectionDialogMode mode)
-  : _mode(mode)
+  : _mode(mode), _dirModel(0), _tree_proxyModel(0), _selectionProxy(0)
```
csoler: please use nullptr instead of 0. The value is the same but the type is not.

retroshare-gui/src/gui/common/RsCollectionDialog.cpp
```diff
@@ -209,21 +210,32 @@ void RsCollectionDialog::init(const QString& collectionFileName)
...
-	_dirModel = new QFileSystemModel(this);
-	_dirModel->setRootPath("/");
-	_dirModel->setFilter(QDir::AllEntries | QDir::NoSymLinks | QDir::NoDotAndDotDot);
-	_dirLoaded = false;
-	connect(_dirModel, SIGNAL(directoryLoaded(QString)), this, SLOT(directoryLoaded(QString)));
+	// 3 Initialize List ONLY in EDIT mode to avoid Windows background scanning freezes.
+	if (_mode == EDIT)
+	{
+		_dirModel = new QFileSystemModel(this);
+		_dirModel->setRootPath("/");
```
csoler: is there anything like a "home dir" on windows? That would probably help a lot to start from that dir instead of /. Obviously you should make sure that this choice doesn't prevent the user to reach other drives/partitions.

retroshare-gui/src/gui/common/RsCollectionDialog.cpp
```diff
@@ -838,12 +851,12 @@ bool RsCollectionDialog::openNewCollection(const RsFileTree& tree)
...
         mb.setInformativeText(tr("File already exists.")+"\n"+tr("What do you want to do?"));
-        QAbstractButton *btnOwerWrite = mb.addButton(tr("Overwrite"), QMessageBox::YesRole);
-        QAbstractButton *btnCancel = mb.addButton(tr("Cancel"), QMessageBox::ResetRole);
+        mb.addButton(tr("Overwrite"), QMessageBox::YesRole);
+        mb.addButton(tr("Cancel"), QMessageBox::ResetRole);
         mb.setIcon(QMessageBox::Question);
         mb.exec();
 
-        if (mb.clickedButton()==btnCancel)
+        if (mb.clickedButton()==(QAbstractButton*)mb.buttons().at(1)) // btnCancel
```
csoler: not so nice, but still works. I suppose that clickedButton() will not return the pointer to the button anyway.

retroshare-gui/src/gui/common/RsCollectionDialog.cpp
jolavillette: fixed

retroshare-gui/src/gui/common/RsCollectionDialog.cpp
jolavillette: fixed

retroshare-gui/src/gui/common/RsCollectionDialog.cpp
jolavillette: Implemented QDir::homePath() instead of /. This successfully prevents the root scan freeze on Windows while still allowing users to navigate to other drives if needed.

retroshare-gui/src/gui/common/RsCollectionDialog.cpp
jolavillette: Refactored. Cleaned up the button handling logic to avoid index-based access (at(1)). This also resolves the compiler warnings regarding unused variables.

PR #3099, opened 2025-12-31
Close the statistics window when the Escape key is pressed

jolavillette:
StatisticsWindow.cpp: override the keyPressEvent function, enabling the statistics window to close when the Escape key is pressed


jolavillette:
Wrong use of git

PR #3100, opened 2025-12-31
Close statistics window with Escape key

jolavillette:
StatisticsWindow: override the keyPressEvent function, enabling the statistics window to close when the Escape key is pressed

Code and comment by Gemini

### DIFF
--- a/retroshare-gui/src/gui/statistics/GxsTransportStatistics.cpp
+++ b/retroshare-gui/src/gui/statistics/GxsTransportStatistics.cpp
@@ -33,6 +33,7 @@
 #include <QTimer>
 #include <QTreeWidget>
 #include <QWheelEvent>
+#include <QPointer> // Required for thread safety check
 
 #include <retroshare/rsgxstrans.h>
 #include <retroshare/rspeers.h>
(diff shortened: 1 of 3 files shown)


csoler:
The statistics window is a static object. It is shown by StatisticsWindow::showYourself();
Therefore it seems to me that it should never have the flag DeleteOnClose and that the Esc key should only call close(). This way the updates will not cause any problem.
FYI this window is a static object because it constructs temporal curves that you may not want to reset everytime you close the window.

jolavillette:
Gemini's comments on last commit

Following Cyril's feedback and the GDB crash analysis, here is a summary of the changes implemented:

Persistence: I disabled the WA_DeleteOnClose attribute in the StatisticsWindow constructor to ensure the window remains a persistent static object, preserving temporal curves and data.

Safe Escape Handling: I implemented a keyPressEvent to map the Escape key to close(), which now safely hides the window without destroying the instance.

Async Protection: I updated GxsTransportStatistics::loadGroups() to use a QPointer guard, ensuring that background threads do not attempt to update a null or invalid UI object, which was the root cause of the SIGSEGV.

retroshare-gui/src/gui/statistics/GxsTransportStatistics.cpp
```diff
@@ -452,42 +453,55 @@ void GxsTransportStatistics::loadGroupStats(const RsGxsGroupId& groupId)
 }
 #endif
 
-
 void GxsTransportStatistics::loadGroups()
 {
 	mStateHelper->setLoading(GXSTRANS_GROUP_META, true);
 
-	RsThread::async([this]()
+	// Use a QPointer to safely track 'this'. Even if the window is static,
```
csoler: I don't think that this is needed anymore, now that the window is not actually deleted.

retroshare-gui/src/gui/statistics/GxsTransportStatistics.cpp
```diff
@@ -452,42 +453,55 @@ void GxsTransportStatistics::loadGroupStats(const RsGxsGroupId& groupId)
...
 		if(!rsGxsTrans->getGroupStatistics(*stats))
 		{
 			RS_ERR("Cannot retrieve group statistics in GxsTransportStatistics");
-            delete stats;
+			delete stats;
 			return;
 		}
 
-        RsQThreadUtils::postToObject( [stats, this]()
+		if (self)
```
csoler: same here

retroshare-gui/src/gui/statistics/GxsTransportStatistics.cpp
jolavillette: Gemini says :

Hi Cyril, thanks for the review. I have updated the PR following your suggestions:

Simplified Async Logic: I removed the QPointer and the redundant if (self) checks in GxsTransportStatistics.cpp.

Window Persistence: Since I explicitly set WA_DeleteOnClose to false in the StatisticsWindow constructor, the instance is guaranteed to remain valid during background tasks, making the direct use of this safe.

Restoration Fix: I also kept the improvements to showYourself() (showNormal() and raise()) to ensure the window correctly reappears if it was minimized or hidden.

Ready for another look!

PR #3108, opened 2026-01-06
display libretroshare version in About dialogs and enable dirty flag support

jolavillette:
display libretroshare version in About dialogs and enable dirty flag support

requires libretroshare pr/229


JohannesKleine (approved):
Tested on Debian 13. To my understanding of c++ it looks fine.

PR #3111, opened 2026-01-09
add date format selection (System, ISO 8601, Text) in user settings, and standardize date and time formatting across the UI

jolavillette:
gui: add option to customize date format in Appearance settings

This commit introduces the UI selection and backend storage for the Date Format preference (System, ISO 8601, Text).

Note: This change only handles the configuration part. The actual date display throughout the application (Chat, Forums, etc.) does not yet use this setting and will be updated in separate PRs

code by Gemini

### DIFF
--- a/retroshare-gui/src/gui/common/FriendListModel.cpp
+++ b/retroshare-gui/src/gui/common/FriendListModel.cpp
@@ -748,8 +748,9 @@ QVariant RsFriendListModel::displayRole(const EntryIndex& e, int col) const
                                 }
                         }
 
-                        if(col == COLUMN_THREAD_LAST_CONTACT) return QVariant(most_recent_time);
-                        if(col == COLUMN_THREAD_IP)           return QVariant(most_recent_ip);
+			// Use the standardized date format from user preferences
+			if(col == COLUMN_THREAD_LAST_CONTACT) return DateTime::formatDateTime(most_recent_time);
+			if(col == COLUMN_THREAD_IP)           return QVariant(most_recent_ip);
 
                         return QVariant();
                 }
@@ -790,7 +791,8 @@ QVariant RsFriendListModel::displayRole(const EntryIndex& e, int col) const
 						return QVariant(QString::fromUtf8(node->node_info.location.c_str()));
 				}
 
-				case COLUMN_THREAD_LAST_CONTACT:   return QVariant(DateTime::DateTimeFromTime_t(node->node_info.lastConnect).toString());
+				// Force usage of the global DateTime formatting utility
+				case COLUMN_THREAD_LAST_CONTACT:   return DateTime::formatDateTime(node->node_info.lastConnect);
 				case COLUMN_THREAD_IP:             return QVariant(  (node->node_info.state & RS_PEER_STATE_CONNECTED) ? StatusDefs::connectStateIpString(node->node_info) : QString("---"));
 				case COLUMN_THREAD_ID:             return QVariant(  QString::fromStdString(node->node_info.id.toStdString()) );
 
--- a/retroshare-gui/src/gui/common/FriendListModel.cpp
+++ b/retroshare-gui/src/gui/common/FriendListModel.cpp
@@ -748,7 +748,7 @@ QVariant RsFriendListModel::displayRole(const EntryIndex& e, int col) const
                                 }
                         }
 
-			if(col == COLUMN_THREAD_LAST_CONTACT) return QVariant((qulonglong)DateTime::DateTimeToTime_t(most_recent_time));
+			if(col == COLUMN_THREAD_LAST_CONTACT) return QVariant((qulonglong)most_recent_time.toTime_t());
 			if(col == COLUMN_THREAD_IP)           return QVariant(most_recent_ip);
 
                         return QVariant();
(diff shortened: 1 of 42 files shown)


csoler:
As far as I understand, the date format is provided by Qt. Doesn't the Qt date format follow the system settings already?

jolavillette:
RetroShare currently uses several date formats everywhere in the UI. In the Network view alone there are 2. This PR aims at unifying the date formats according to the user's choice.

csoler:
ok! So there's two different problems here: (1) dates not being consistent throughout the software (very bad) and (2) tweaking the date according to user.

jolavillette:
Indeed, and this pr provides the bricks to fix both

jolavillette:
1. add option to customize date format in Appearance settings
Introduces the UI selection and backend storage for the Date Format preference (System, ISO 8601, Text)

2. util/DateTime.cpp
/**
 * This utility class provides standardized date and time formatting across the application.
 * - formatDate() and formatTime(): These methods dynamically adjust the output
 * based on the user's selected preference in the application settings
 * (System = Qt Short Format, ISO 8601, Text = Qt Long Format).
 * - formatLongDate() and formatLongDateTime(): use standard Qt Long Format
 */





most code by Gemini

retroshare-gui/src/util/DateTime.cpp
```diff
@@ -23,63 +23,102 @@
...
-	return formatLongDate(DateTimeFromTime_t(dateValue).date());
+	return formatDate(DateTimeFromTime_t(dateValue).date());
 }
 
-QString DateTime::formatLongDate(const QDate &dateValue)
+QString DateTime::formatDate(const QDate &dateValue)
 {
-	QString customDateFormat = RsApplication::customDateFormat();
+	/* Retrieve the date format index from global settings */
+	int dateFormat = Settings->getDateFormat();
```
csoler: This method will be called many times. Storing the value in QSettings is not a good idea since this involves disk access and serialzation.

retroshare-gui/src/util/DateTime.cpp
```diff
@@ -23,63 +23,102 @@
...
+// --- Time Functions ---
+
 QString DateTime::formatTime(time_t timeValue)
 {
 	return formatTime(DateTimeFromTime_t(timeValue).time());
 }
 
 QString DateTime::formatTime(const QTime &timeValue)
 {
+	int dateFormat = Settings->getDateFormat();
```
csoler: same here

retroshare-gui/src/gui/gxschannels/GxsChannelPostFilesModel.cpp
```diff
@@ -386,7 +386,7 @@ QVariant RsGxsChannelPostFilesModel::sortRole(const ChannelPostFileInfo& fmpe,in
     {
 	case COLUMN_FILES_NAME: return QVariant(QString::fromUtf8(fmpe.mName.c_str()));
 	case COLUMN_FILES_SIZE: return QVariant(qulonglong(fmpe.mSize));
-	case COLUMN_FILES_DATE: return QVariant(qulonglong(fmpe.mPublishTime));
+	case COLUMN_FILES_DATE: return DateTime::formatDateTime(fmpe.mPublishTime);
```
csoler: What you're doing here is sorting in alphabetical order of the published date, as written using the UI settings. Is that really what you intend to do? It seems to me that date sorting should always be based on the actual time it represents.

retroshare-gui/src/gui/common/FriendListModel.cpp
```diff
@@ -748,8 +748,9 @@ QVariant RsFriendListModel::displayRole(const EntryIndex& e, int col) const
                                 }
                         }
 
-                        if(col == COLUMN_THREAD_LAST_CONTACT) return QVariant(most_recent_time);
-                        if(col == COLUMN_THREAD_IP)           return QVariant(most_recent_ip);
+			// Use the standardized date format from user preferences
+			if(col == COLUMN_THREAD_LAST_CONTACT) return DateTime::formatDateTime(most_recent_time);
```
csoler: if the item model uses a number (time_t) in this column, it's very likely that the item delegate will turn it into a proper date. This is therefore in the item delegate that formatDateTime() should be used.

jolavillette:
Following Cyril's comments:
- instant refresh via global cache
- use Model-Delegate pattern with raw timestamps for correct chronological sorting

plus fix live previews in settings

retroshare-gui/src/gui/common/FriendListModel.cpp
```diff
@@ -748,8 +748,8 @@ QVariant RsFriendListModel::displayRole(const EntryIndex& e, int col) const
                                 }
                         }
 
-                        if(col == COLUMN_THREAD_LAST_CONTACT) return QVariant(most_recent_time);
-                        if(col == COLUMN_THREAD_IP)           return QVariant(most_recent_ip);
+			if(col == COLUMN_THREAD_LAST_CONTACT) return QVariant((qulonglong)DateTime::DateTimeToTime_t(most_recent_time));
```
csoler: I think you forgot to fix this one. Model => return the actual time_t

jolavillette: fixed

PR #3120, opened 2026-01-13
allow changing pseudonymity in identity edit dialog

dbear496:
This PR enables switching the pseudonymity of an identity from the GUI.

My testing indicates that switching from pseudonymous to linked works pretty well. Though it can take a considerable time for friends to recognize the change from "linked to unknown node" to "linked to friend node".

Switching from linked to pseudonymous does not work fully. The owning node seems to not recognize the change at all -- the identity is still labeled "Identity owned by you, linked to your Retroshare node". Other nodes seem to get the update since they change the label to "Anonymous identity"; however, the serviceString still indicates that the identity is linked.

Altogether, I think this is acceptable. The use case of switching from linked to pseudonymous is dubious anyway since obviously any malicious node could easily just retain the pgp signature and/or remember the old link between the identity and profile.

### DIFF
--- a/retroshare-gui/src/gui/Identity/IdEditDialog.cpp
+++ b/retroshare-gui/src/gui/Identity/IdEditDialog.cpp
@@ -320,9 +320,15 @@ void IdEditDialog::loadExistingId(const RsGxsIdGroup& id_group)
 	{
 		ui->radioButton_Pseudo->setChecked(true);
 	}
-	// these are not editable for existing Id.
-  ui->radioButton_GpgId->setEnabled(true);
-  ui->radioButton_Pseudo->setEnabled(true);
+  // typically don't allow switching from linked to anonymous because it could
+  // lead a false sense of security
+  #ifndef ALLOW_ID_SWITCH_TO_ANONYMOUS
+  if(realid)
+  {
+    ui->radioButton_GpgId->setEnabled(false);
+    ui->radioButton_Pseudo->setEnabled(false);
+  }
+  #endif
 
 	// DOES THIS TRIGGER ALREADY???
 	// force - incase it wasn't triggered.
--- a/retroshare-gui/src/gui/Identity/IdEditDialog.cpp
+++ b/retroshare-gui/src/gui/Identity/IdEditDialog.cpp
@@ -154,7 +154,7 @@ void IdEditDialog::setupNewId(bool pseudo,bool enable_anon)
             std::cerr << "IdEditDialog::setupNewId: Error. Cannot init with pseudo-anonymous id when anon ids are disabled." << std::endl;
             pseudo = false ;
         }
-        
+
 	mIsNew = true;
 	mGroupId.clear();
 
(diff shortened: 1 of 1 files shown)


csoler:
If you have an non-anonymous identity, making it anonymous will not make everyone in the network forget who it was in the first place. This gives some false sense of security.

- anonymous identities are not "linked to unknown node". That last statement means it's signed, but you do not know the profile that signed it. Anonymous identities are just not signed.

> Switching from linked to pseudonymous does not work fully. The owning node seems to not recognize the change at all -- the identity is still labeled "Identity owned by you, linked to your Retroshare node". Other nodes seem to get the update since they change the label to "Anonymous identity"; however, the serviceString still indicates that the identity is linked.

The service string needs to be updated and ID needs to be re-loading in the cache in order for the change to get visible for you.

dbear496:
> anonymous identities are not "linked to unknown node". That last statement means it's signed, but you do not know the profile that signed it. Anonymous identities are just not signed.

I am aware of this. For switch from anonymous to linked, neighboring nodes seem to fairly quickly detect the switch from "anonymous" to "linked to unknown node", but the signature validation can have a considerable delay, so it is slow switching from "linked to unknown node" to "linked to friend node".

> This gives some false sense of security.

Do you suppose then it would be better to not offer the option at all to switch from linked to anonymous?

> The service string needs to be updated and ID needs to be re-loading in the cache in order for the change to get visible for you.

I will see about implementing this.

Would it also be good for neighbor nodes to update the service string -- effectively "forgetting" that the ID was ever linked?

dbear496:
I noticed that in some places, IDs are determined to be linked/anonymous based on the flag `RSGXSID_GROUPFLAG_REALID`, and in other places it is determined by examining the service string. (Not to mention `RSGXSID_GROUPFLAG_REALID_kept_for_compatibility`.) This can lead to conflicting information being shown to users. Is there any guidance about when one method should be used over another?

csoler:
> > anonymous identities are not "linked to unknown node". That last statement means it's signed, but you do not know the profile that signed it. Anonymous identities are just not signed.
> 
> I am aware of this. For switch from anonymous to linked, neighboring nodes seem to fairly quickly detect the switch from "anonymous" to "linked to unknown node", but the signature validation can have a considerable delay, so it is slow switching from "linked to unknown node" to "linked to friend node".
> 

Signature validation is what brings you the "linked to unknown/known node". The unknown vs. friend node is because the profile key is somewhat not accessible (means "unknown node" because in this case we only have the profile key ID and not the full profile key)


> > This gives some false sense of security.
> 
> Do you suppose then it would be better to not offer the option at all to switch from linked to anonymous?
> 

I don't like that security relies on ambiguous situations like this one. So IMO it's better to create a new identity.

> Would it also be good for neighbor nodes to update the service string -- effectively "forgetting" that the ID was ever linked?

it will be forgotten at some point, but the fact that there is no way to force this is a problem.

csoler:
> I noticed that in some places, IDs are determined to be linked/anonymous based on the flag `RSGXSID_GROUPFLAG_REALID`, and in other places it is determined by examining the service string. (Not to mention `RSGXSID_GROUPFLAG_REALID_kept_for_compatibility`.) This can lead to conflicting information being shown to users. Is there any guidance about when one method should be used over another?

The service string is the most reliable, because REALID was misused for a while and there's some cooking to determine the actual status of the ID so as to be backward compatible. I think this is done when determining the service string.

dbear496:
I disabled the UI for switching IDs from linked to anonymous to avoid leading a false sense of security. I don't know if it's worth diving into the service string stuff right now because that only benefits the linked-to-anonymous case.

PR #3125, opened 2026-01-15
update VOIP for current libavcodec

dbear496:
Replaces calls to deprecated/removed libavcodec functions with their modern equivalents. Tested with libavcodec version 62.

### DIFF
--- a/plugins/VOIP/VOIP.pro
+++ b/plugins/VOIP/VOIP.pro
@@ -43,7 +43,7 @@ INCLUDEPATH += $$PWD/../../retroshare-gui/src/temp/ui
 # when rapidjson is mainstream on all distribs, we will not need the sources anymore
 # in the meantime, they are part of the RS directory so that it is always possible to find them
 
-INCLUDEPATH += ../../rapidjson-1.1.0
+INCLUDEPATH += ../../supportlibs/rapidjson/include
 
 
 #################################### Linux #####################################
--- a/plugins/VOIP/gui/VideoProcessor.cpp
+++ b/plugins/VOIP/gui/VideoProcessor.cpp
@@ -409,7 +409,7 @@ bool JPEGVideo::encodeData(const QImage& image,uint32_t /* size_hint */,RsVOIPDa
     encoded_frame.save(&buffer,"JPEG") ;
 
     voip_chunk.data = rs_malloc(HEADER_SIZE + qb.size());
-    
+
     if(!voip_chunk.data)
         return false ;
 
@@ -431,7 +431,9 @@ bool JPEGVideo::encodeData(const QImage& image,uint32_t /* size_hint */,RsVOIPDa
 
 FFmpegVideo::FFmpegVideo()
 {
+#if LIBAVCODEC_VERSION_MAJOR < 58
     avcodec_register_all();
+#endif
     // Encoding
 
     encoding_codec = NULL ;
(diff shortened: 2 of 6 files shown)


plugins/VOIP/VOIP.pro
```diff
@@ -43,7 +43,7 @@ INCLUDEPATH += $$PWD/../../retroshare-gui/src/temp/ui
 # when rapidjson is mainstream on all distribs, we will not need the sources anymore
 # in the meantime, they are part of the RS directory so that it is always possible to find them
 
-INCLUDEPATH += ../../rapidjson-1.1.0
```
csoler, line 46: I have no idea what rapidjson is doing in the VOIP plugin. Do you?

plugins/VOIP/gui/VideoProcessor.cpp
```diff
@@ -686,13 +693,22 @@ bool FFmpegVideo::encodeData(const QImage& image, uint32_t target_encoding_bitra
...
 	pkt.size = 0;
 
 	//    do
 	//    {
 	int ret = avcodec_encode_video2(encoding_context, &pkt, encoding_frame_buffer, &got_output) ;
+#else
+
+  int ret = avcodec_send_frame(encoding_context, encoding_frame_buffer);
+  if(!ret || ret == AVERROR(EAGAIN))
+    ret = avcodec_receive_packet(encoding_context, &pkt);
```
csoler: some libs (e.g. openssl) require an arbitrary number of calls after EAGAIN (meaning you should call until it works). Isn't it the case for libav as well?

plugins/VOIP/VOIP.pro
dbear496: I don't think VOIP uses rapidjson directly, but libretroshare uses it in some of its headers so it is transitively required.

plugins/VOIP/gui/VideoProcessor.cpp
dbear496: Here is a quick overview of how the new avcodec encode/decode functions work: https://ffmpeg.org/doxygen/8.0/group__lavc__encdec.html. Essentially, when `avcodec_send_frame` fails with `AVERROR(EAGAIN)`, that means the internal buffer is full and it must be emptied with one or more calls to `avcodec_receive_packet`. Similarly, when `avcodec_receive_packet` fails with `AVERROR(EAGAIN)`, that means the internal buffer is empty and it must be filled with one or more calls to `avcodec_send_frame`.

I cheat a little bit here because this is a live feed. If `avcodec_send_frame` fails with `AVERROR(EAGAIN)`, then I just drop the frame. It seems that in practice video frames and packets have a one-to-one affinity, so this case should never happen anyway. Perhaps it would be more prudent to loop through several calls to `avcodec_receive_packet` and drop the returned packets to avoid increased latency from a full buffer. Though I am unsure of the implications of dropping packets; I do not know: depending on the type of compression, it may corrupt the stream; whereas I think dropping frames is relatively safe.

The root issue is the function signature of `VideoCodec::encodeData` assumes a one-to-one affinity between frames and packets, making it impossible to produce more than one packet for a single frame. I also couldn't find any authoritative statement that `avcodec_send_packet` could or not consume concatenated data produced by more than one call to `avcodec_receive_packet`; thus, to avoid issues on the decode side, I did not want to squeeze more than one packet into a single `RsVOIPDataChunk`.

dbear496: tldr, if I had a better understanding of libavcodec, I could probably make this better, but some questions seem to go unanswered by libavcodec docs.

csoler: for sure you cannot wait for all the data to come in a loop. In openssl what we do is keep the state (last returned value) and continue when coming back in the same method. If you do not fix this now please add a comment (your middle paragraph above), so that if one needs to fix this later (as you say because it somehow breaks the encoding) then it will be much easier to spot why it fails.

plugins/VOIP/VOIP.pro
dbear496: I don't have the qmake knowhow to make the requirement transitive. I know in CMake, it would be done declaring the include path with the `PUBLIC` keyword.

plugins/VOIP/gui/VideoProcessor.cpp
dbear496: Okay, I updated the code so now it calls `avcodec_receive_packet` in a loop. I had to modify the function signature for `VideoCodec::encodeData` so it can produce more than one `RsVOIPDataChunk`.

PR #3129, opened 2026-01-17
BW statistics: remove useless values, display outqueue size in bytes, drain, total in, total out, save config

jolavillette:
This pr needs libretroshare pr/245

Refactor: Bandwidth Control Window Improvements

- Sticky Totals Row: Implemented a custom `BwCtrlWidgetItem` that keeps the "Totals" row pinned at the top of the list, regardless of the column being sorted or the sort direction.

- Totals row correctly hides irrelevant data

- Numeric data (Rates, Max, Allowed) now features 1-decimal precision and is right-aligned for better readability.

- New column Drain = OutQueue bytes / Out KB/s column display the estimated time to empty the OutQueue at current speed

- Conditional Visual Alerts: Added dynamic coloring to the Drain column to highlight network congestion: Orange for values > 30s and Red with white text for values > 60s.

- Implemented column width persistence using `QSettings` to maintain user layout preferences across sessions.

Code and comments by Gemini


jolavillette:
Per peer total in and total out are lost when the peer disconnects
Global total in and total out are persistent throughout session
All values are lost when the session is closed

defnax:
> Per peer total in and total out are lost when the peer disconnects Global total in and total out are persistent throughout session All values are lost when the session is closed

Maybe we need a feature when this done anytime later keep data till connected again start from last calculated one , I think then need storing/loading feature

csoler:
outAllowed and InAllowed *are* relevant: they are the limits when accounting for the upload/DL limit of each friend.
So currently OutMax for instance is inconsistent with what is in the columns below.

jolavillette:
I made a mistake in the PR description.

As you can see in the PR code or if you ran it, outAllowed is still here.

- outAllowed is a value that we receive from our friend via the bandwidthctrl service: it is the maximum speed that the friend is willing to receive from us, it is calculated by the friend in pqihandler
- outMax is a value that we derive in pqihandler from our own constraint on UP BW, taking into account that outMax cannot be greater than outAllowed

- the same applies to inAllowed received from our friend, that's the max he wants to send to us
- and also for inMax, except that it does not take into account inAllowed received from our friend. that's why inMax and inAllowed have always the same values, and we could remove inAllowed

see the code in pqihandler::UpdateRates

csoler:
Very cool reshape. It's more useful now! would you mind adding tooltips for each column (on the header) to describe what's in each column?

jolavillette:
InAllowed column removed because peers don't provide any information about their upload capacity
see rsconfig.h and p3bwctrl.cc
The only relevant value is our download capacity calculated in pqihandler.cc and displayed in the InMax column

PR #3142, opened 2026-01-25
Fix Identity/Chat refresh logic to handle instant avatar updates

jolavillette:
Need libretroshare pr/253

Code and comment by Antigravity

Fix Identity/Chat refresh logic to handle instant avatar updates
This PR updates the GUI to listen for GXS Identity events, fixing a regression where avatars would disappear or fail to update due to cache invalidation in the core.
1. Fix People Tab (IdentityListModel)
- **Issue:** The existing [timerUpdate] logic only refreshed category headers, failing to repaint child Identity items when their data changed (e.g., after a cache miss/reload).
- **Fix:** Added [handleIdentityEvent] to listen for `RsEventType::GXS_IDENTITY`. It triggers a precise `dataChanged` signal for the specific identity row, forcing an immediate repaint.
2. Fix Chat Lobby (ChatLobbyDialog)
- **Issue:** Participant icons in the Chat Lobby were cached and never updated unless the lobby was reloaded.
- **Fix:** Added `handleIdentityEvent` to force an update of the participant's `QTreeWidgetItem` icon when their identity changes.

### DIFF
--- a/retroshare-gui/src/gui/Identity/IdDialog.cpp
+++ b/retroshare-gui/src/gui/Identity/IdDialog.cpp
@@ -451,6 +451,9 @@ IdDialog::IdDialog(QWidget *parent)
     updateIdTimer.setSingleShot(true);
 	connect(&updateIdTimer, SIGNAL(timeout()), this, SLOT(updateIdList()));
 
+    updateCirclesTimer.setSingleShot(true);
+	connect(&updateCirclesTimer, SIGNAL(timeout()), this, SLOT(updateCircles()));
+
     mFontSizeHandler.registerFontSize(ui->idTreeWidget, 0, [this] (QAbstractItemView*, int fontSize) {
         // Set new font size on all items
         mIdListModel->setFontSize(fontSize);
@@ -540,7 +543,10 @@ void IdDialog::handleEvent_main_thread(std::shared_ptr<const RsEvent> event)
 		case RsGxsCircleEventCode::CACHE_DATA_UPDATED:
 
 			if (isVisible())
-                updateCircles();
+            {
+                if(!updateCirclesTimer.isActive())
+                    updateCirclesTimer.start(200);
+            }
 			else
 				needUpdateCirclesOnNextShow = true;
 		default:
@@ -1426,13 +1432,11 @@ void IdDialog::updateIdListRequest()
 {
     if(updateIdTimer.isActive())
     {
-        std::cerr << "updateIdListRequest(): restarting timer"<< std::endl;
         updateIdTimer.stop();
         updateIdTimer.start(1000);
     }
     else
     {
-        std::cerr << "updateIdListRequest(): starting timer"<< std::endl;
         updateIdTimer.start(1000);
     }
 }
--- a/retroshare-gui/src/gui/Identity/IdDialog.cpp
+++ b/retroshare-gui/src/gui/Identity/IdDialog.cpp
@@ -545,7 +545,7 @@ void IdDialog::handleEvent_main_thread(std::shared_ptr<const RsEvent> event)
 			if (isVisible())
             {
                 if(!updateCirclesTimer.isActive())
-                    updateCirclesTimer.start(200);
+                    updateCirclesTimer.start(1000);
             }
 			else
 				needUpdateCirclesOnNextShow = true;
@@ -1430,15 +1430,7 @@ void IdDialog::updateSelection(const QItemSelection& /* new_sel */,const QItemSe
 
 void IdDialog::updateIdListRequest()
 {
-    if(updateIdTimer.isActive())
-    {
-        updateIdTimer.stop();
-        updateIdTimer.start(1000);
-    }
-    else
-    {
-        updateIdTimer.start(1000);
-    }
+    updateIdTimer.start(1000);
 }
 
 void IdDialog::updateIdList()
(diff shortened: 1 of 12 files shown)


jolavillette:
Fixed UI freeze in People view by throttling IdentityListModel updates

The `IdentityListModel` was processing `NEW_IDENTITY` and `DELETED_IDENTITY` events synchronously, causing full model reloads for every single event. This led to UI freezes during event bursts.

This commit:
- Implements a throttling mechanism using a 200ms single-shot timer.
- Coalesces rapid `NEW` and `DELETED` events into a single update.
- Removes a duplicate `QTimer` allocation in the constructor.

retroshare-gui/src/gui/gxschannels/GxsChannelDialog.cpp
```diff
@@ -67,7 +70,7 @@ void GxsChannelDialog::handleEvent_main_thread(std::shared_ptr<const RsEvent> ev
         switch(e->mChannelEventCode)
         {
         case RsChannelEventCode::STATISTICS_CHANGED:      // [[fallthrough]];
-            updateDisplay(true);                          // no breaks, on purpose!
+            if(!mUpdateTimer->isActive()) mUpdateTimer->start(200);
```
csoler: this line doesn't appear to be necessary because of the fallthrough below (the updateDisplay() was also here before by mistake)

retroshare-gui/src/gui/gxschannels/GxsChannelDialog.cpp
```diff
@@ -82,7 +85,7 @@ void GxsChannelDialog::handleEvent_main_thread(std::shared_ptr<const RsEvent> ev
         case RsChannelEventCode::NEW_CHANNEL:             // [[fallthrough]];
         case RsChannelEventCode::DELETED_CHANNEL:             // [[fallthrough]];
         case RsChannelEventCode::SUBSCRIBE_STATUS_CHANGED:// reloads group summary (calling GxsGroupFrameDialog parent method)
-            updateDisplay(true);
+            if(!mUpdateTimer->isActive()) mUpdateTimer->start(200);
```
csoler: I understand that the timer is here to factor multiple updates in a single call to updateDisplay. But if so, 200ms is rather short.

retroshare-gui/src/gui/Identity/IdDialog.cpp
```diff
@@ -1426,13 +1432,11 @@ void IdDialog::updateIdListRequest()
 {
     if(updateIdTimer.isActive())
     {
-        std::cerr << "updateIdListRequest(): restarting timer"<< std::endl;
         updateIdTimer.stop();
```
csoler: apparently the call to stop() is not necessary, or at least not consistent with the rest of the code in this PR.

retroshare-gui/src/gui/Identity/IdentityListModel.cpp
```diff
@@ -56,8 +57,71 @@ RsIdentityListModel::RsIdentityListModel(QObject *parent)
...
-    mIdentityUpdateTimer = new QTimer();
+    mIdentityUpdateTimer = new QTimer(this);
     connect(mIdentityUpdateTimer,SIGNAL(timeout()),this,SLOT(timerUpdate()));
+
+    mThrottlingTimer = new QTimer(this);
+    mThrottlingTimer->setSingleShot(true);
+    mThrottlingTimer->setInterval(200);
+
+    // Using a lambda for connection to avoid adding a new slot in header file if not strictly necessary,
+    // though the plan mentioned a slot. But wait, checkInternalData is a slot?
```
csoler: When you copy the comments from the AI, you may shorten them to remove its "internal" thoughts ;-)

retroshare-gui/src/gui/Identity/IdentityListModel.cpp
```diff
@@ -56,8 +57,71 @@ RsIdentityListModel::RsIdentityListModel(QObject *parent)
...
+            if(idx.isValid())
+            {
+                EntryIndex e;
+                convertInternalIdToIndex(idx.internalId(), e);
+                
+                if(e.type == ENTRY_TYPE_IDENTITY && e.category_index < mCategories.size() && e.identity_index < mCategories[e.category_index].child_identity_indices.size())
+                {
+                    // Force refresh of details
+                    mIdentities[mCategories[e.category_index].child_identity_indices[e.identity_index]].last_update_TS = 0;
+                    emit dataChanged(idx, idx);
```
csoler: Does this actually work? The member mIdentities needs to be updated. I think you need a specal function in IdentityListModel to do that because it currently only allows to entirely reload the list of identities.

csoler (commented):
Some potential issues.

retroshare-gui/src/gui/Identity/IdentityListModel.cpp
jolavillette: fixed according to your comment:

Added refreshIdentityDetails(const RsGxsId& id) to IdentityListModel
Instead of just resetting last_update_TS and hoping dataChanged triggers a reload:
- Reloads details from backend (getIdDetails)
- Checks if category changed (Own / Contacts / All) → triggers full model reload if so
- Updates in-place + emits dataChanged if category is unchanged

PR #3149, opened 2026-02-02
Turtle Router Statistics: improve sorting, data formatting, and memory usage

jolavillette:
### Memory Usage Optimization
* **Dynamic UI Cleanup**: Overloaded [hideEvent()] to clear all `QTreeWidget` items when the dialog is hidden. This prevents keeping thousands of inactive `QTreeWidgetItem` objects in RAM when the statistics are not being actively viewed.
* **Display Limiting**: Restricted the **Tunnel Requests** list to the 100 most recent entries. This provides a sufficient window for sorting and analysis while preventing excessive memory consumption and UI sluggishness on highly active routing nodes.
* **Resource Management**: Implemented [TurtleTreeWidgetItem] to handle numeric and complex string sorting efficiently without redundant data duplication.

### DIFF
--- a/retroshare-gui/src/gui/statistics/TurtleRouterDialog.cpp
+++ b/retroshare-gui/src/gui/statistics/TurtleRouterDialog.cpp
@@ -356,7 +356,7 @@ void TurtleRouterDialog::updateTunnelRequests(	const std::vector<std::vector<std
 	}
 	top_level_s_requests->setText(0, tr("Search requests") + " (" + QString::number(search_reqs_info.size()) + ")" ) ;
 
-	uint32_t limit = std::min((uint32_t)tunnel_reqs_info.size(), (uint32_t)100);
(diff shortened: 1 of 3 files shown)


retroshare-gui/src/gui/statistics/TurtleRouterDialog.cpp
```diff
@@ -225,7 +200,27 @@ void TurtleRouterDialog::processSettings(bool bLoad)
 
 }
 
-bool sr_Compare( TurtleSearchRequestDisplayInfo m1,  TurtleSearchRequestDisplayInfo m2)  { return m1.age < m2.age; }
+void TurtleRouterDialog::hideEvent(QHideEvent *event)
+{
+	// Clear tree widgets to save memory (Cyril's recommendation)
```
csoler, line 205: just call _f2f_TW->clear(): it will cleanly delete all its items. No need to do it manually.

PR #3152, opened 2026-02-08
ADD: libavcodec-dev in Ubuntu workflows

JohannesKleine:


defnax:
Qt6 has no` QCameraInfo `more
i think you need rewrite code to can use qt5 or qt6 defines
image

JohannesKleine:
ok then my latest commit is here nonsense

G10h4ck:
AFAIU Qt5 is EOL so we should hurry up porting to Qt6 and get rid of Qt5 stuff

PR #3158, opened 2026-02-11
added missing initialization to rsevent handler

csoler:


csoler:
I'm merging this right away as it is a trivial bug fix.

PR #3159, opened 2026-02-12
fix GXS details not loading at startup, and Subscribe button not updating when subscribing

jolavillette:
fix GXS details not loading at startup, and Subscribe button not updating when subscribing

# Fix: GXS Channels/Forums Details (Startup) & Subscribe Button
This PR addresses two specific issues in GXS Channels and Forums GUI.
## 1. Missing Group Details at Startup
**Issue:**
Channel details (title, icon, description) were failing to load *at startup* in the details view.
**Fix:**
-   Modified `GxsChannelPostsModel.h` to expose `getChannelGroup()`, allowing the widget to correctly retrieve group data during its initialization phase.
## 2. Subscribe Button Not Updating
**Issue:**
The "Subscribe" button in the channel/forum view did not update to "Subscribed" after being clicked (it remained in the "Subscribe" state).
**Fix & Race Condition Resolution:**
-   We initially added a `SUBSCRIBE_STATUS_CHANGED` handler to update the button, but this introduced a side-effect: the group would remain in the "Popular" list instead of moving to "Subscribed".
-   This was caused by a race condition between the parent dialog (`GxsGroupFrameDialog`) refreshing its list and the child widget updating its button state concurrently.
-   **Final Solution:** Implemented a **1.5 second delayed async update** in `GxsChannelPostsWidgetWithModel.cpp` and `GxsForumThreadWidget.cpp`.
    -   This delay ensures the parent dialog finishes its list update *before* the button state update is triggered, solving both the button state and the list categorization issues.
## Modified Files
-   `retroshare-gui/src/gui/gxschannels/GxsChannelPostsModel.h`
-   `retroshare-gui/src/gui/gxschannels/GxsChannelPostsWidgetWithModel.cpp`
-   `retroshare-gui/src/gui/gxsforums/GxsForumThreadWidget.cpp`

### DIFF
--- a/retroshare-gui/src/gui/gxschannels/GxsChannelPostsWidgetWithModel.cpp
+++ b/retroshare-gui/src/gui/gxschannels/GxsChannelPostsWidgetWithModel.cpp
@@ -913,7 +913,7 @@ void GxsChannelPostsWidgetWithModel::handleEvent_main_thread(std::shared_ptr<con
                         {
                              RsQThreadUtils::postToObject([this, info=channelsInfo[0]](){
                                  mGroup = info;
-                                 setSubscribeButtonText(info.mMeta.mGroupId, info.mMeta.mSubscribeFlags, info.mMeta.mPop);
+                                 insertChannelDetails(mGroup);
                              }, this);
                         }
                     });
--- a/retroshare-gui/src/gui/gxschannels/GxsChannelPostsWidgetWithModel.cpp
+++ b/retroshare-gui/src/gui/gxschannels/GxsChannelPostsWidgetWithModel.cpp
@@ -899,28 +899,19 @@ void GxsChannelPostsWidgetWithModel::handleEvent_main_thread(std::shared_ptr<con
         {
             if(e->mChannelGroupId == groupId())
             {
-                // Async update of subscription status to avoid blocking the GUI thread with waitToken
-                // Delayed by 1.5s to avoid conflict with GxsGroupFrameDialog's own update (which requests all groups)
-                // This prevents the bug where the group remains in "Popular" instead of moving to "Subscribed"
-                RsGxsGroupId grpId = groupId();
-                QTimer::singleShot(1500, [this, grpId](){
-                    RsThread::async([this, grpId](){
-                        std::vector<RsGxsChannelGroup> channelsInfo;
-                        std::list<RsGxsGroupId> ids;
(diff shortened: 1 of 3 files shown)


retroshare-gui/src/gui/gxschannels/GxsChannelPostsWidgetWithModel.cpp
```diff
@@ -895,6 +895,33 @@ void GxsChannelPostsWidgetWithModel::handleEvent_main_thread(std::shared_ptr<con
         }
         break;
 
+        case RsChannelEventCode::SUBSCRIBE_STATUS_CHANGED:
+        {
+            if(e->mChannelGroupId == groupId())
+            {
+                // Async update of subscription status to avoid blocking the GUI thread with waitToken
+                // Delayed by 1.5s to avoid conflict with GxsGroupFrameDialog's own update (which requests all groups)
+                // This prevents the bug where the group remains in "Popular" instead of moving to "Subscribed"
```
csoler: I think it's a bad way to fix things. You can't rely on a timer like this to delay something hoping that everything else that could cause a conflicts is already finished.

csoler: The problem is probably elsewhere.

retroshare-gui/src/gui/gxsforums/GxsForumThreadWidget.cpp
```diff
@@ -404,6 +416,33 @@ void GxsForumThreadWidget::handleEvent_main_thread(std::shared_ptr<const RsEvent
...
             break;
+
+        case RsForumEventCode::SUBSCRIBE_STATUS_CHANGED:
+        {
+            if(e->mForumGroupId == mForumGroup.mMeta.mGroupId)
+            {
+                // Async update of subscription status to avoid blocking the GUI thread with waitToken
+                // Delayed by 1.5s to avoid conflict with GxsGroupFrameDialog's own update (which requests all groups)
+                // This prevents the bug where the group remains in "Popular" instead of moving to "Subscribed"
+                RsGxsGroupId grpId = mForumGroup.mMeta.mGroupId;
```
csoler: same comment as above: you cannot trust this way of fixing things.

csoler (commented):
Fixing things by delaying them is not really reliable. The problem is probably elsewhere.

PR #3168, opened 2026-02-24
TheWire: Fix UI

jolavillette:
# Resolve UI Freezes in "The Wire" feature
## Description
This PR addresses recurring UI freezes associated with "The Wire" feature. The root cause was identified as synchronous, blocking calls to the background service (`rsWire->getGroups`) performed on the main UI thread. These freezes occurred not only when viewing the "The Wire" tab but also in the background due to event handlers triggering data refreshes upon receiving network updates.
## Changes
- **Asynchronous Data Loading**: Refactored `WireDialog::requestGroupData` to use `RsThread::async` for fetching group data in a background thread.
- **Thread-Safe UI Updates**: Used `RsQThreadUtils::postToObject` to safely update the UI components from the background thread once data is available.
- **Non-Blocking Actions**: Converted [subscribe], [unsubscribe], and [PVHfollow] methods to be asynchronous to prevent minor stalls during user interactions.
- **Code Optimization**: Updated `WireDialog::updateGroups` to take a `const std::vector<RsWireGroup>&` to improve efficiency and ensure compatibility with the new asynchronous patterns.

### DIFF
--- a/retroshare-gui/src/gui/TheWire/WireDialog.cpp
+++ b/retroshare-gui/src/gui/TheWire/WireDialog.cpp
@@ -553,7 +553,7 @@ void WireDialog::deleteGroups()
 			std::cerr << std::endl;
 
 			item = alayout->takeAt(i);
-			delete item->widget();
+			item->widget()->deleteLater();
 			delete item;
 		}
 		else
@@ -936,7 +936,7 @@ void WireDialog::clearTwitterView()
 			std::cerr << std::endl;
 
 			item = alayout->takeAt(i);
-			delete item->widget();
+			item->widget()->deleteLater();
 			delete item;
 		}
 		else
--- a/retroshare-gui/src/gui/TheWire/WireDialog.cpp
+++ b/retroshare-gui/src/gui/TheWire/WireDialog.cpp
@@ -431,14 +431,20 @@ bool WireDialog::setupPulseAddDialog()
 
 void WireDialog::subscribe(RsGxsGroupId &groupId)
 {
-	uint32_t token;
-	rsWire->subscribeToGroup(token, groupId, true);
+	RsThread::async([groupId]()
+	{
+		uint32_t token;
+		rsWire->subscribeToGroup(token, groupId, true);
+	});
 }
 
 void WireDialog::unsubscribe(RsGxsGroupId &groupId)
 {
-	uint32_t token;
-	rsWire->subscribeToGroup(token, groupId, false);
+	RsThread::async([groupId]()
+	{
+		uint32_t token;
+		rsWire->subscribeToGroup(token, groupId, false);
+	});
 }
 
 void WireDialog::notifyGroupSelection(WireGroupItem *item)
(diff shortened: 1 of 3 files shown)


retroshare-gui/src/gui/TheWire/WireDialog.cpp
```diff
@@ -431,14 +436,20 @@ bool WireDialog::setupPulseAddDialog()
 
 void WireDialog::subscribe(RsGxsGroupId &groupId)
 {
-	uint32_t token;
-	rsWire->subscribeToGroup(token, groupId, true);
+	RsThread::async([groupId]()
+	{
+		uint32_t token;
+		rsWire->subscribeToGroup(token, groupId, true);
```
csoler: plz do not use tokens in the UI. If needed add in p3wire and rswire the proper API to do all the token-related stuff in libretroshare.

jolavillette: done, see libretroshare pr/276

TheWire: Use new token-free subscribe API

Replace subscribeToGroup(token, ...) calls with simplified subscribe() 
API that doesn't require token management in UI code.

Co-Authored-By: Claude <noreply@anthropic.com>

retroshare-gui/src/gui/TheWire/WireDialog.cpp
```diff
@@ -547,7 +556,7 @@ void WireDialog::deleteGroups()
 			std::cerr << std::endl;
 
 			item = alayout->takeAt(i);
```
csoler, line 558: First time I see this. delete item is not sufficient? It seems to me that QTreeWidgetItem holds the memory for it's own widget.

jolavillette: the AI answered this:

"Hi @csoler,

You are absolutely right that a QTreeWidgetItem manages the memory for its own widget. However, in this specific function (WireDialog::deleteGroups()), item is NOT a QTreeWidgetItem.

If you look closely at lines 546-551:

QLayout *alayout = ui.groupsWidget->layout();
QLayoutItem *item;
// ...
item = alayout->takeAt(i);
item is actually a QLayoutItem, which we retrieve by taking it out of a QLayout (not a QTreeWidget).

When you call takeAt() on a QLayout, it removes the layout item from the layout, but it does not delete the item, nor does it delete the underlying widget it manages. We have to delete both manually to avoid memory leaks.

As for why I used item->widget()->deleteLater() instead of a direct delete item->widget(): In Qt, using a forced 
delete on a widget that might still be processing events or signals in the event loop (like UI updates happening in the background during deleteGroups()) can cause unexpected crashes. Using deleteLater() schedules the widget for deletion safely at the end of the current event loop iteration, preventing these crashes."

csoler: my question is rather than when you delete the layout item, it will also delete its content, so you shouldn't have to call deleteLater() on it anyway.

jolavillette: again the AI, not me:

"I understand why you would think that, as it's a very common Qt intuition! However, according to the official Qt documentation for QLayoutItem::~QLayoutItem(): 'Destroys the QLayoutItem. This does not destroy the child widget, child layout, or child layout item. It is the caller's responsibility to delete them.'

Because QLayoutItem is just a wrapper and not a QObject that takes memory ownership of its widget, delete item only frees the item wrapper. If we don't call deleteLater() on the widget itself, the widget remains alive in the parent's memory tree and accumulates invisibly every time we refresh the groups. Therefore, deleting both the item and its widget is necessary here."

PR #3176, opened 2026-03-03
Fix the wire forced compilation

jolavillette:
Make TheWire compilation optional
To build with TheWire use qmake "CONFIG+=rsgxsthwire"


csoler:
oups sorry I didn't see this PR. Could you plz make sure it's still OK (rebase it) and tell if it can be merged?

PR #3180, opened 2026-03-03
chat nicks colors based on hash

retrop00h:


retroshare-gui/src/gui/chat/ChatStyle.cpp
```diff
@@ -356,10 +357,10 @@ QString ChatStyle::formatMessage(enumFormatMessage type
 	if (flag & CHAT_FORMATMSG_SYSTEM) {
 		color = Qt::darkBlue;
 	} else {
-		// Calculate color from the name
+		// Calculate color from the id
 		uint hash = 0;
-		for (int i = 0; i < name.length(); ++i) {
-			hash = (((hash << 1) + (hash >> 14)) ^ ((int) name[i].toLatin1())) & 0x3fff;
+		for (int i = 0; i < gxsid.length(); ++i) {
+			hash = (((hash << 1) + (hash >> 14)) ^ ((int) gxsid[i].toLatin1())) & 0x3fff;
```
csoler, line 363: you use a QString for gxsid. It would be wiser to use a RsGxsId so that gxsid[i] is a uint8_t, which allows more possibilities than a char in between [0-9] and [a-f].

PR #3181, opened 2026-03-08
Added Zoom via CTRL Z/- keys

defnax:


csoler:
Usually softwares use Ctrl+ and Ctrl- for zoom. Ctrl Z looks a bit awkward. 
Also I couldn't see where the zoom in RSTreeView actually takes effect.

PR #3184, opened 2026-03-11
Changed circles group tree icons

defnax:


csoler:
Some time ago there was a color notification system to indicate if you are member of a circle, or invited/waiting for approval or none of these (resp. green / yellow / grey). I understand that it's also nice to give colored icons to circles, but then we look this notification (which was done by the item delegate). How can we get this back?

defnax:
code updated @csoler

csoler:
The circle icons look great and can easily replace the bullets so I suggest we keep them.
- Just use the same colors than the bullets (grey/yellow/green) as it will be more clear than blue/yellow/green. 
- Also remove the "use circle icon" menu. It's really not useful.

defnax:
updated @csoler

defnax:
whats about use for admins the blue icon? then we see in ouer eyes better thats my?

csoler:
I prefer that we only use one visual for one information type. Now the icon color is about membership. So ownership should be shown with something else. One option would be to 
- use the tree for ownership (Circles I administrate / other circles)
- use bullet colors for membership (as now)

In the tree there is "Circles I belong to", which is inconsistent. Other option:
- use the tree for membership (Circles I belong to / pending membership / invitations / other visible circles)
- use bullet colors for ownership

I prefer the first one, but it's up to you. What matters imo is that the info is not mixed up across multiple formats.

defnax:
at the moment its done not doing more things here maybe in future, its done now for merge

defnax:
done for merg @csoler

PR #3186, opened 2026-03-17
GUI/Channels: Fix channel appearing unsubscribed after posting by reloading metadata instead of toggling flag

jolavillette:
GUI/Channels: Fix channel appearing unsubscribed after posting by reloading metadata instead of toggling flag

bug likely introduced by pr/3159 FixChannelsDetailsAndSubscribedButton commit 1658f9bd

### DIFF
--- a/retroshare-gui/src/gui/gxsforums/GxsForumThreadWidget.cpp
+++ b/retroshare-gui/src/gui/gxsforums/GxsForumThreadWidget.cpp
@@ -1761,7 +1761,7 @@ void GxsForumThreadWidget::editForumMessageData(const RsGxsForumMsg& msg)
         return;
     }
 
-    CreateGxsForumMsg *cfm = new CreateGxsForumMsg(groupId(), msg.mMeta.mParentId, msg.mMeta.mMsgId, msg.mMeta.mAuthorId, mForumGroup.mAdminList.ids);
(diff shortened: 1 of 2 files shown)


retroshare-gui/src/gui/gxsforums/GxsForumThreadWidget.cpp
```diff
@@ -1761,7 +1761,7 @@ void GxsForumThreadWidget::editForumMessageData(const RsGxsForumMsg& msg)
         return;
     }
 
-    CreateGxsForumMsg *cfm = new CreateGxsForumMsg(groupId(), msg.mMeta.mParentId, msg.mMeta.mMsgId, msg.mMeta.mAuthorId, mForumGroup.mAdminList.ids);
+    CreateGxsForumMsg *cfm = new CreateGxsForumMsg(groupId(), msg.mMeta.mParentId, msg.mMeta.mOrigMsgId, msg.mMeta.mAuthorId, mForumGroup.mAdminList.ids);
```
zapek: Why do you do this? It has nothing to do with the fix and it breaks the specs (again).

zapek (changes requested):

retroshare-gui/src/gui/gxsforums/GxsForumThreadWidget.cpp
jolavillette: GxsForumThreadWidget.cpp was commited by mistake. Restored.
Cyril would have politely pointed the issue.

PR #3188, opened 2026-03-19
Posted: Improved context menu with copy text/links and fixed right-click trigger in RSTreeView

jolavillette:
Posted: Improved context menu with copy text/links and fixed right-click trigger in RSTreeView

# Summary of Changes - RetroShare

### 1. Improved Right-Click UX
*   **Files:** `RSTreeView.h/.cpp`
*   **Action:** Fixed the bug where two right-clicks were required to open the context menu. The menu now appears on the **first click**.

### 2. Context Menu Enhancements (Posts)
*   **File:** [PostedListWidgetWithModel.cpp]
*   **Action:** Added requested actions without increasing UI complexity:
    *   **Copy Selected Text**: Option to copy text selected within the post notes.
    *   **Link Extraction**: The menu automatically scans the description and offers unique "Copy link" actions for any HTTP/HTTPS URLs found.
    *   **Conditional Display**: HTTP link options only appear when the post is expanded. The main RetroShare link remains always available.

### 3. Stability and Security
*   **File:** `RSTextBrowser.cpp`
*   **Action:** Fixed a potential crash when clicking on links pointing to non-existent local files.

### 4. Implementation Minimalism
*   **Action:** Reverted the complex migration to `RSTextBrowser` for notes, returning to the original `QLabel` based implementation. This ensures the codebase remains lightweight and easy to maintain with only **4 GUI files modified**.

### DIFF
--- a/retroshare-gui/src/gui/Posted/PostedListWidgetWithModel.cpp
+++ b/retroshare-gui/src/gui/Posted/PostedListWidgetWithModel.cpp
@@ -239,6 +239,11 @@ QWidget *PostedPostDelegate::createEditor(QWidget *parent, const QStyleOptionVie
 	w->updateGeometry();
 	w->adjustSize();
 
+    // Ensure all parts of the editor bubble right-clicks to the view.
+    w->setContextMenuPolicy(Qt::NoContextMenu);
+    foreach(QWidget *child, w->findChildren<QWidget*>())
+        child->setContextMenuPolicy(Qt::NoContextMenu);
+
 	return w;
 }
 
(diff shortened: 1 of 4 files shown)


csoler:
It seems to work, except for one thing: when the title of the post is not a RS link but a http link, and there's no http link elsewhere in the post, then the menu shows "Copy http link" instead of "Copy link: [the link]" as the code seems to indicate. I don't understand why.

See the 2nd post of this board for instance: retroshare://posted?name=test2&id=ef34da9d70772646b4a2a19e6f28d446

csoler:
apparently it's only looking in the notes for links, but then why does the menu also say "copy http link" when there's one in the title? The fix should be that the menu entry for http links in the title should also mention the link to copy.

PR #3192, opened 2026-04-03
[Wire] Added to count Wire group box & read message

defnax:
now this pr require -> https://github.com/RetroShare/libretroshare/pull/283


csoler:
It seems ok. One detail: the setAllAsRead() function will trigger many events. If they overload the UI, then think of using a timer (usually timers are not so good, but this can save you in this type of situation).

PR #3199, opened 2026-04-13
feat: Bandwidth Totals KB/MB/GB + Charts

HrachShah:
New:
- BwCtrlWindow GB cols/session totals per friend/total
- Graphs: speed + cumulative GB
- Turtle Router GB labels
- TrafficChartsPage Pie/Bar + Statistics tab
Demo: bandwidth-demo.html


csoler:
This PR only contains a .txt file that actually contains the pasted code of multiple cpp files. This may be the result of an editing error. In any case we can close this.

PR #3194, opened 2026-04-06
Improve Usage stats & added circle links

defnax:
*Added for circles retroshare links feature

PR #3200, opened 2026-04-17
Added circle links

defnax:


defnax:
this not needed always done in #3194

PR #3204, opened 2026-04-20
[Wire] Added to show which posts are new & read function

defnax:


retroshare-gui/src/gui/TheWire/PulseViewItem.cpp
```diff
@@ -305,7 +309,7 @@ void PulseDataItem::showPulse()
 						pixmap,GxsIdDetails::ORIGINAL))
 				{
 					headshotOkay = true;
-					pixmap = pixmap.scaled(50,50);
+					pixmap = pixmap.scaled(50,50,Qt::KeepAspectRatio,Qt::SmoothTransformation);
```
csoler, line 312: Using an absolute size like this is a bit suspicious. Not a problem for the PR itself, but you should consider computing the appropriate size as a factor of the font size.

csoler (approved):

PR #3205, opened 2026-04-21
[Fix] Fixed import Friends list with groups 🐞

defnax:


retroshare-gui/src/gui/common/NewFriendList.cpp
```diff
@@ -1574,7 +1574,7 @@ bool NewFriendList::importFriendlist(QString &fileName, bool &errorPeers, bool &
             while (!pgpID.isNull()) {
                 // add pgp id to group
                 RsPgpId rsPgpId(pgpID.attribute("id").toStdString());
-                if(rsPgpID.isNull() || !rsPeers->assignPeerToGroup(groupId, rsPgpId, true)) {
+                if(!rsPeers->assignPeerToGroup(groupId, rsPgpId, true)) {
```
csoler: This is wrong. I think it should be if(rsPgpId.isNull() || ...)
There's another variable above called rsPgpID but the one to use is the one just created before this line.

csoler (changes requested):

PR #3207, opened 2026-04-24
Mark contacts bold in chat userlist

retrop00h:


retroshare-gui/src/gui/chat/ChatLobbyDialog.cpp
```diff
@@ -633,6 +633,15 @@ void ChatLobbyDialog::updateParticipantsList()
                 widgetitem->setData(COLUMN_NAME, Qt::ForegroundRole, QVariant());
             }
 
+            // Mark contacts in bold
+            RsIdentityDetails detail;
+            bool isContact = rsIdentity->getIdDetails(it2->first, detail)
```
csoler, line 638: you could directly call rsIdentity->isARegularContact(it2->first))

csoler: it's actually a bit faster since it only looks into the contact list instead of populating the entire RsIdentityDetails structure.

csoler (changes requested):
I approve the idea. Just follow the comment (one line to replace) and it can be merged.

PR #3209, opened 2026-04-25
Added function to export & import contacts

defnax:


retroshare-gui/src/gui/Identity/IdDialog.cpp
```diff
@@ -2802,3 +2811,121 @@ void IdDialog::sortColumn(int col,Qt::SortOrder so)
...
+    ownGroup.setAttribute("name", "own");
+    root.appendChild(ownGroup);
+
+    QDomElement contactGroup = doc.createElement("group");
+    contactGroup.setAttribute("name", "contacts");
+    root.appendChild(contactGroup);
+
+    // Get all identity summaries
+    std::list<RsGroupMetaData> summaries;
+    if (!rsIdentity->getIdentitiesSummaries(summaries)) return false;
```
csoler: So you're getting all identities, then discard the ones that are not contacts? That's pretty inefficient. It would be better to get the selection from the tree widget and then for each of them see if you put them in ownGroup of contactGroup.

retroshare-gui/src/gui/Identity/IdDialog.cpp
```diff
@@ -2802,3 +2811,121 @@ void IdDialog::sortColumn(int col,Qt::SortOrder so)
...
+
+    for (const RsGroupMetaData& meta : summaries) {
+        RsGxsId id(meta.mGroupId);
+        RsIdentityDetails details;
+
+        // Check if it's an OWN ID or a regular CONTACT
+        if (rsIdentity->getIdDetails(id, details)) {
+            bool isOwn = (details.mFlags & RS_IDENTITY_FLAGS_IS_OWN_ID);
+            bool isContact = rsIdentity->isARegularContact(id);
```
csoler: it's faster to look for RS_IDENTITY_FLAGS_IS_A_CONTACT

csoler (commented):
The behavior of this PR is rather inconsistent: the selected list of contacts has no impact on what is saved, and sometimes, nothing is saved at all (empty list in a correct xml format).

Why can we only save contacts? You may want to export any selected identity (or a group of them). Same for import.

PR #3213, opened 2026-05-01
Fixed SecurityItem

defnax:


.github/workflows/ci-mingw64-Qt6.yml
```diff
@@ -65,6 +65,11 @@ jobs:
           git submodule update --init --remote libbitdht/ libretroshare/ retroshare-webui/
           git submodule update --init supportlibs/librnp supportlibs/rapidjson supportlibs/restbed
 
+      - name: Workaround librnp strlen build regression
```
csoler, line 68: it's terrible that we have to do this.

csoler (changes requested):
BAD_CERTIFICATE is only issued when the connection is *not* a RS user: the PGP id is missing, or the SSL certificate is not supplied. So I think we should do this:

Hide name, trust, Location, PeerId. Only the IP is actually relevant in this case.

In SecurityItem.cpp:105: add the IP in the identifier, so that we can see all connection attempts instead of only the first one.

What are "Connection Method: Neighbor" and "Status: Neighbor" supposed to mean?

.github/workflows/ubuntu-qt5_c-cpp.yml
```diff
@@ -22,7 +22,7 @@ jobs:
     - name: run apt update
       run: sudo apt-get update
     - name: apt install
-      run: sudo apt-get install g++ cmake qt5-qmake qtmultimedia5-dev libqt5x11extras5-dev libasio-dev libbz2-dev libjson-c-dev libssl-dev libsqlcipher-dev libupnp-dev libxss-dev rapidjson-dev libbotan-2-dev doxygen libsecret-1-dev libxml2-dev libxslt1-dev libcurl4-openssl-dev
+      run: sudo apt-get install g++ cmake qt5-qmake qtmultimedia5-dev libqt5x11extras5-dev libasio-dev libbz2-dev libjson-c-dev libssl-dev libsqlcipher-dev libupnp-dev libxss-dev rapidjson-dev libbotan-2-dev doxygen libsecret-1-dev libxml2-dev libxslt1-dev libcurl4-openssl-dev libavcodec-dev libavutil-dev libspeex-dev libspeexdsp-dev
```
csoler, line 25: this is irrelevant to this PR.

.github/workflows/ci-ucrt64.yml
```diff
@@ -64,6 +64,11 @@ jobs:
           git submodule update --init --remote libbitdht/ libretroshare/ retroshare-webui/
           git submodule update --init supportlibs/librnp supportlibs/rapidjson supportlibs/restbed
 
+      - name: Workaround librnp strlen build regression
```
csoler, line 67: This is irrelevant to this PR.

csoler (changes requested):
The changes should only be in SecurityItem.cpp. Please remove other changes that are not relevant.

defnax:
i will look next days on this

csoler (changes requested):
It needs to be improved a bit before merging:

In the situation when the ID is 0, the item should:
- hide the Peer Details button (it's irrelevant)
- say "connection attempt from [IP]". Now it says "connection attempt from 00000..." which is irrelevant.
- remove the blank line on top so that "connection attempt" is aligned with the top of the item
- if possible, show the IP better when it's IPv4. For instance when it's 185.247.137.178 it shows ipv6://[::ffff:185.247.137.178]:36587. Just show the IP.
- do not allow to expand the item. There's no relevant information in it (name, peer ID, trust, location, etc: all irrelevant)

We also face this problem now: there are many connection attempts. I see multiple options to fix this:
1 - only keep 1 item for unknown connection attempts (meaning Id=0), listing the last 20 attempts IPs+ports+time (in this case don't put the IP in the hash). In the item just show the latest, and when you unfold it, show them all. When you get a new connection attempt, just update the item if already there.
2 - simpler but less efficient: when you compute the hash, don't include the port. It will reduce the number of items.

Issue #306, opened 2016-03-13
There are functions which use MD5 encryption


realityfabric:
MD5 encryption has been known to be vulnerable for 20 years, and has been proven to be vulnerable more than 10 years ago.

TheSentry:
MD5 is a hashing algorithm, not encryption.

But yes, we could set an example and use SHA-3 instead.

csoler:
MD5 should not be used for signature. Where did you find this??

realityfabric:
It's in some of the files inside of the openpgpsdk libraries. It's possible that the md5 portions aren't actually used, but if they aren't they should be removed. The license allows for the modification of the libraries so long as the changes are logged.

PR #3216, opened 2026-05-05
gui: add Friend Requests page (closes #1794)

### LINKED ISSUE #1794: Add a Friend Requests Page
Display the **Friend Requests** in a own page would be useful on retroshare.

We has
"**Friend Requests**" Notify for feed/messages/toaster that is when some one added you as friend to connect but it display as one time popup on news feeds.

You have no good over view more who wanted Friends with you. every day the poping up is not userfriendly. After restart is everything lost.

ideas:

* Notify with counting on Top of the Contacts list
* A Friend Requests Tab for list all requests.
* Reject FR feature ( not popup every time)
* After restart visible too ( needs to be stored? remember it)

fr
fr2

64johnlee:
## Description
Adds a new Friend Requests page to the RetroShare GUI, allowing users to view and manage incoming friend requests.

## Changes
- Added Friend Requests GUI page
- Updated libretroshare submodule for friend requests backend
- Implemented friend requests implementation files

## Testing
- Built on Linux (Qt-based)
- GUI page visible in navigation

## Bounty
Closes #1794


64johnlee:
Backend companion PR opened: RetroShare/libretroshare#306 — it adds the `rsFriendRequest` service (`p3FriendRequest`) that this GUI page links against, plus the CMake/qmake registration. This superproject branch's `libretroshare` submodule pointer references the head of that branch, so the two should be merged together. Also rebased on latest `master` to clear the previous conflict.

csoler:
Showing friend requests is a nice idea. 
Adding yet another tab is somewhat problematic: we have too many already and the managing of friends is already in the Network tab. 

I think that it would be much better to add your new UI in either the right tabs of Network (next to broadcast, network graph and keyring) or as a top level list element "Pending friend requests" in the Friend list (network tab left part). This is a bit more tricky to do (you'd have to modify the existing friend list model) but has its merit as well.

64johnlee:
Thanks @csoler — that makes sense, the tab list is already crowded and friend management belonging under Network is the right call.

I've reworked it to your first suggestion: the Friend Requests UI is now a sub-tab inside the **Network tab**, alongside *Network graph* and *Keyring*, instead of a new top-level page. The pending-request count is surfaced in the tab label (e.g. `Friend Requests (3)`) rather than a sidebar badge.

No changes needed on the libretroshare side (#306) — this was purely a GUI placement change. I've pushed the updated branch (commit 3d7e612).

I went with the Network sub-tab over the friend-list entry to keep the change self-contained and avoid touching the existing friend-list model, but happy to explore that direction in a follow-up if you'd prefer it there long-term.

defnax:
This branch has conflicts you can solve it? rebase? @64johnlee

PR #3221, opened 2026-05-09
Fix FeedReader: truncate oversized feed messages to prevent deserialization failure on startup

jolavillette:
Fix FeedReader: truncate oversized feed messages to prevent deserialization failure on startup

Issue Summary: Oversized RSS Feed Posts Disappearing on Startup

Bug
In FeedReader plugin subscribe to https://ubuntu.com/blog/feed
Notice this post: https://ubuntu.com//blog/canonical-releases-ubuntu-26-04-lts-resolute-raccoon
Notice that after restart this post is missing, and after manual or automatic update it appears again as new

Root Cause
When an RSS feed post is exceptionally large (e.g., > 256 KB due to rich HTML content or embedded base64 images), it is successfully serialized and written to feedreader.cfg at shutdown. However, during startup, the deserializer rejects any item exceeding RsSerialiser::MAX_SERIAL_SIZE (strictly capped at 256 KB for network safety limits), throwing a deserialization failure and discarding the post. Since the post is missing from memory on startup, RetroShare re-downloads it as a brand-new unread post upon the next feed update, creating an endless cycle of disappearing and reappearing unread posts.

Solution
In RsFeedReaderSerialiser::sizeMsg, we now check if the serialized message size exceeds the MAX_SERIAL_SIZE limit. If it does, we cleanly truncate the description (or descriptionTransformed) so that the item fits perfectly under the safety limit, and prepend a stylized warning informing the user:

[Warning: This message has been truncated by RetroShare because its size exceeded the 256 KB safety limit.]

This ensures the post is successfully loaded on startup, preventing it from disappearing and endlessly resetting its read/unread status.

### DIFF
--- a/plugins/FeedReader/services/p3FeedReader.cc
+++ b/plugins/FeedReader/services/p3FeedReader.cc
@@ -1314,6 +1314,7 @@ bool p3FeedReader::retransformMsg(uint32_t feedId, const std::string &msgId)
 		std::string errorString;
 		std::string descriptionTransformed = mi->descriptionTransformed;
 		if (p3FeedReaderThread::processTransformation(*fi, mi, errorString) == RS_FEED_ERRORSTATE_OK) {
+			RsFeedReaderSerialiser::correctMsgSize(mi);
 			if (mi->descriptionTransformed != descriptionTransformed) {
 				msgChanged = true;
 			}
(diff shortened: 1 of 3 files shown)


csoler (changes requested):
This file is using the "old style" serialization system. We can keep it that way, since it's in a plugin anyway.

The problem in this PR is that sizeMsg() which is responsible for computing the serialized size, is also used to modify the content of the msg so as to fit the serialization requirements. This is really inconsistent.

I propose that you do this: move the lines you added into sizeMsg() into a new method correctMsgSize() that actually modifies the msg is needed. Then call it when necessary. That will remove a dangerous ambiguity.

PR #3224, opened 2026-05-14
Fix FeedReader text truncation on Windows MSYS2 by forcing UTF-8 and improving XML extraction

jolavillette:
Fix FeedReader text truncation on Windows MSYS2 by forcing UTF-8 and improving XML extraction

Description:
This PR fixes a bug where FeedReader messages were being truncated when containing special characters, most notably smart quotes (curly quotes).
The bug exists on all Windows MSYS2 build, including recent builds published by Thunder on his dedicated channel
When a post has a "<anything here>" sequence it is troncated just before the second " and �� are displayed instead
Check feed ubuntu.com/blog/feed at post https://ubuntu.com//blog/rethinking-byod-security-protecting-data-without-trusting-devices
The issue was specifically triggered in Windows MSYS2 builds, where libxml2 failed to correctly detect the encoding when left unspecified. This caused the parser to abort immediately upon encountering non-ASCII UTF-8 bytes, resulting in truncated posts.

Key Changes:
Forced UTF-8 encoding in HTMLWrapper to ensure libxml2 correctly processes special characters regardless of the system's locale/environment.
Improved XML text extraction in XMLWrapper by aggregating all child text nodes instead of just the first one, ensuring full content integrity.
Cleaned up string conversion logic to prevent memory leaks and unnecessary encoding round-trips.

### DIFF
--- a/plugins/FeedReader/util/XMLWrapper.cpp
+++ b/plugins/FeedReader/util/XMLWrapper.cpp
@@ -327,6 +327,10 @@ bool XMLWrapper::getChildText(xmlNodePtr node, const char *childName, std::strin
 		return nodeDump(div, text, true);
 	}
 
+	if (child->children->type != XML_TEXT_NODE) {
+		return false;
+	}
+
 	return getContent(child, text, false);
 }
 
--- a/plugins/FeedReader/util/XMLWrapper.cpp
+++ b/plugins/FeedReader/util/XMLWrapper.cpp
@@ -124,42 +124,24 @@ void XMLWrapper::attach(xmlDocPtr document)
 
 bool XMLWrapper::convertToString(const xmlChar *xmlText, std::string &text)
 {
-	bool result = false;
-
-	xmlBufferPtr in = xmlBufferCreateStatic((void*) xmlText, xmlStrlen(xmlText));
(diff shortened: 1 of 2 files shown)


plugins/FeedReader/util/XMLWrapper.cpp
```diff
@@ -345,15 +327,7 @@ bool XMLWrapper::getChildText(xmlNodePtr node, const char *childName, std::strin
...
-	if (child->children->type != XML_TEXT_NODE) {
-		return false;
-	}
-
-	if (child->children->content) {
-		return convertToString(child->children->content, text);
-	}
-
-	return true;
+	return getContent(child, text, false);
```
csoler, line 330: Is getContent() compatible with nodes of type != XML_TEXT_NODE ?

csoler: I think it's undefined behavior that depends on the implementation. So you'd better keep the test.

csoler (changes requested):

plugins/FeedReader/util/XMLWrapper.cpp
csoler: I was talking of child>type here.

PR #3225, opened 2026-05-14
Distant VOIP

jolavillette:
Implementation of Anonymous VOIP Video over Turtle Tunnels

This PR compiles and works under Ubuntu, Windows MSYS2 and macOS 

It is NOT compatible with current master, it requires cmake build PR/3229

This work enables the ability to stream VOIP video during anonymous chat sessions by routing data through Turtle tunnels.

### DIFF
--- a/plugins/VOIP/CMakeLists.txt
+++ b/plugins/VOIP/CMakeLists.txt
@@ -118,6 +118,18 @@ if(WIN32)
     target_link_libraries(VOIP PRIVATE retroshare retroshare-gui)
 endif()
 
+# pkg-config gives the FFmpeg/speex libs by name only (avcodec, avutil, ...);
+# their directory must be added explicitly or the linker won't find them when
+# they live outside the default search path (e.g. Homebrew /opt/homebrew/... on
+# macOS -> "ld: library 'avcodec' not found"). Harmless on Linux/Windows where
+# these dirs are already on the default path.
+target_link_directories(VOIP PRIVATE
+    ${AVCODEC_LIBRARY_DIRS}
+    ${AVUTIL_LIBRARY_DIRS}
+    ${SPEEX_LIBRARY_DIRS}
+    ${SPEEXDSP_LIBRARY_DIRS}
+)
+
 # NOTE: the base VOIP plugin makes no direct use of OpenSSL (it relies on
 # libretroshare for any cryptography). The optional RsTurtleVOIPBridge is the
 # exception: it calls OpenSSL DH directly, so we link OpenSSL::Crypto only when
--- a/plugins/VOIP/CMakeLists.txt
+++ b/plugins/VOIP/CMakeLists.txt
@@ -35,6 +35,19 @@ set(VOIP_SOURCES
     gui/VOIPToasterNotify.cpp
 )
 
+# RsTurtleVOIPBridge (anonymous / distant VOIP over turtle tunnels) calls OpenSSL
+# Diffie-Hellman (DH_*, BN_*) directly to build the session key, so when it is
+# compiled in the plugin must link libcrypto itself (on Windows every symbol of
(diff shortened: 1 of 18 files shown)


zapek:
This is overkill. Why do you need a "bridge" (new service)? Why another custom DH implementation? How do you handle the latency problem?

jolavillette:
Changes

Build (CMake)
- 7102e8ff6 — compile RsTurtleVOIPBridge in the CMake build (guarded by if(EXISTS)), originally linking OpenSSL::Crypto for the DH code (fixes Windows/MSYS2 undefined reference to DH_*). (This OpenSSL link is later removed — see transport rework.)
- 20cc7104b — add target_link_directories for FFmpeg/speex so the linker finds Homebrew libs on macOS (ld: library 'avcodec' not found). No‑op on Linux/Windows.

macOS camera capture
- 498abbe22 — replace QVideoProbe (on macOS/AVFoundation it reports active but never delivers a frame → LED on, no image, and it suppressed the fallback) with a QAbstractVideoSurface viewfinder that reliably receives frames on all backends. Fixes "no video on macOS".

Transport & security (core change)
- eedc8f853 — route distant VOIP through libretroshare's p3GxsTunnelService instead of raw turtle + the dead DH. The bridge becomes an RsGxsTunnelService::RsGxsTunnelClientService; all cryptography (real DH + AES + HMAC + GXS‑id authentication) is delegated to libretroshare. The media is now end‑to‑end encrypted and the peer authenticated (previously cleartext, visible to turtle relays). Plugin no longer links OpenSSL. Public API and on‑the‑wire envelope unchanged → p3VOIP and the GUI are untouched. Net −137 lines.
- 6b47d14c3 — fix: the GXS tunnel id is derived only from the (from,to) identity pair, not the service id. When a distant chat tunnel to the same identity already exists, the VOIP request reuses it and gxstunnel does not re‑fire CAN_TALK, so the bridge never saw the tunnel become ready and the GUI never sent the call invitation. Fix: poll getTunnelInfo() for readiness instead of relying only on the notification.

UX
- 54651cdbd — mirror the local self‑view (horizontal flip), like mainstream video apps; the frame sent to the peer stays un‑mirrored.

Video quality
- 1d6f79b1f — unblock the encoder bitrate. It was effectively ~10–30 kbps (hardcoded constructor bit_rate, the dynamic bit_rate line commented out, and a missing bytes→bits ×8). Now ~1 Mbps default, ~2 Mbps ceiling, units fixed.

⚠️ Notes for reviewers
- Protocol‑breaking: distant VOIP now rides the GXS tunnel — both endpoints must run this version. Intermediate turtle relays are unaffected (they only forward opaque, now‑encrypted data; no update needed on relays).
- Security improvement: distant VOIP media is now end‑to‑end encrypted (AES + HMAC, GXS‑id authenticated). Previously only hop‑by‑hop friend TLS protected it — relays could observe it.
- The available bw field in the VOIP config panel is a preview‑only simulator and does not affect live calls.

G10h4ck:
This part is incorrect

> Security improvement: distant VOIP media is now end‑to‑end encrypted (AES + HMAC, GXS‑id authenticated). Previously only hop‑by‑hop friend TLS protected it — relays could observe it.

The VoIP plugin worked only over direct connection, so there was no relay that could observe

jolavillette:
> This part is incorrect
> 
> > Security improvement: distant VOIP media is now end‑to‑end encrypted (AES + HMAC, GXS‑id authenticated). Previously only hop‑by‑hop friend TLS protected it — relays could observe it.
> 
> The VoIP plugin worked only over direct connection, so there was no relay that could observe

it is a misunderstanding
- indeed the original VOIP node to node plugin was safe.
- the first version of the gxsid to gxsid VOIP PR was unsafe: there was no end to end encryption, intermediate nodes could intercept the communication, that was fixed in today's commits

PR #3229, opened 2026-06-02
C make migration modular cross platform

jolavillette:
see https://github.com/RetroShare/libretroshare/pull/304


plugins/FeedReader/services/p3FeedReader.cc
```diff
@@ -2604,13 +2604,13 @@ bool p3FeedReader::updatePostedGroup(const RsPostedGroup &postedGroup, const std
 	return true;
 }
 
-bool p3FeedReader::waitForToken(RsGxsIfaceHelper *interface, uint32_t token)
+bool p3FeedReader::waitForToken(RsGxsIfaceHelper *helper, uint32_t token)
```
G10h4ck: Totally unrelated!

plugins/FeedReader/services/p3FeedReader.h
```diff
@@ -106,7 +106,7 @@ class p3FeedReader : public RsPQIService, public RsFeedReader
 	bool updateForumGroup(const RsGxsForumGroup &forumGroup, const std::string &groupName, const std::string &groupDescription);
 	bool getPostedGroup(const RsGxsGroupId &groupId, RsPostedGroup &postedGroup);
 	bool updatePostedGroup(const RsPostedGroup &postedGroup, const std::string &groupName, const std::string &groupDescription);
-	bool waitForToken(RsGxsIfaceHelper *interface, uint32_t token);
+	bool waitForToken(RsGxsIfaceHelper *helper, uint32_t token);
```
G10h4ck: Totally unrelated!

plugins/FeedReader/CMakeLists.txt
```diff
@@ -0,0 +1,137 @@
...
+
+set_target_properties(FeedReader PROPERTIES
+    PREFIX "lib"
+    LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/plugins"
+    RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/plugins"
+    AUTOUIC_SEARCH_PATHS "${CMAKE_CURRENT_SOURCE_DIR}/gui"
+)
+
+# --- Links ---
+find_package(OpenSSL REQUIRED)
```
G10h4ck, line 99: Is this a direct dependency of feedreader plugin?

plugins/FeedReader/CMakeLists.txt
```diff
@@ -0,0 +1,137 @@
...
+
+if(WIN32)
+    target_link_libraries(FeedReader PRIVATE retroshare retroshare-gui)
```
G10h4ck, line 107: Is this line needed on windows only?

plugins/VOIP/CMakeLists.txt
```diff
@@ -0,0 +1,141 @@
...
+)
+
+set_target_properties(VOIP PROPERTIES
+    PREFIX "lib"
+    LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/plugins"
+    RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/plugins"
+    AUTOUIC_SEARCH_PATHS "${CMAKE_CURRENT_SOURCE_DIR}/gui"
+)
+
+target_compile_definitions(VOIP PRIVATE __STDC_CONSTANT_MACROS)
```
G10h4ck, line 96: Why is this line needed? What does it does?

plugins/VOIP/CMakeLists.txt
```diff
@@ -0,0 +1,141 @@
...
+if(WIN32)
+    target_compile_definitions(VOIP PRIVATE WINDOWS_SYS)
+endif()
+
+# --- Links ---
+find_package(OpenSSL REQUIRED)
```
G10h4ck: Does it requires openssl directly, or does it just uses encryption from libretroshare?

retroshare-friendserver/CMakeLists.txt
```diff
@@ -0,0 +1,80 @@
...
+	"Disable optimization to speed up build, enable verbose build log"
+	OFF )
+
+option(
+	RS_RNPLIB
+	"Enable use RNP lib for PGP"
+	ON )
+
+find_package(ZLIB REQUIRED)
+find_package(BZip2 REQUIRED)
```
G10h4ck: do friendserver does direct usage of ZLIB and BZip2? If it is because of libretroshare usage we don't need those here

retroshare-friendserver/CMakeLists.txt
```diff
@@ -0,0 +1,80 @@
...
+	src/network.cc
+)
+
+add_executable(${PROJECT_NAME} ${FRIENDSERVER_SOURCES})
+set_target_properties(${PROJECT_NAME} PROPERTIES CXX_STANDARD 17)
+target_compile_options(${PROJECT_NAME} PRIVATE -Wno-deprecated-declarations)
+
+
+# Find libretroshare
+if(NOT TARGET retroshare)
```
G10h4ck, line 55: this seems a bit strange here, why is this if needed?

retroshare-friendserver/CMakeLists.txt
```diff
@@ -0,0 +1,80 @@
...
+	endif()
+endif()
+
+# Include directories
+target_include_directories(${PROJECT_NAME} PRIVATE
+	"${CMAKE_CURRENT_SOURCE_DIR}/src"
+)
+
+# Link libraries
+find_package(OpenSSL REQUIRED)
```
G10h4ck, line 85: again doe it makes direct usage of openssl or it is because of libretroshare? in the second case we should't depend on openssl directly here

retroshare-friendserver/CMakeLists.txt
```diff
@@ -0,0 +1,80 @@
...
+target_link_libraries(${PROJECT_NAME} PRIVATE retroshare OpenSSL::SSL OpenSSL::Crypto)
+
+if(WIN32)
+	target_link_libraries(${PROJECT_NAME} PRIVATE ws2_32 winmm iphlpapi mswsock)
```
G10h4ck: Same as openssl comment

retroshare-gui/src/gui/msgs/MessageComposer.cpp
```diff
@@ -267,24 +267,24 @@ MessageComposer::MessageComposer(QWidget *parent, Qt::WindowFlags flags)
     actionAlignJustify->setShortcut(Qt::CTRL + Qt::Key_J);
     actionAlignJustify->setCheckable(true);
     
-    QActionGroup *grp2 = new QActionGroup(this);
-    connect(grp2, SIGNAL(triggered(QAction*)), this, SLOT(textStyle(QAction*)));
+    QActionGroup *grpStyle = new QActionGroup(this);
```
G10h4ck: Totally unrelated move it to another PR

retroshare-gui/CMakeLists.txt
```diff
@@ -68,32 +73,39 @@ cmake_dependent_option(
...
 set(CMAKE_AUTORCC ON)
-set(CMAKE_AUTOUIC ON)
-
-find_package( Qt6 COMPONENTS Core         REQUIRED)
-find_package( Qt6 COMPONENTS Widgets      REQUIRED)
-find_package( Qt6 COMPONENTS Xml          REQUIRED)
-find_package( Qt6 COMPONENTS Network      REQUIRED)
-find_package( Qt6 COMPONENTS Multimedia   REQUIRED)
-find_package( Qt6 COMPONENTS PrintSupport REQUIRED)
+set(CMAKE_AUTOUIC OFF)
```
G10h4ck, line 99: This seems strange...

retroshare-gui/CMakeLists.txt
```diff
@@ -33,9 +33,11 @@ option(
 	 just for development purposes, not suitable for library usage"
 	 ON )
 
+option(RS_JSON_API "Use restbed to expose libretroshare as JSON API via HTTP" ON)
+
 option(
-	RS_JSON_API
-	"Use restbed to expose libretroshare as JSON API via HTTP"
+	RS_USE_NATIVE_DIALOGS
+	"Enable native dialogs (replaces CONFIG+=rs_use_native_dialogs)"
```
G10h4ck: no need of `(replaces CONFIG+=rs_use_native_dialogs)` if some more explaination is neede dplease add it here verbatim not referring to the deprecated qmake

retroshare-gui/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@
 ################################################################################
 
 cmake_minimum_required (VERSION 3.18.0)
-project(retroshare-gui VERSION 0.6.6 LANGUAGES CXX)
+project(retroshare-gui VERSION 0.6.6 LANGUAGES CXX C)
```
G10h4ck: do you mean 0.6.7 ?

retroshare-gui/CMakeLists.txt
```diff
@@ -106,7 +118,7 @@ find_package(Git REQUIRED)
 ################################################################################
 
 if(RS_DEVELOPMENT_BUILD)
-	set(CMAKE_VERBOSE_MAKEFILE ON)
+	set(CMAKE_VERBOSE_MAKEFILE OFF)
```
G10h4ck: Restore it to ON

retroshare-gui/CMakeLists.txt
```diff
@@ -119,10 +131,64 @@ if (NOT Qt6_FOUND)
 	qt5_wrap_ui(RS_UI_HEADERS ${RS_GUI_FORMS})
 endif()
 
-add_executable(${PROJECT_NAME} ${RS_GUI_SOURCES} ${RS_UI_HEADERS} ${RS_GUI_QTRESOURCES})
+# Generate include/qmake_info.h in the binary directory so that CMake compiles can find it
```
G10h4ck: This blocks is a bit smelly, why do we need such a thing??

G10h4ck (changes requested):
I have reviewed half of it, see inline comments, please move all unrelated stuff to other PR

plugins/FeedReader/services/p3FeedReader.cc
jolavillette: This one is actually not unrelated: interface is a reserved macro in the Windows SDK (<objbase.h>, pulled in transitively by <windows.h>), expanded to struct. A parameter named interface therefore fails to compile this plugin on Windows/MinGW. I've kept the rename and added a comment explaining why. Happy to move it to a dedicated "Windows portability" PR if you'd prefer it out of the build PR — just let me know.

plugins/FeedReader/services/p3FeedReader.h
jolavillette: Same answer as above

plugins/FeedReader/CMakeLists.txt
jolavillette: Yes, it's a direct dependency: services/p3FeedReaderThread.cc does #include <openssl/evp.h> and uses the EVP API directly (not only through libretroshare). Kept + commented.

plugins/FeedReader/CMakeLists.txt
jolavillette: Yes, Windows-only. A plugin DLL (PE) must resolve every symbol at link time, so it has to link explicitly against the libraries that provide them. On ELF (Linux) the loader resolves those symbols from the hosting retroshare-gui process at load time, so the explicit link isn't needed there. Kept + commented.

plugins/VOIP/CMakeLists.txt
jolavillette: Required by the FFmpeg/libav headers (libavcodec/libavutil, included from gui/VideoProcessor.*). They use the C99 constant macros (INT64_C, UINT64_C, …), which <stdint.h> only exposes in a C++ TU when __STDC_CONSTANT_MACROS is defined before inclusion. Kept + commented.

plugins/VOIP/CMakeLists.txt
jolavillette: No direct use — VOIP relies on libretroshare for any crypto. Removed find_package(OpenSSL) and the OpenSSL::SSL/Crypto links.

retroshare-friendserver/CMakeLists.txt
jolavillette: No direct use (they weren't even linked, only searched). Removed both. They come transitively through libretroshare.

retroshare-friendserver/CMakeLists.txt
jolavillette: Guard against adding libretroshare twice: when the friendserver is configured as part of the top-level build, the retroshare target already exists (added once via add_subdirectory), and adding the same subdirectory again is a hard CMake error. Standalone, the target doesn't exist, so we add/fetch it here. Kept + commented.

retroshare-friendserver/CMakeLists.txt
jolavillette: Same as #6: no direct use. Removed; provided transitively by the retroshare target.

plugins/FeedReader/services/p3FeedReader.cc
G10h4ck: is this still needed also with `WIN32_LEAN_AND_MEAN`?

plugins/FeedReader/services/p3FeedReader.h
G10h4ck: is this still needed also with `WIN32_LEAN_AND_MEAN`?

plugins/FeedReader/CMakeLists.txt
G10h4ck: ok

plugins/FeedReader/CMakeLists.txt
G10h4ck: ok

retroshare-friendserver/CMakeLists.txt
jolavillette: Partially the same: the friendserver does use the Winsock API directly (src/network.cc: socket/bind/listen/accept/setsockopt/…), so I kept ws2_32 explicitly ("link what you use"). winmm and iphlpapi are already exported PUBLIC by libretroshare, and mswsock isn't used here, so I removed those three.

retroshare-gui/src/gui/msgs/MessageComposer.cpp
jolavillette: Agreed, unrelated. Reverted — the file is back to its original state and no longer appears in this PR. (I also reverted the same kind of cosmetic rename in RsNetUtil.cpp (lst2 → lstSlash), which wasn't flagged yet.)

retroshare-gui/CMakeLists.txt
jolavillette: Intentional: the .ui forms are collected in src/CMakeLists.txt (RS_GUI_FORMS) and processed explicitly via qt_wrap_ui/qt5_wrap_ui, which emit the ui_*.h into the build dir (already on the include path). Leaving AUTOUIC on would process the same forms a second time. Added a comment.

retroshare-gui/CMakeLists.txt
jolavillette: Reworded with no reference to qmake: now describes the actual effect (native OS file/font/directory dialogs vs Qt's own DontUseNativeDialog ones).

retroshare-gui/CMakeLists.txt
jolavillette: Yes — corrected to 0.6.7 (matches the embedded qmake_info.h: RS_MAJOR_VERSION=0 RS_MINOR_VERSION=6 RS_MINI_VERSION=7).

retroshare-gui/CMakeLists.txt
jolavillette: Restored to ON.

retroshare-gui/CMakeLists.txt
jolavillette: Agreed it's smelly. It exists because src/gui/AboutWidget.cpp unconditionally does #include "include/qmake_info.h" and prints Rs_qmake_DEFINES/Rs_qmake_CONFIG in the About dialog; that header is normally produced by the qmake build and is git-ignored, so a CMake build has nothing to include. For now it's a minimal shim to keep AboutWidget compiling; I added a comment + a TODO to either generate it from the real CMake configuration (configure_file) or refactor AboutWidget to drop the qmake-specific header. I can do that follow-up in a separate change.

retroshare-gui/src/util/RsNetUtil.cpp
```diff
@@ -47,14 +47,14 @@ bool RsNetUtil::parseAddrFromQString(const QString& s, struct sockaddr_storage&
 	uint32_t s2 = (*it).toInt(&ok) ; ++it ; if(!ok) return false ; if(it ==lst.end()) return false ;
 	uint32_t s3 = (*it).toInt(&ok) ; ++it ; if(!ok) return false ; if(it ==lst.end()) return false ;
 
-	QStringList lst2 = (*it).split("/") ;
+	QStringList lstSlash = (*it).split("/") ;
```
G10h4ck: is this still needed also with `WIN32_LEAN_AND_MEAN`? Anyway it seems better like this then before

retroshare-gui/CMakeLists.txt
```diff
@@ -68,32 +73,39 @@ cmake_dependent_option(
...
-
-find_package( Qt6 COMPONENTS Core         REQUIRED)
-find_package( Qt6 COMPONENTS Widgets      REQUIRED)
-find_package( Qt6 COMPONENTS Xml          REQUIRED)
-find_package( Qt6 COMPONENTS Network      REQUIRED)
-find_package( Qt6 COMPONENTS Multimedia   REQUIRED)
-find_package( Qt6 COMPONENTS PrintSupport REQUIRED)
+set(CMAKE_AUTOUIC OFF)
+
+find_package( Qt5 COMPONENTS Core         REQUIRED)
```
G10h4ck, line 101: Oh no please, now that we are at it, please do all of this with Qt6, Qt5 is EOL, and RS got removed from some distributions because it still depends on EOL Qt5

retroshare-gui/CMakeLists.txt
```diff
@@ -119,10 +131,64 @@ if (NOT Qt6_FOUND)
...
+	# but keep it visible for Debug and Development builds to view stdout/stderr logs.
+	if(NOT RS_DEVELOPMENT_BUILD AND NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
+		set(GUI_TYPE WIN32)
+	endif()
+elseif(APPLE)
+	set(GUI_TYPE MACOSX_BUNDLE)
+endif()
+
+add_executable(${PROJECT_NAME} ${GUI_TYPE} ${RS_GUI_SOURCES} ${RS_IMPLEMENTATION_HEADERS} ${RS_UI_HEADERS} ${RS_GUI_QTRESOURCES})
+set_target_properties(${PROJECT_NAME} PROPERTIES RUNTIME_OUTPUT_NAME "retroshare")
```
G10h4ck: I think the exectuable for RetroShare GUI should be named retroshare-gui on all platforms

retroshare-gui/CMakeLists.txt
```diff
@@ -218,25 +294,18 @@ if(RS_GUI_CMARK)
...
-		libsam3
-		GIT_REPOSITORY "https://github.com/i2p/libsam3.git"
-		GIT_TAG "origin/master"
-		GIT_SHALLOW TRUE
-		GIT_PROGRESS TRUE
-		TIMEOUT 10
-		)
-	FetchContent_MakeAvailable(libsam3)
+if(RS_USE_I2P_SAM3)
+	target_link_libraries(${PROJECT_NAME} PRIVATE sam3)
```
G10h4ck, line 336: Is `sam3` used directly by retroshare-gui?

retroshare-gui/CMakeLists.txt
```diff
@@ -291,21 +360,25 @@ endif()
 if(UNIX)
 	find_package(PkgConfig REQUIRED)
 
-	pkg_check_modules(X11 REQUIRED x11)
-	pkg_check_modules(XSCRNSAVER REQUIRED xscrnsaver)
+	if(NOT APPLE)
+		pkg_check_modules(REQUIRED x11)
+		pkg_check_modules(X11 REQUIRED x11)
+		pkg_check_modules(XSCRNSAVER REQUIRED xscrnsaver)
```
G10h4ck, line 398: I remember we make use of this things, but for what? Is this still useful nowadays? With all the compositors and wayland boiling around ?

retroshare-gui/CMakeLists.txt
```diff
@@ -291,21 +360,25 @@ endif()
...
-	target_include_directories(retroshare-gui PRIVATE ${XSCRNSAVER_INCLUDE_DIRS})
-	target_compile_options(retroshare-gui PRIVATE ${XSCRNSAVER_CFLAGS_OTHER})
+		target_include_directories(retroshare-gui PRIVATE ${XSCRNSAVER_INCLUDE_DIRS})
+		target_compile_options(retroshare-gui PRIVATE ${XSCRNSAVER_CFLAGS_OTHER})
+
+		target_compile_definitions(${PROJECT_NAME} PRIVATE HAVE_XSS)
+	endif()
 
-	target_compile_definitions(${PROJECT_NAME} PRIVATE HAVE_XSS)
 	target_compile_definitions(${PROJECT_NAME} PRIVATE _FILE_OFFSET_BITS=64)
```
G10h4ck, line 415: Do retroshare-gui uses this directly or it just rely on libretroshare for files operations?

retroshare-gui/CMakeLists.txt
```diff
@@ -490,16 +563,35 @@ endif(RS_SANITIZE)
 
 target_link_libraries(${PROJECT_NAME} PRIVATE ${RS_LINK_LIBRARIES})
 
-target_include_directories(${PROJECT_NAME} PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/src/)
 
-set( CMAKE_CXX_FLAGS "-Wno-deprecated-declarations" )
+target_compile_options(${PROJECT_NAME} PRIVATE -Wno-deprecated-declarations)
```
G10h4ck: I think this kind of settings should be controlled by an option like I did in libretroshare `CMakeLists.txt`

retroshare-gui/CMakeLists.txt
```diff
@@ -490,16 +563,35 @@ endif(RS_SANITIZE)
...
 target_compile_definitions(${PROJECT_NAME} PUBLIC  RS_NO_WARN_DEPRECATED )
 target_compile_definitions(${PROJECT_NAME} PRIVATE RS_RELEASE_VERSION )
 target_compile_definitions(${PROJECT_NAME} PRIVATE TARGET=\"retroshare\")
```
G10h4ck, line 612: RetroShare GUI executable should be named retroshare-gui on all platforms

jolavillette:
Compiles OK on ubuntu, win msys2, macos

retroshare-gui/CMakeLists.txt
```diff
@@ -490,16 +563,35 @@ endif(RS_SANITIZE)
...
+target_compile_definitions(${PROJECT_NAME} PRIVATE RS_DIRECT_CHAT)
 
 if(RS_GXS_CIRCLES)
 	target_compile_definitions(${PROJECT_NAME} PRIVATE RS_USE_CIRCLES )
 endif(RS_GXS_CIRCLES)
 
-#add_dependencies(${PROJECT_NAME} libretroshare)
+find_library(BOTAN_LIBRARY NAMES botan botan-3 botan-2 libbotan-3 libbotan-2 REQUIRED)
```
G10h4ck, line 626: Do RetroShare GUI uses botan directly? I quess it is rnplib who needs it, so no need for this line here

retroshare-gui/CMakeLists.txt
```diff
@@ -490,16 +563,35 @@ endif(RS_SANITIZE)
...
 
+if(WIN32)
+    # Robust resolution including RNP and sexpp inside the link group for Windows MinGW
+    target_link_libraries(${PROJECT_NAME} PRIVATE -Wl,--start-group retroshare librnp sexpp z bz2 json-c ${BOTAN_LIBRARY} -Wl,--end-group)
```
G10h4ck, line 630: who need json-c? We have rapidjson for direct usage in RetroShare

build_retroshare.sh
```diff
@@ -0,0 +1,57 @@
...
+# Detect Qt5 path on macOS if Homebrew is installed
+QT_PREFIX=""
+if command -v brew &> /dev/null; then
+    QT_PREFIX=$(brew --prefix qt@5)
+fi
+
+echo "=== Starting RetroShare Compilation ==="
+
+# 1. Compile libretroshare (core library) and librnp automatically
+echo ">>> Step 1: Compiling libretroshare..."
```
G10h4ck: This step should not be needed anymore, CMake will deal with it by itself

CMakeLists.txt
```diff
@@ -0,0 +1,88 @@
...
+set(CMAKE_ENABLE_EXPORTS ON)
+
+# Configuration Options
+option(RS_GUI "Build RetroShare GUI" ON)
+option(RS_SERVICE "Build RetroShare Service" ON)
+option(RS_FRIENDSERVER "Build RetroShare Friendserver" ON)
+option(RS_PLUGINS "Build RetroShare Plugins" OFF)
+option(RS_USE_I2P_SAM3 "Enable I2P SAMv3 support" ON)
+
+if(RS_USE_I2P_SAM3)
```
G10h4ck, line 39: I think each target that uses `sam3` such as libretroshare will deal with this automatically, please don't duplicate code if unnecessary ;-)

CMakeLists.txt
```diff
@@ -0,0 +1,88 @@
...
+		if(WIN32)
+			target_link_libraries(sam3 PUBLIC ws2_32)
+		endif()
+	endif()
+endif()
+
+# Global configurations for Windows builds
+if(WIN32)
+	# Force Unicode mappings globally for all Windows targets (fixes MoveFileEx in libbitdht)
+	add_definitions(-DUNICODE -D_UNICODE)
```
G10h4ck: This should be dealt with in each target not here

CMakeLists.txt
```diff
@@ -0,0 +1,88 @@
...
+
+	# Remove multiple definitions hack for plugins linking to static libraries exported by retroshare-gui
+
+	# Downgrade strict GCC 15 errors to warnings for old submodules
+	if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
```
G10h4ck: Each module that requires such hacks should to it in their own CMakeList.txt

build_scripts/OSX/deploy-macos.sh
```diff
@@ -0,0 +1,197 @@
...
+    else
+        echo "WARNING: hdiutil not found, cannot create .dmg."
+    fi
+else
+    echo "  WARNING: macdeployqt tool not found in PATH! Make sure Qt is installed and its bin folder is in PATH."
+    echo "  Cannot bundle Qt dependencies or create DMG automatically."
+    echo "================================================================================"
+    echo "SUCCESS: App bundle generated at: ${TARGET_APP}"
+    echo "================================================================================"
+fi
```
G10h4ck, line 224: This whole file looks like a hell to maintain, it seems strange that in 2026 to package an app for MacOS all this trickery is needed... please double check it

build_scripts/Windows-msys2/deploy-windows.sh
```diff
@@ -0,0 +1,363 @@
...
+        echo "================================================================================"
+        echo "SUCCESS: Standalone package created at: ${BUILD_DIR}/${ARCHIVE_BASE}.zip"
+        echo "================================================================================"
+    else
+        echo "================================================================================"
+        echo "SUCCESS: Folder generated at: ${DEPLOY_DIR}"
+        echo "NOTE: Neither '7z' nor 'zip' commands were found. Archive creation skipped."
+        echo "================================================================================"
+    fi
+)
```
G10h4ck, line 374: This whole file looks like a hell to maintain, it seems strange that in 2026 to package an app for Windows all this trickery is needed... please double check it

Makefile
```diff
@@ -0,0 +1,304 @@
...
+	@echo "  RS_USE_I2P_SAM3       Enable I2P SAMv3 support (default: ON)"
+	@echo "  RS_BITDHT             BitTorrent-DHT peer discovery (default: ON)"
+	@echo "  RS_MINIUPNPC          NAT/UPnP port forwarding via miniupnpc (default: ON)"
+	@echo "  RS_BRODCAST_DISCOVERY LAN peer discovery (default: ON)"
+	@echo "  RS_SQLCIPHER          SQLCipher encryption for the GXS database (default: ON)"
+	@echo "  RS_V07_BREAKING_CHANGES  Enable 0.7.0 retro-incompatible changes (default: OFF)"
+	@echo "  CMAKE_BUILD_TYPE      Debug/Release/RelWithDebInfo/MinSizeRel (default: Release)"
+	@echo "  NPROC                 Number of build threads (default: detected CPUs)"
+	@echo "  CMAKE_FLAGS           Extra flags appended to the cmake configure line"
+	@echo "=============================================================================="
```
G10h4ck: Oh no please not another wrapper, CMake is used to avoid writing makefiles by ourselves, so please doesn't add more Makefiles i believe we should't need them at all...

G10h4ck (changes requested):
See inline comments

plugins/VOIP/CMakeLists.txt
G10h4ck: ok

G10h4ck (changes requested):
See inline comments

retroshare-gui/src/util/RsNetUtil.cpp
jolavillette: 1 and 2. p3FeedReader interface rename / WIN32_LEAN_AND_MEAN: You're right. libretroshare defines WIN32_LEAN_AND_MEAN PUBLIC, the plugin inherits it on Windows, so <windows.h> no longer pulls in the interface macro. Reverted — the rename is gone and these files leave the PR. Confirmed building on Windows.

3 . RsNetUtil.cpp lst2/lstSlash: unrelated to WIN32_LEAN_AND_MEAN (not a macro). Already reverted to the original lst2.

retroshare-gui/CMakeLists.txt
jolavillette: Qt5 → Qt6: I'd keep this PR as a 1:1 qmake→CMake migration (same deps, including Qt5) so it stays verifiable as "same binary". Qt5 EOL is real — I'll do the Qt6 port in a dedicated follow-up PR (this migration is its prerequisite).

retroshare-gui/CMakeLists.txt
jolavillette: executable naming: GUI executable is now retroshare-gui on all platforms (RUNTIME_OUTPUT_NAME, + macOS CFBundleExecutable). Note: the TARGET define is not the executable name — it's the single-instance IPC socket name (main.cpp/rshare.cpp), left unchanged on purpose (commented).

retroshare-gui/CMakeLists.txt
jolavillette: sam3: used directly by the GUI (ServerPage.cpp: <libsam3.h>, sam3CloseConnection()). Kept + commented.

retroshare-gui/CMakeLists.txt
jolavillette: X11 / XScreenSaver: used directly for idle detection (idle/idle_platform.cpp, HAVE_XSS). Kept + commented. It only works under X11/XWayland — modernising idle detection for Wayland is a separate effort.

retroshare-gui/CMakeLists.txt
jolavillette: _FILE_OFFSET_BITS=64: large-file-support flag, no-op on 64-bit; kept for consistency with libretroshare (commented).

retroshare-gui/CMakeLists.txt
jolavillette: -Wno-deprecated-declarations: now gated behind an RS_WARN_DEPRECATED option, mirroring libretroshare/CMakeLists.txt.

retroshare-gui/CMakeLists.txt
jolavillette: executable naming: GUI executable is now retroshare-gui on all platforms (RUNTIME_OUTPUT_NAME, + macOS CFBundleExecutable). Note: the TARGET define is not the executable name — it's the single-instance IPC socket name (main.cpp/rshare.cpp), left unchanged on purpose (commented).

retroshare-gui/CMakeLists.txt
jolavillette: botan / json-c: not used directly by the GUI (it uses rapidjson). They're transitive deps of librnp, linked here only because static librnp doesn't propagate them. Proper fix = export them from librnp/libretroshare; left as a libretroshare follow-up (commented in place).

retroshare-gui/CMakeLists.txt
jolavillette: botan / json-c: not used directly by the GUI (it uses rapidjson). They're transitive deps of librnp, linked here only because static librnp doesn't propagate them. Proper fix = export them from librnp/libretroshare; left as a libretroshare follow-up (commented in place).

build_retroshare.sh
jolavillette: build_retroshare.sh: removed from the repo (CMake builds everything itself).

CMakeLists.txt
jolavillette: top-level sam3 definition: defined once at the top because it's linked by both libretroshare and the GUI (direct use, #6); a single definition avoids a duplicate-target error. Commented. Can be dropped if libretroshare exports sam3.

CMakeLists.txt
jolavillette: global UNICODE/_UNICODE: moved to a per-target definition on bitdht.

CMakeLists.txt
jolavillette: GCC ≥ 15 error downgrade: moved from global to the specific legacy C submodule targets.

build_scripts/OSX/deploy-macos.sh
jolavillette: deploy-macos.sh / deploy-windows.sh: deployment is a first-class step (it was in the qmake scripts too), so kept. Double-checked: the core work is macdeployqt/windeployqt; the surrounding glue handles RetroShare specifics those tools don't cover (macOS .dylib plugin requirement + Apple-Silicon signing; MSYS2 plugin/runtime-DLL gaps + portable layout). Rationale documented at the top of each script.

build_scripts/Windows-msys2/deploy-windows.sh
jolavillette: deploy-macos.sh / deploy-windows.sh: deployment is a first-class step (it was in the qmake scripts too), so kept. Double-checked: the core work is macdeployqt/windeployqt; the surrounding glue handles RetroShare specifics those tools don't cover (macOS .dylib plugin requirement + Apple-Silicon signing; MSYS2 plugin/runtime-DLL gaps + portable layout). Rationale documented at the top of each script.

Makefile
jolavillette: Makefile: removed from the repo.

CMakeLists.txt
jolavillette: global UNICODE/_UNICODE: moved to a per-target definition on bitdht.

CMakeLists.txt
jolavillette: GCC ≥ 15 error downgrade: moved from global to the specific legacy C submodule targets.

build_scripts/Windows-msys2/deploy-windows.sh
jolavillette: deploy-macos.sh / deploy-windows.sh: deployment is a first-class step (it was in the qmake scripts too), so kept. Double-checked: the core work is macdeployqt/windeployqt; the surrounding glue handles RetroShare specifics those tools don't cover (macOS .dylib plugin requirement + Apple-Silicon signing; MSYS2 plugin/runtime-DLL gaps + portable layout). Rationale documented at the top of each script.

build_scripts/OSX/deploy-macos.sh
jolavillette: deploy-macos.sh / deploy-windows.sh: deployment is a first-class step (it was in the qmake scripts too), so kept. Double-checked: the core work is macdeployqt/windeployqt; the surrounding glue handles RetroShare specifics those tools don't cover (macOS .dylib plugin requirement + Apple-Silicon signing; MSYS2 plugin/runtime-DLL gaps + portable layout). Rationale documented at the top of each script.

Makefile
jolavillette: Makefile: removed from the repo.

retroshare-gui/src/util/RsNetUtil.cpp
jolavillette: 1 ans 2. p3FeedReader interface rename / WIN32_LEAN_AND_MEAN: You're right. libretroshare defines WIN32_LEAN_AND_MEAN PUBLIC, the plugin inherits it on Windows, so <windows.h> no longer pulls in the interface macro. Reverted — the rename is gone and these files leave the PR. Confirmed building on Windows.

3. RsNetUtil.cpp lst2/lstSlash: unrelated to WIN32_LEAN_AND_MEAN (not a macro). Already reverted to the original lst2.

retroshare-gui/src/gui/PluginManager.cpp
```diff
@@ -217,7 +217,8 @@ PluginManager::pluginWidget(QString pluginName)
         QString em = tr("Error: no plugin with name '%1' found")
                         .arg(pluginName);
         emit errorAppeared( em );
-    }      
+    }
+    return 0; // pluginWidget() returns QWidget*: every path must return (else -Werror=return-type)
```
G10h4ck: Better use `nullptr` instead of `0`

retroshare-gui/CMakeLists.txt
G10h4ck: ok

retroshare-gui/CMakeLists.txt
G10h4ck: We should change this in the future, better name for the macro and use retroshare-gui for the IPC socket name too, but please fix that in another PR

G10h4ck (changes requested):

retroshare-gui/src/gui/PluginManager.cpp
jolavillette: done

G10h4ck:
Well done @jolavillette now seems to be merge-able to me, if we have a couple tests from other people before merging that would be even better @csoler  @defnax

G10h4ck (changes requested):
@jolavillette  Please move both windows an macos deploy scripts to build_scripts directory to avoid root dir pollution

jolavillette:
Latest changes (4 commits)
Build-system follow-ups (after dropping the orchestrator Makefile) plus cross-platform build fixes found while validating a full build on Linux, Windows (MSYS2/MinGW) and macOS.

1. build: move deploy scripts under build_scripts/<platform>/
deploy-macos.sh → build_scripts/OSX/deploy-macos.sh
deploy-windows.sh → build_scripts/Windows-msys2/deploy-windows.sh
Each script now cds to the repo root from its own location and reads the build dir from BUILD_DIR (default Build-cmake), so they no longer depend on the Makefile or on the current directory.

2. build(macos): detect Qt version for the deploy archive name
deploy-macos.sh had QT_VERSION="UnknownQt" hardcoded. It now queries qmake (and the keg-only Homebrew qt@5 qmake), like deploy-windows.sh, so the .dmg carries the real Qt version.

3. build(gui): exclude libsam3.a from re-exported symbols (fixes Ninja Windows build)
The GUI links with -Wl,--export-all-symbols; libsam3.a was missing from --exclude-libs, so the GUI re-exported the sam3 symbols and plugins that also link libsam3.a (FeedReader) failed with multiple definition of sam3GenerateKeys under the Ninja generator on Windows. Added libsam3.a to --exclude-libs. (The MSYS Makefiles generator was not affected; this keeps Ninja builds working too.)

4. build(friendserver): link OpenSSL again (needed via inlined libretroshare headers)
friendserver.cc pulls in libretroshare public headers whose inline functions call the OpenSSL EVP API (EVP_DigestUpdate, …). libretroshare links OpenSSL PRIVATE and is built shared on Windows, so those symbols are not propagated to consumers → undefined reference to EVP_DigestUpdate on a full Windows build. The friendserver now links OpenSSL itself.

Validation: full build (GUI + service + friendserver + plugins, RS_PLUGINS=ON) compiles on Linux, Windows (MSYS2/MinGW) and macOS.

defnax:
when get time i need check for compile

defnax:
code is clean i compiled with msys64 & qmake & qt 5.15(nothing broken), but not tested yet compile via cmake file

defnax:
seems need some info howto build on qt creator with the cmake file on windows, on qt creator not yet compileable
image

jolavillette:
here are the command lines that work for me

linux

rm -rf Build-cmake

cmake -B Build-cmake -S . \
  -DCMAKE_POLICY_VERSION_MINIMUM=3.5 -DCMAKE_BUILD_TYPE=Release \
  -DRS_RNPLIB=ON -DRS_JSON_API=ON -DRS_WEBUI=ON \
  -DRS_GUI=ON -DRS_SERVICE=ON -DRS_FRIENDSERVER=ON -DRS_PLUGINS=ON \
  -DRS_USE_I2P_SAM3=ON -DRS_BITDHT=ON -DRS_MINIUPNPC=ON \
  -DRS_BRODCAST_DISCOVERY=ON -DRS_SQLCIPHER=ON

cmake --build Build-cmake -j8


macos

HB="$(brew --prefix)"
LFLAGS="-L$HB/lib"; for d in "$HB"/opt/*/lib; do LFLAGS="$LFLAGS -L$d"; done
IFLAGS="-I$HB/include"; for d in "$HB"/opt/*/include; do IFLAGS="$IFLAGS -I$d"; done

rm -rf Build-cmake

cmake -B Build-cmake -S . \
  -DCMAKE_POLICY_VERSION_MINIMUM=3.5 -DCMAKE_BUILD_TYPE=Release \
  -DRS_RNPLIB=ON -DRS_JSON_API=ON -DRS_WEBUI=ON \
  -DRS_GUI=ON -DRS_SERVICE=ON -DRS_FRIENDSERVER=ON -DRS_PLUGINS=ON \
  -DRS_USE_I2P_SAM3=ON -DRS_BITDHT=ON -DRS_MINIUPNPC=ON \
  -DRS_BRODCAST_DISCOVERY=ON -DRS_SQLCIPHER=ON \
  -DCMAKE_C_FLAGS="$IFLAGS" -DCMAKE_CXX_FLAGS="$IFLAGS" \
  -DCMAKE_EXE_LINKER_FLAGS="$LFLAGS" \
  -DCMAKE_SHARED_LINKER_FLAGS="$LFLAGS" \
  -DCMAKE_MODULE_LINKER_FLAGS="$LFLAGS"

cmake --build Build-cmake -j8


windows

rm -rf Build-cmake

cmake -G Ninja -B Build-cmake -S . \
  -DCMAKE_POLICY_VERSION_MINIMUM=3.5 -DCMAKE_BUILD_TYPE=Release \
  -DRS_LIBRETROSHARE_STATIC=OFF -DRS_LIBRETROSHARE_SHARED=ON \
  -DRS_RNPLIB=ON -DRS_JSON_API=ON -DRS_WEBUI=ON \
  -DRS_GUI=ON -DRS_SERVICE=ON -DRS_FRIENDSERVER=ON -DRS_PLUGINS=ON \
  -DRS_USE_I2P_SAM3=ON -DRS_BITDHT=ON -DRS_MINIUPNPC=ON \
  -DRS_BRODCAST_DISCOVERY=ON -DRS_SQLCIPHER=ON

cmake --build Build-cmake -j8

defnax:
i tried only first qt creator it can use cmake file then stuck with openssl stuff

retroshare-gui/src/gui/PluginManagerWidget.cpp
```diff
@@ -160,9 +162,10 @@ PluginManagerWidget::registerNewPlugin(QString pluginName)
 void
 PluginManagerWidget::installPluginButtonClicked()
 {
-    QString fileName = misc::getOpenFileName(this, RshareSettings::LASTDIR_PLUGIN, tr("Open Plugin to install"), "./", tr("Plugins (*.so *.dll)"));
-
-    if (!fileName.isNull())
+    // misc::getOpenFileName returns bool and yields the path through an out-param
+    // (see util/misc.h); the previous call used an obsolete QString-returning form.
+    QString fileName;
+    if (misc::getOpenFileName(this, RshareSettings::LASTDIR_PLUGIN, tr("Open Plugin to install"), tr("Plugins (*.so *.dll)"), fileName))
```
csoler, line 168: would be nice to make two cases: on linux (and macOS?) the plugin are .so, on windows they are .dll.

csoler (changes requested):
I'm incapable of reviewing the cmake part. For the rest it's ok, up to the small #ifdef WINDOWS (or similar) that is needed for plugin extension.

G10h4ck (approved):
There are a few minor issue to fix, but leave those for after we merge Qt6 support, and merge this AS IS before is too late

PR #3231, opened 2026-06-06
fix(macos): add camera & microphone usage descriptions to Info.plist

jolavillette:
macOS 10.14+ terminates the app with SIGABRT when it accesses the camera or microphone without the matching usage-description keys in the bundle's Info.plist. Any VOIP audio/video call therefore crashes on macOS as soon as the capture device is opened (AVCaptureDeviceInput via QVideoInputDevice::start()).

Add NSCameraUsageDescription and NSMicrophoneUsageDescription so the system shows the standard permission prompt instead of aborting the process.


defnax:
cant test more voip not works on my old mac

PR #3232, opened 2026-06-10
split friendserver catch blocks into runtime_error + std::exception

HrachShah:
Both catch blocks in retroshare-friendserver/src/friendserver.cc (in handleClientPublish and handleIncomingClientData) used a single catch(std::exception& e) over a try-block whose body only ever throws std::runtime_error. Splitting the handler so a primary catch(const std::runtime_error& e) runs first (and keeps the existing e.what() log) and a fallback catch(const std::exception& e) runs for std-lib exceptions (bad_alloc, length_error, etc.) preserves the existing safety net while making the documented error path explicit and type-tagged.

The fallback handler also includes typeid(e).name() in the log so the operator can tell at a glance whether the underlying failure was the expected runtime_error or some other std-lib exception.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-## RetroShare
+README: updated# RetroShare
 
 RetroShare is a decentralized, private, secure, cross-platform, communication
 toolkit.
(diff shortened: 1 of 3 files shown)


defnax:
the readme file is broken look:
https://github.com/HrachShah/RetroShare/tree/fix/retroshare-cuboid-comment-or-ratray

PR #3237, opened 2026-06-15
Qt6 compatibility for the CMake build (version-agnostic Qt5/Qt6)

jolavillette:
> **Based on #3229 (CMakeMigration-Modular-CrossPlatform).** This PR is stacked on
> top of that branch and should be merged **after** #3229. Until then, its diff
> includes #3229's commits.

## What this does
Makes the **CMake** build compile against **Qt6** as well as **Qt5**, without
dropping Qt5. The C++ sources were already Qt6-guarded upstream; this only wires
Qt6 into CMake and fixes one CMake-specific issue.

## Changes
- **Version-agnostic Qt detection** in `retroshare-gui/CMakeLists.txt`:
  `find_package(QT NAMES Qt6 Qt5 ...)` + `Qt${QT_VERSION_MAJOR}::` targets. On Qt6
  it pulls `Core5Compat` (for `QRegExp` et al.); `QtX11Extras` is now requested on
  Qt5 only (`idle_platform.cpp` already uses `QNativeInterface::QX11Application`
  on Qt6). The dead `qt5_wrap_ui` double-wrap is removed in favour of a single
  versionless `qt_wrap_ui`.
- **Qt6 metatype fix** (`GxsGroupFeedItem.h`, `GxsFeedItem.h`, `PopupChatWindow.h`):
  `Q_DECLARE_METATYPE(RsGxsId/RsGxsGroupId/RsGxsMessageId)` is now guarded to Qt5
  only. Under Qt6, CMake AUTOMOC concatenates every `moc_*.cpp` into a single TU,
  where these complete types get their `QMetaTypeId` implicitly instantiated by one
  moc section and then explicitly specialized by another → *"specialization after
  instantiation"*. Qt6 auto-registers complete types, so the declaration is
  redundant there. (qmake compiles each moc separately and never hits this.)
- **CI**: three new Qt6 CMake workflows (Ubuntu, Windows/UCRT64, macOS), mirroring
  the existing Qt5 CMake ones.

## CI status
All six CMake jobs are green on this branch:

| CMake CI | Qt5 | Qt6 |
|----------|:---:|:---:|
| Ubuntu   | ✅  | ✅  |
| Windows (UCRT64) | ✅ | ✅ |
| macOS    | ✅  | ✅  |

No Qt5 regression.

## Out of scope
- **Plugins** are built `OFF` in the Qt6 CI: VOIP still uses the Qt5-era
  QtMultimedia API (`QAudioInput`/`QAudioOutput`/`QCameraInfo`, redesigned in Qt6)
  and FeedReader's CMake still hardcodes Qt5. Both can be ported in follow-ups.
- **qmake workflows are intentionally untouched.** Their pre-existing failures are
  inherited from the qmake→CMake migration in #3229, not introduced here.


G10h4ck:
There are a few minor issue to fix, but leave those for after we merge Qt6 support, and merge this AS IS before is too late

PR #3242, opened 2026-06-19
fix(macos): bundle ffmpeg transitive dylibs in deploy-macos.sh

jolavillette:
macdeployqt copies a plugin's direct deps (libVOIP -> libavcodec, libavutil, libspeex) but does not recurse into them, so libavcodec keeps an absolute reference to /opt/homebrew/.../libswresample.6.dylib. On the build Mac that path exists so the bundle works; on a clean Mac the lib is missing and the VOIP plugin fails to dlopen.

After macdeployqt, pull in the remaining third-party Homebrew deps (to a fixed point) and rewrite their references to @executable_path/../Frameworks. Scope is deliberately narrow: only the flat third-party dylibs in Frameworks/ and the RetroShare plugins in Resources/ -- Qt frameworks and Contents/PlugIns are left to macdeployqt. The install id is excluded so a lib is never treated as its own dependency.

### DIFF
--- a/build_scripts/OSX/deploy-macos.sh
+++ b/build_scripts/OSX/deploy-macos.sh
@@ -164,31 +164,51 @@ fi
 
 # 7. Déploiement Qt via macdeployqt
 echo ">>> Running macdeployqt..."
-if command -v macdeployqt &> /dev/null; then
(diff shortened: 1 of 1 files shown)


G10h4ck (approved):
Seems legit to me, but I don't have a Mac to test and inspect it deeply

PR #3244, opened 2026-06-22
Fix Windows/MinGW crash on thread exit: avoid creating a QObject on the caller thread (postToObject/postToThread)

jolavillette:
## What
`RsQThreadUtils::postToObject()` / `postToThread()` no longer create a temporary
`QObject` on the calling thread; they use
`QMetaObject::invokeMethod(obj, fun, Qt::QueuedConnection)` instead.

## Why
The previous implementations created a `QObject` on the **calling** thread and
relied on its `destroyed()` signal. When these helpers are called from a worker
`std::thread` that Qt did not create (e.g.
`GxsGroupFrameDialog::updateGroupSummary()`), creating that QObject makes Qt
**adopt** the thread: it allocates a `QThreadData` and registers a per-thread
cleanup. On Windows with Qt6, when such an adopted `std::thread` exits that
cleanup dereferences thread data that is not valid for a foreign thread and
crashes — a SIGSEGV in Qt's `Cleanup` teardown (`qtbase/.../qthread_win.cpp`).
RetroShare crashes a few seconds after startup on the Qt6 Windows build.

Creating a QObject on a thread Qt did not create is fragile by design; the right
fix is simply not to do it.

## How / compatibility
`QMetaObject::invokeMethod(obj, fun, Qt::QueuedConnection)` posts the functor to
`obj`'s own thread **without** creating any QObject on the caller, so a foreign
`std::thread` is never adopted by Qt. Behaviour is unchanged (and slightly safer:
Qt drops the call if `obj` is destroyed first). The functor overload of
`invokeMethod` exists since Qt 5.10 — the same `#if` branch this code already
used — so it builds against both Qt5 and Qt6. Fixes the Qt6 Windows startup
crash; no behavioural change on Linux/macOS.

### DIFF
--- a/retroshare-gui/src/util/qtthreadsutils.h
+++ b/retroshare-gui/src/util/qtthreadsutils.h
@@ -43,14 +43,15 @@ void postToObject(F &&fun, QObject *obj = qApp)
 {
 	if (qobject_cast<QThread*>(obj))
 		qWarning() << "posting a call to a thread object - consider using postToThread";
-	QObject src;
-	auto type = obj->metaObject();
-	QObject::connect( &src, &QObject::destroyed, obj,
-	                  [fun = std::move(fun), type, obj]
-	{
-		// ensure that the object is not being destructed
-		if (obj->metaObject()->inherits(type)) fun();
-	}, Qt::QueuedConnection );
+	// Post the functor to obj's thread WITHOUT creating a temporary QObject on the
+	// calling thread. Creating a QObject here adopts Qt on a non-Qt worker thread
+	// (e.g. std::thread): it allocates per-thread QThreadData and registers a
+	// thread_local cleanup, which crashes when that worker thread exits on MinGW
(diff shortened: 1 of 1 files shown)


csoler:
Seems correct to me, but I'm letting G10H4ck review this as well.

G10h4ck:
AFAIR the QObject creation was needed up to a certain version of Qt 5 (that now is most probably EOL), @jolavillette have you tested if with this change the RetroShare GUI works fine with latest Qt 5 version available?

jolavillette:
@G10h4ck yes. The change only touches the existing `Qt >= 5.10` #if branch (the pre-5.10 fallback is untouched), and `invokeMethod` with a functor is available since 5.10. I run this patch daily on Qt 5.15.13 (Linux) and tested the Qt5 builds on Linux, macOS and Windows (MSYS2) — GUI works fine, no regression.

PR #3247, opened 2026-06-23
📰 Fixed feeds issues ✔

defnax:


defnax:
@csoler

csoler:
I don't understand why you re-implemented a new label class. ZoomableLabel does exactly what you need already. Also in this PR the labels change size before loading and the images are not centered / fit to the label frame.

PR #3250, opened 2026-06-23
fix(deploy-windows): lock Qt plugin deployment to the linked major version

jolavillette:
fix(deploy-windows): lock Qt plugin deployment to the linked major version

windeployqt, the manual platforms/styles/imageformats fallbacks and the Qt translations were all picked from whichever Qt happened to be first on PATH or listed first in the script (qt6 before qt5). On a MSYS2 box where both Qt5 and Qt6 are installed, a Qt5 binary ended up bundled with the Qt6 platforms/ qwindows.dll, so Qt could not load any platform plugin and the app crashed at startup with 'no Qt platform plugin could be initialized'. The archive was also mislabelled (e.g. Qt-6.10 for a Qt5 build).

Detect the Qt major version actually linked by retroshare-gui.exe (objdump on its imports) and drive everything off it: select a version-matching windeployqt, copy plugins and translations only from the qt${QT_MAJOR} directories, and query the matching qmake for the archive name. Also derive the env tag (ucrt64/ mingw64/clang64) from $MSYSTEM instead of hard-coding mingw64.

### DIFF
--- a/build_scripts/Windows-msys2/deploy-windows.sh
+++ b/build_scripts/Windows-msys2/deploy-windows.sh
@@ -137,8 +137,14 @@ fi
 # Base name of the final archive. The environment tag reflects the MSYS2 shell
 # actually used (ucrt64 / mingw64 / clang64) instead of a hard-coded 'mingw64'
 # that lied on UCRT64/CLANG64 builds.
+#
+# Naming mirrors the Linux AppImage scheme:
+#   RetroShare-v<ver>-<YYYYMMDD>-<count>-g<hash>-Qt-<qtver>-<env>-msys2-x86_64
+# GIT_SUFFIX already carries "-<YYYYMMDD>-<count>-g<hash>" (or just the date when
+# exactly on a tag). The runtime slot is the MSYS2 environment (ucrt64-msys2),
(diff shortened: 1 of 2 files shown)


build_scripts/Windows-msys2/deploy-windows.sh
```diff
@@ -62,33 +62,81 @@ else
     GIT_SUFFIX="-${DATE_STR}"
 fi
 
-# Recherche de l'exécutable windeployqt adéquat
+# ------------------------------------------------------------------------------
```
csoler, line 67: in english plz ;-)

csoler (commented):
All comments should be in english.
Anyone who can test on windows + Qt6 is welcme to report.

G10h4ck:
seems good, but I would harmonize the EXE_NAME to `retroshare-gui.exe` if it doesn't require huge refactors, so we have name aligned as in linux. for the rest I really have no way to test this, so I can just entrust you to test it thoroughly

jolavillette:
@csoler comments translated to English in 7d288ffa8.
@G10h4ck the deployed exe is already retroshare-gui.exe everywhere (portable + installer); the .nsi fallback default now matches too (15320d644).

PR #3258, opened 2026-07-02
added warning for identities using deprecated signature format

csoler:


retroshare-gui/src/gui/Identity/IdentityListModel.cpp
```diff
@@ -465,6 +466,14 @@ QVariant RsIdentityListModel::data(const QModelIndex &index, int role) const
 	}
 }
 
+// This function checks that the ID has been signed using the old protocol that used to
+// include the service string into the signed data. New RS instances will not accept these identities anymore.
+
+static bool checkDate_deprecated_ids(const time_t t)
+{
+    return t > QDateTime(QDate(2026,4,20),QTime(0,0)).toSecsSinceEpoch();
+}
```
zapek, line 475: The fix was added in the source at that date but what happens if a user updated his identity with an older version after that? That date should be changed at the time of the release (which is in the future, so it needs to be revisited later).

There might be false positives for people who upgraded using an unofficial release but it's harmless.

csoler: Indeed. I don't think there is a universal solution to this problem.

PR #3260, opened 2026-07-03
build(cmake): enable the WikiPoos and The Wire GUI features

jolavillette:
Add the CMake plumbing to actually build the (experimental, off-by-default) GXS Wiki (WikiPoos) and The Wire features. Until now the retroshare-gui CMakeLists carried RS_GXSWIKIPOS / RS_GXSTHEWIRE blocks, but the options were never declared and the blocks contained bugs that only surfaced once enabled, so there was no working way to build these features with CMake.

- Declare RS_GXSWIKIPOS / RS_GXSTHEWIRE options at the super-project level (default OFF), so they propagate to libretroshare (which compiles the p3wiki/p3wire services and defines RS_USE_WIKI/RS_USE_WIRE) and to the GUI.
- retroshare-gui/src/CMakeLists.txt: the wiki/wire blocks called target_*() on the retroshare-gui target, which does not exist yet where src is include()d (add_executable runs later). Switch to directory-scoped add_definitions(). Also drop a stray qmake-ism ("RESOURCES +=") in the TheWire qrc list, and re-add two sources lost in the qmake->CMake port: WikiUserNotify (wiki) and CustomFrame (wire).
- Build the bundled peg-markdown used by WikiPoos: add supportlibs/pegmarkdown/CMakeLists.txt (static lib with the bundled dummy glib, no system glib-2.0 dependency) and link it into the GUI when RS_GXSWIKIPOS is on, mirroring the existing cmark integration.

The libretroshare backend change lives in its own repository/branch; the submodule gitlink is intentionally left untouched here so maintainers can advance it when merging. Both features remain OFF by default.

### DIFF
--- a/retroshare-gui/src/CMakeLists.txt
+++ b/retroshare-gui/src/CMakeLists.txt
@@ -945,19 +945,23 @@ if(RS_GXSPHOTOSHARE)
 endif(RS_GXSPHOTOSHARE)
 
 if(RS_GXSWIKIPOS)
-	target_compile_definitions(
-		${PROJECT_NAME} PUBLIC RS_USE_WIKI 
-		)
-
-	target_include_directories(${PROJECT_NAME} PUBLIC ../../supportlibs/pegmarkdown)
+	# NOTE: the retroshare-gui target is created by the parent CMakeLists.txt
+	# (add_executable) only AFTER this file is include()d, so target_* commands
+	# do not work here. add_definitions() is directory-scoped and applies to the
+	# target created later in this same scope. RS_USE_WIKI is also provided
+	# PUBLIC by libretroshare, but we set it explicitly for clarity. The
+	# pegmarkdown include path comes from linking the pegmarkdown target in the
+	# parent CMakeLists.txt.
+	add_definitions(-DRS_USE_WIKI)
 
 	list(
 		APPEND RS_IMPLEMENTATION_HEADERS
 		src/gui/WikiPoos/WikiDialog.h 
 		src/gui/WikiPoos/WikiAddDialog.h 
-		src/gui/WikiPoos/WikiEditDialog.h 
-		src/gui/gxs/WikiGroupDialog.h 
-		src/gui/gxs/RsGxsUpdateBroadcastBase.h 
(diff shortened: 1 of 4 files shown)


retroshare-gui/src/CMakeLists.txt
```diff
@@ -972,9 +976,10 @@ if(RS_GXSWIKIPOS)
 		APPEND RS_GUI_SOURCES
 		src/gui/WikiPoos/WikiDialog.cpp 
 		src/gui/WikiPoos/WikiAddDialog.cpp 
-		src/gui/WikiPoos/WikiEditDialog.cpp 
-		src/gui/gxs/WikiGroupDialog.cpp 
-		src/gui/gxs/RsGxsUpdateBroadcastBase.cpp 
+		src/gui/WikiPoos/WikiEditDialog.cpp
+		src/gui/WikiPoos/WikiUserNotify.cpp
+		src/gui/gxs/WikiGroupDialog.cpp
+		src/gui/gxs/RsGxsUpdateBroadcastBase.cpp
```
csoler, line 983: These 3 files shouldn't be here since they are not related to wiki.

csoler (commented):
Just one comment. I let the cmake druids review this.

retroshare-gui/src/CMakeLists.txt
jolavillette: These files aren't added by this PR — WikiEditDialog.cpp, WikiGroupDialog.cpp and RsGxsUpdateBroadcastBase.cpp were already in the RS_GXSWIKIPOS block on master (and in the qmake wikipoos block before that). They only show up in the diff because I trimmed trailing whitespace on the neighbouring lines when inserting WikiUserNotify.cpp. The only real additions here are WikiUserNotify (wiki) and CustomFrame (TheWire), both needed to fix link errors.

csoler: yes of course. I thought that it would be a good opportunity to move them to the right place. Also I suspect that they are already elsewhere otherwise it wouldn't compile with wiki disabled.

PR #3265, opened 2026-07-04
fix(gui): give RetroShare GUI its own version, independent from libretroshare

jolavillette:
fix(gui): give RetroShare GUI its own version, independent from libretroshare

following Gio's comments, new attempt at displaying RS GUI version without interfering with libretroshare

The GUI reported RS_HUMAN_READABLE_VERSION (a libretroshare macro) as "the RetroShare version". In the CMake build libretroshare exports RS_MAJOR_VERSION as a PUBLIC define, so the GUI inherited the engine's version and displayed it as its own: About showed the same number for both.

Give the GUI its own version, as discussed:

- Add retroshare-gui/src/rsguiversion.h defining RS_GUI_VERSION with a built-in default, so every build path (including qmake/Android) compiles even without version injection.
- Inject RS_GUI_VERSION from 'git describe' of the super-project in both CMake (retroshare-gui/CMakeLists.txt) and qmake (retroshare.pri).
- RsApplication::retroshareVersion() now returns the GUI's own version instead of libretroshare's.
- About (HelpDialog) shows the GUI version at the top; libretroshare's version is listed among the other libraries, like any dependency. AboutWidget already displayed both lines and is now correct automatically.

libretroshare needs no change: it already exposes its own version through RsInit::libRetroShareVersion() in both build systems.

### DIFF
--- a/retroshare-gui/CMakeLists.txt
+++ b/retroshare-gui/CMakeLists.txt
@@ -622,6 +622,27 @@ target_compile_definitions(${PROJECT_NAME} PRIVATE RS_RELEASE_VERSION )
 target_compile_definitions(${PROJECT_NAME} PRIVATE TARGET=\"retroshare\")
 target_compile_definitions(${PROJECT_NAME} PRIVATE RS_DIRECT_CHAT)
 
+# RetroShare GUI own version (RS_GUI_VERSION), independent from libretroshare.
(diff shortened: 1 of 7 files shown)


G10h4ck (approved):
Seems good to me

Issue #316, opened 2016-03-20
RetroShare Reference Manual in PDF Format -- Draft -- Created 3/20/2016 for Review


dtschmitz:
Dear RetroShare Developers,

Please find a rough draft RetroShare Reference Manual created today in PDF format (from LaTeX) for your review/feedback/use.

Let me know your thoughts.  -- Dietrich
[refman.pdf](https://github.com/RetroShare/RetroShare/files/181414/refman.pdf)

csoler:
It's a huge document! I suppose that users can generate it themselves when needed, using 
what's described here: http://stackoverflow.com/questions/4099572/how-to-get-a-single-pdf-document-from-doxygen. We should probably add the howto to generate the file in the toplevel README.

dtschmitz:
This type of document would be of greatest benefit to Developers, needless to say.  It's a 'draft' as mentioned and proffered as a way to provide assistance to those unfamiliar with RS apis and the object hiearchy -- coming to terms with a broader understanding of the project will only enhance their abilities to code.

As for 'users'.  I am not so interested in tendering it to them.  The wiki's format provides just enough of a veneer to satisfy their curiosity and isn't difficult to navigate to find setup information and FAQs.

I should think however, putting the doxygen HTML version on-line might be the preferred way to go.  Yes?

csoler:
Given that generating this document requires a simple command, I don't see a point in committing a large file that is going to change on every commit. However, listing that command in the README file seems a good idea to me.

dtschmitz:
csoler wrote:

> We should probably add the howto to generate the file in the toplevel README.

See new [issue](https://github.com/RetroShare/RetroShare/issues/321)
Closing 316.

PR #3270, opened 2026-07-06
build(appimage): linuxdeploy-based AppImage packaging (local + portable glibc 2.31)

jolavillette:
## What

Adds a maintained way to produce a RetroShare **GUI AppImage** from the CMake
build, using `linuxdeploy` + `linuxdeploy-plugin-qt`. Two entry points:

| Script | Compiles? | Result runs on |
|---|---|---|
| `build-appimage-local.sh` | no — packages the local `Build-cmake` binary | distros with glibc ≥ the build host |
| `build-appimage-docker.sh` | yes — inside a Debian 11 container | **any distro with glibc ≥ 2.31** (MX 21, Slackware 15, Debian 11+, Ubuntu 20.04+, Fedora 34+…) |

Both lay out a clean FHS `AppDir` by hand (the project's own `install()` rules
put the `.desktop` under `prefix/data/` with `RS_SERVICE_DESKTOP` off, which is
not AppImage-friendly) and bundle Qt + non-system libs.

## Files

- `build-appimage-local.sh` — package the local build (host glibc).
- `build-appimage-docker.sh` — host driver: build the toolchain image, run the container build.
- `Dockerfile.bullseye` — Debian 11 (glibc 2.31) toolchain + Qt5 deps + AppImage tools.
- `_appimage-inside.sh` — runs inside the container: cmake → AppDir → linuxdeploy.
- `BUILD.md` — full guide (glibc table, versioning gotcha, troubleshooting).
- `TESTERS.txt` — ships alongside the `.AppImage` for end users.

## Notes

- Uses **`linuxdeploy`**, not `probonopd/linuxdeployqt`: the latter's
  "continuous" build 107 fails silently right after the icon stage on this
  toolchain. `APPIMAGE_EXTRACT_AND_RUN=1` is set (Ubuntu 24.04 / containers lack
  libfuse2).
- **The glibc 2.31 (Docker) build requires the `RS_NO_LTO` opt-out**
  (libretroshare **PR #316**): Debian 11's GCC 10 ICEs (`lto1`) when linking
  libretroshare with LTO on. Documented in `BUILD.md`. The `local` build is
  unaffected.
- The pre-existing probonopd Qt4/trusty `Recipe` + `*.yml` + `makeAppImage*.sh`
  are left untouched (obsolete; flagged as such in `BUILD.md`), to keep this
  change purely additive.

### DIFF
--- a/build_scripts/AppImage/BUILD.md
+++ b/build_scripts/AppImage/BUILD.md
@@ -44,6 +44,12 @@ It downloads `linuxdeploy` + `linuxdeploy-plugin-qt` into
 `build_scripts/AppImage/tools/` (once), lays out a clean `AppDir`, and bundles
 Qt + libs.
 
+**Qt5 or Qt6:** the script reads the Qt major from the compiled binary and picks
+the matching `qmake` automatically (`qmake6` for a Qt6 build, `qmake`/`qmake-qt5`
+for Qt5), and aborts early on a qmake/binary Qt mismatch. So to ship a Qt6
+AppImage, just compile the GUI against Qt6 — no script change. Override the
+choice with `QMAKE=/path/to/qmake` if needed.
+
 ---
 
 ## 2. Portable build — `build-appimage-docker.sh`
--- a/build_scripts/AppImage/BUILD.md
+++ b/build_scripts/AppImage/BUILD.md
@@ -1,0 +1,227 @@
+# Building RetroShare AppImages (Linux)
+
+Two ways to produce a RetroShare GUI AppImage. Both **package the CMake build**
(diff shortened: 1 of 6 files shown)


G10h4ck (commented):
Huge but good overall, can you answer to my inline questions too?

jolavillette:
@G10h4ck no inline questions are visible on the PR — maybe still pending on your side?

jolavillette:
@G10h4ck friendly ping: your inline questions from July 23 never showed up here — they are most likely still in a pending review on your side, which nobody else can see. The PR has been conflict-free since July 6 with all 11 CI jobs green. Happy to answer as soon as they land.

PR #3274, opened 2026-07-12
Fix UI freeze on "mark all as read/unread" for forums, channels and boards

jolavillette:
needs libretroshare pr/331

Fix UI freeze on "mark all as read/unread" for forums, channels and boards

## Problem

Marking a large GXS group read — right-click → **Mark all as read** (or **unread**) — froze the UI for a very long time and could appear to hang indefinitely. On an 8000-post forum the hourglass stayed up for over an hour; force-quitting mid-way left thousands of posts still unread.

## Root cause

Two independent amplifiers, both triggered *per message*:

1. **GUI side** — the model spawned **one detached `std::thread` per changed post** (each blocking up to 5s waiting for its token to complete) and emitted **one `dataChanged()` per post**. Marking N posts created ~N threads and ~N signals, all driven from the GUI thread.
2. **Backend side** — `RsGenExchange::processMsgMetaChanges()` wrote each status change to the GXS database in **its own implicit transaction** (one fsync per message). On the SQLCipher-encrypted store this meant thousands of serial encrypted commits.

## Fix

**Backend (`libretroshare`)**

- `RsGeneralDataService`: new batch `updateMessageMetaData(std::vector<MsgLocMetaData>)` overload (default loops over the single version). `RsDataService` overrides it to persist the whole batch in **a single transaction**, held under one DB mutex so nothing interleaves.
- `RsGenExchange::processMsgMetaChanges()`: resolve all status masks first, then persist every queued change through that single transaction. This benefits **every** GXS service.
- New bulk read-status APIs that queue all changes at once and emit a **single** event:
  - `RsGxsForums::markRead(forumId, msgIds, read)`
  - `RsGxsChannels::setMessageReadStatus(channelId, msgIds, read)`
  - `RsPosted::setPostReadStatus(boardId, msgIds, read)`

**GUI (`retroshare-gui`)**

- `GxsForumModel`, `GxsChannelPostsModel` and `PostedPostsModel`: "mark all as read/unread" now collects the affected message ids, updates the in-memory model, issues **one** background batch call, and refreshes the view **once** — instead of one detached thread + one event + one `dataChanged()` per post.

## Result

The GUI thread now does `O(n)` in-memory work plus **one** thread and **one** signal, regardless of how many posts are affected, so it stays responsive on arbitrarily large groups. Persistence happens in a **single transaction** (one fsync). Works symmetrically for **read and unread**, on **forums, channels and boards**.

## Notes

- Backwards compatible: the existing single-message `markRead` / `setMessageReadStatus` / `setPostReadStatus` are unchanged; the batch variants are additive.
- This PR pairs with the companion in the other repository (`retroshare-gui` ⇄ `libretroshare`): the GUI batch calls require the new backend overloads. Companion PR: https://github.com/RetroShare/libretroshare/pull/331

## Testing

Open a forum/channel/board with thousands of posts, then right-click → **Mark all as read**, and again → **Mark all as unread**:

- the UI stays responsive throughout (no hourglass hang);
- unread counts update immediately;
- the state persists across a restart.

---

## Follow-up: single-post read/unread latency (4 additional commits)

Field testing on an 8259-post forum showed that marking a **single** post read or unread still froze the interface for about a second. Instrumented and fixed here, since the cause sits in the code paths this PR introduced:

- **Size the view refresh on changed rows, not message ids.** The refresh heuristic compared `changed_msgs.size() > 5`, but that vector holds one id per *stored version* of a post (an edited post keeps all its versions, sharing one read status). One click on a post with 47 stored versions crossed the threshold and emitted `dataChanged()` over the whole model: **1072 ms of frozen GUI, measured**. Counting changed rows instead brings the same click to about 1 ms.
- **Drop the expanded-items save/restore around read-status changes.** `setMsgReadStatus()` only emits `dataChanged()` and never resets the model, so nothing was ever lost; the restore paid a linear scan of the post array plus a linear scan of the view items per expanded item -- quadratic work on the GUI thread per toggle.
- **Re-read the group on `SUBSCRIBE_STATUS_CHANGED` instead of flipping the flag.** Guessing the new state by inverting the current flag goes wrong on duplicated or stale events; a wrong guess silently disabled read/unread marking until the next reload.
- **Opt-in latency probes** (`RS_GUI_PROFILE=<threshold-ms>`, inert when unset): per-operation timings plus a 50 ms GUI-thread watchdog that reports any event-loop stall and names the last probed operation. This is what located the 1072 ms freeze after several plausible-from-code-reading fixes had failed; happy to drop this commit if you would rather not carry it.


retroshare-gui/src/gui/gxsforums/GxsForumModel.cpp
```diff
@@ -871,19 +868,56 @@ void RsGxsForumModel::setMsgReadStatus(const QModelIndex& i,bool read_status,boo
...
+	// Larger operations (mark-all, with-children, versioned posts) use the
+	// batched call, which persists everything in a single transaction and emits
+	// a single event when the whole set is done.
+	if(changed_msgs.size() == 1)
+		RsThread::async( [grpId=mForumGroup.mMeta.mGroupId,msgId=changed_msgs[0],read_status]()
+		{
+			rsGxsForums->markRead(std::make_pair(grpId, msgId), read_status);
+		});
+	else if(!changed_msgs.empty())
+		RsThread::async( [grpId=mForumGroup.mMeta.mGroupId,changed_msgs,read_status]()
```
csoler: in order to avoid a deep copy of changed_msgs (which can be very long) it's possible to turn it into a pointer, then destroy this pointer in the async part just after calling markRead()

retroshare-gui/src/gui/Posted/PostedPostsModel.cpp
```diff
@@ -737,20 +737,38 @@ void RsPostedPostsModel::createPostsArray(std::vector<RsPostedPost>& posts)
...
 void RsPostedPostsModel::setAllMsgReadStatus(bool read)
 {
-    // make a temporary listof pairs
+    // Collect the posts whose status actually changes, persist them all in a
+    // single background batch (one thread and one event for the whole set,
+    // instead of one detached thread + one event per post, which froze the UI
+    // on large boards), then update the local model and refresh once.
 
-    std::list<RsGxsGrpMsgIdPair> pairs;
+    std::vector<RsGxsMessageId> msgIds;
```
csoler, line 770: same, use a pointer

retroshare-gui/src/gui/Posted/PostedPostsModel.cpp
```diff
@@ -737,20 +737,38 @@ void RsPostedPostsModel::createPostsArray(std::vector<RsPostedPost>& posts)
...
-    for(auto& p:pairs)
-        RsThread::async([read,p]()
+    if(!msgIds.empty())
+        RsThread::async([boardId=mPostedGroup.mMeta.mGroupId, msgIds, read]()
         {
-            rsPosted->setPostReadStatus(p,read);
+            rsPosted->setPostReadStatus(boardId, msgIds, read);
         } );
+
+    // Update the local model immediately, since we don't catch the resulting
```
csoler: not nice. This is a hack that breaks the flow of information. In particular if you do not catch the post update event you will miss other changes that do not come from here. Suppose for instance that the webUI is changing the posts read status...

retroshare-gui/src/gui/gxsforums/GxsForumThreadWidget.cpp
```diff
@@ -253,6 +254,13 @@ GxsForumThreadWidget::GxsForumThreadWidget(const RsGxsGroupId &forumId, QWidget
 {
     ui->setupUi(this);
 
+    // Single-shot timer used to coalesce the full-forum reloads requested by
+    // incoming GXS events (see scheduleForumReload()). Created first thing:
+    // setGroupId(forumId) below reaches updateDisplay(), which touches this timer.
+    mDeferredReloadTimer = new QTimer(this);
```
csoler, line 260: don't forget to delete this one in the destructor of GxsForumThreadWidget

retroshare-gui/src/gui/gxsforums/GxsForumModel.cpp
```diff
@@ -871,19 +869,86 @@ void RsGxsForumModel::setMsgReadStatus(const QModelIndex& i,bool read_status,boo
 	if(!convertRefPointerToTabEntry(ref,entry) || entry >= mPosts.size())
 		return ;
```
csoler, line 870: That should have a #ifdef DEBUG_SOMETHING around.

csoler (changes requested):

retroshare-gui/src/gui/gxsforums/GxsForumModel.cpp
jolavillette: done.

retroshare-gui/src/gui/Posted/PostedPostsModel.cpp
jolavillette: done.

retroshare-gui/src/gui/Posted/PostedPostsModel.cpp
jolavillette: You were right — a webUI-initiated batch would have been missed. The batched READ_STATUS_CHANGED event now carries the affected msg ids and their new state (lib side), and the model updates from its event handler like for any other change; the direct local update is gone.

retroshare-gui/src/gui/gxsforums/GxsForumThreadWidget.cpp
jolavillette: The timer is parented to the widget (`new QTimer(this)`), so Qt deletes it with the widget — no leak.

retroshare-gui/src/gui/gxsforums/GxsForumModel.cpp
jolavillette: Probes removed from the PR altogether (probe commit reverted), consistently with the other PRs.

PR #3275, opened 2026-07-17
Feature/network global search

jolavillette:
Feature/network global search

in Network tab, in Friend list and in Keyring

The search is now generic and works on all columns: Name, PGP ID, Node ID, connection, IP, etc.


defnax:
for me looks like you add to much text , we need keep it simple only one word is enough not 2, 3 and more text on a column or placeholder text

jolavillette:
done — placeholders back to "Search", column headers back to "Id".

PR #3280, opened 2026-07-27
GUI: stop copying and sorting the channel post array in the Qt thread

jolavillette:
GUI: stop copying and sorting the channel post array in the Qt thread

works with libretroshare pr/351
https://github.com/RetroShare/libretroshare/pull/351

RsGxsChannelPostsModel::setPosts() runs in the Qt thread through postToObject(). It copied the whole post array — thumbnails included — and then sorted it there, so opening a channel with a few thousand posts froze the interface for the duration of both.

The sort moves to the loader thread in update_posts(), where the array already sits after the service call, and setPosts() takes its argument by rvalue reference so the array is moved in rather than copied. The sort is kept in setPosts() as a safety net: on an already ordered array it only runs comparisons and moves nothing.

Measured on a channel with 1931 posts, setPosts went from a visible freeze to 0-3 ms of copy and 0 ms of sort.

The second commit adds opt-in profiling of the two GUI phases (RS_GXS_PROFILE), matching the libretroshare side.

Requires libretroshare pr/351: this branch includes the profiling header added there, and eliminating the deep copies depends on the move operations it introduces.

### DIFF
--- a/retroshare-gui/src/gui/gxschannels/GxsChannelPostsModel.cpp
+++ b/retroshare-gui/src/gui/gxschannels/GxsChannelPostsModel.cpp
@@ -527,7 +527,7 @@ void RsGxsChannelPostsModel::updateSinglePost(const RsGxsChannelPost& post,std::
     triggerViewUpdate(true,false);
 }
 
-void RsGxsChannelPostsModel::setPosts(const RsGxsChannelGroup& group, std::vector<RsGxsChannelPost>& posts)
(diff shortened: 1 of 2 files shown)


csoler:
You may remove the profiling code that helped you find the CPU costs.

csoler (changes requested):
Could you remove all the profiling from this PR? It's a very useful tool, but it shouldn't be kept in the optimized version of the code. The optimization here is clearly a combination of the std::move (which avoids a lots of deep copy because of images) and the offline std::sort. The former is probably more significant.

jolavillette:
done

PR #3281, opened 2026-07-27
Fix tor

thunder2:
Requires libretroshare #352

### DIFF
--- a/retroshare-gui/src/rshare.h
+++ b/retroshare-gui/src/rshare.h
@@ -132,13 +132,6 @@ public:
   /** Returns Rshare's application startup time. */
   static QDateTime startupTime();
 
-  /** Returns the location Rshare uses for its data files. */
-  static QString dataDirectory();
-  /** Returns the default location of Rshare's data directory. */
-  static QString defaultDataDirectory();
-  /** Creates Rshare's data directory, if it doesn't already exist. */
-  static bool createDataDirectory(QString *errmsg);
-  
   /** Writes <b>msg</b> with severity <b>level</b> to RetroShare's log. */
   static Log::LogMessage log(Log::LogLevel level, QString msg);
   
--- a/retroshare-service/src/retroshare-service.cc
+++ b/retroshare-service/src/retroshare-service.cc
@@ -283,7 +283,56 @@ int main(int argc, char* argv[])
 #endif /* defined(RS_JSONAPI) && defined(RS_WEBUI)
 	&& defined(RS_SERVICE_TERMINAL_WEBUI_PASSWORD) */
 
-	conf.main_executable_path = argv[0];
+	std::string mainExecutablePath;
+#ifdef WINDOWS_SYS
+	{
+		// Use a dynamic buffer to handle paths longer than MAX_PATH safely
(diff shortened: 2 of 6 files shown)


retroshare-gui/src/rshare.h
```diff
@@ -132,13 +132,6 @@ class RsApplication : public QApplication
   /** Returns Rshare's application startup time. */
   static QDateTime startupTime();
 
-  /** Returns the location Rshare uses for its data files. */
-  static QString dataDirectory();
```
csoler, line 136: This one is used in the constructor of PluginManager

thunder2: I’ve seen it. It’s only compiled if it’s enabled in the configuration via "pluginmgr", and is never used.
```
pluginmgr {

        SOURCES += gui/PluginsPage.cpp \
            	gui/PluginManagerWidget.cpp \
            	gui/PluginManager.cpp

        HEADERS += gui/PluginsPage.h  \
            	gui/PluginManagerWidget.h \
            	gui/PluginManager.h

        DEFINES *= PLUGINMGR

}

```
If we need this, we should add a function similar to RsAccounts::AccountDirectory to libretroshare, so that both parts of the program use the same directory.

csoler: These pages are the ones that hang the plugin manager as a tab. Since it has moved to settings, I think we can remove it.

retroshare-service/src/retroshare-service.cc
```diff
@@ -283,7 +283,56 @@ int main(int argc, char* argv[])
 #endif /* defined(RS_JSONAPI) && defined(RS_WEBUI)
 	&& defined(RS_SERVICE_TERMINAL_WEBUI_PASSWORD) */
 
-	conf.main_executable_path = argv[0];
+	std::string mainExecutablePath;
+#ifdef WINDOWS_SYS
+	{
```
csoler, line 288: This is crazy. Isn't there a function in RS that already does this stuff for file paths windows?

csoler (commented):
See my comment about windows file paths. It might be work adding the necessary function in a specific windows cpp file.

defnax:
hope soon is fixed its importend to get work rs again with tor before a release done

retroshare-service/src/retroshare-service.cc
csoler: Can you plz move this terrible code in retrosharewin32.cpp, so that it's more or less hidden before we merge this PR?

thunder2: Sorry, I haven't received any notification emails from GitHub.
The Retroshare service has no retrosharewin32.cpp so I have to create a new one in the new folder util?

csoler: ah sorry! I forgot that. I thought retrosharewin32.cpp was in libretroshare. Well it's up to you. I just thought that having this horrible windows-only code in the middle wasn't so great. We can move it later on if that PR is urgent.

thunder2: I will move the code in a next PR so main is clean.

PR #3285, opened 2026-08-03
⚙️ Added tray option & create account for retroshare-service

defnax:
image

Account creation is only for default standard account for hidden needs changes in libretroshare to get work.


defnax:
@G10h4ck maybe check if its ok about what i implemented? create service works very well from terminal when there is no account

* Optional Tray function via `RS_SERVICE_TRAY`
* Create account feature `RS_SERVICE_TERMINAL_LOGIN`

PR #3286, opened 2026-08-05
GUI: relaunch focuses existing instance, add -n/--new-instance

### LINKED ISSUE #1225: Have retroshare command show existing instance if running
I use Xmonad so I don't have a notification icon for Retroshare. I want to be able to reopen the window by running the `retroshare` command via dmenu. I understand that some people want to run multiple instances. I suggest a command line switch to start a new instance for those cases.

Elitenotavailable:
When a RetroShare instance is already running, running the retroshare command again now hands the arguments over to that instance - or, when no arguments are given, simply asks it to show its main window - and then exits, instead of silently doing nothing.

A new -n/--new-instance command line option forces the start of a separate instance. Passing a different base directory with -c <dir> also still starts a separate instance for that profile.

Fixes #1225

### DIFF
--- a/retroshare-gui/src/gui/MainWindow.cpp
+++ b/retroshare-gui/src/gui/MainWindow.cpp
@@ -1296,6 +1296,14 @@ void MainWindow::receiveNewArgs(QStringList args)
 {
     RsInfo() << "Received new arguments from operating system call.";
 
+    if (args.isEmpty())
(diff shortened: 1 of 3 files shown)


Elitenotavailable:
Hello , I am new to open source , how long does it usually takes to review a PR ?

Issue #361, opened 2016-04-30
(SSL_pending()) core dump 201604300815


dtschmitz:
[rs-coredump-201604300815.txt](https://github.com/RetroShare/RetroShare/files/243587/rs-coredump-201604300815.txt)
(Coincidentally, I had just added 'Auto-Subscribe' to @DTS RS instance.)

<strike>tailing the retro.log, i see 'tons' of this scrolling by:
72f35def5aa878d SSL_read() SSL_ERROR_ZERO_RETURN : nReadZero: 12930 
(Sat Apr 30 08:47:54 2016 Z: 37714, lvl:1): pqissl::readdata() -> calling reset() </strike>
Disregard ^^^ it was transient.  Now gone.

<strike>I will roll back to previous commit until i hear back on this.  Dietrich</strike>

I have rolled forward to the SSL_pending() commit and will stand watch tailing retro.log and journal.ctl (vm 'yo' has been stable). (shrugs)

dtschmitz:
closing issue.  it's been stable since restarting.  So, I won't "borrow trouble".

PR #3288, opened 2026-08-06
reorganize CMakeLists.txt files

dbear496:
- centralizes handling of dependency details to supportlibs/CMakeLists.txt
- properly integrates FetchContent with find_package for dependency management
- Uses ExternalProject_Add to independently build libsam3 using its officially supported build system (GNU Make), and implements a stand-in libsam3-config.cmake file to add a CMake target for the library.
- moves libbitdht to supportlibs directory
- removes use of CMAKE_POLICY_VERSION_MINIMUM (which is not intended for use by developers)
- fixes the Exec value in the .desktop file
- .gitignore's unrecognized plugins in /plugins/

I haven't done much work in the retroshare-gui directory even though the CMakeLists.txt there could also use some tidying up.

This PR should be paired with RetroShare/libretroshare#361 .


jolavillette:
I built this pair locally on Linux, Qt5 and Qt6. Two blockers live in this repo; the rest of my findings, the testing caveat and the libretroshare-side issues are in RetroShare/libretroshare#361 so the discussion stays in one place.

**1. `supportlibs/libbitdht` is pinned to a commit that doesn't exist publicly.** It points at `51aebb6e`, which the GitHub API reports as 422 in `RetroShare/BitDHT`, and there's no `dbear496/BitDHT` fork either. That makes this PR impossible to check out for anyone but you, and it's what kills all 11 CI checks at `Init submodules` after ~8s — none of those runs contains any compiler output. master is on `4355ea96`.

You probably can't see it locally, because of the point below.

**2. `supportlibs/CMakeLists.txt:306` — the sed expression aborts configure.** FetchContent copies `PATCH_COMMAND` arguments verbatim into the generated sub-build `CMakeLists.txt`, and `\1` isn't a valid CMake escape, so re-parsing it fails with `Invalid character escape '\1'`. Dropping the capture group fixes it:

```cmake
PATCH_COMMAND sed -i -e "s/^cmake_minimum_required(VERSION 2.8)\$/cmake_minimum_required(VERSION 2.8...4.3.4)/" CMakeLists.txt
```

**On the `if(EXISTS <local>) … else(GIT_REPOSITORY …)` pattern.** It silently clones upstream when a submodule is missing, with no warning. That's almost certainly why the dead libbitdht pin is invisible on your machine — CMake just clones BitDHT master into `_deps` instead. It also means a build can succeed against an unpinned dependency while you believe you're testing the pinned one. I only caught it by checking what the dependency had actually resolved to. Worth a `message(WARNING)` on that path.

**Smaller things in this repo:**

- `BUILD_ENVIRONMENT_MODIFICATION` in the libsam3 `ExternalProject_Add` is silently accepted and ignored by CMake 3.28 (tested), so `CC=${CMAKE_C_COMPILER}` never applies and libsam3 builds with the default `cc`. It needs a newer CMake than the declared 3.24 minimum, which matters for cross-compilation.
- Raising the minimum from 3.5 to 3.24 drops Ubuntu 20.04 (3.16) and Debian 11 (3.18). Intentional? It cuts against the packaging benefit of the `find_package` rework, since those are the distros most likely to want system dependencies.
- The per-target Windows and GCC-15 fixups master had are gone: `UNICODE`/`_UNICODE` on `bitdht`, and `-Wno-error=incompatible-pointer-types,int-conversion,implicit-function-declaration` on `sam3`/`bitdht`/`miniupnpc`. Only udp-discovery keeps them. Note `bitdht` is created from `supportlibs/`, so libretroshare's `add_definitions(-DUNICODE -D_UNICODE)` never reaches it — different directory scope.
- Since `SOURCE_DIR` points at the submodules when present, the `sed -i` patches rewrite the developer's own working tree in place rather than a downloaded copy.
- The description links libretroshare#136; I think you mean #361.

I have the sed fix as a one-line commit and can open a PR against `cmake-refactor`.

dbear496:
Oh, my bad--I forgot to push a PR for bitdht.

The sed PATCH_COMMAND works for me, so I think it is a bit strange you see different behavior. I'm suspecting sed might be using different defaults between us, so I need to figure out how to make the behavior consistent. Are you using GNU sed?

I agree, a `message(WARNING ...)` when the submodule is not found would be a good idea.

According to the docs, `BUILD_ENVIRONMENT_MODIFICATION` was added in CMake 4.2. I don't think this is an issue for the usual use case, but as you mentioned, it is important for cross-compilation. Is it important to support cross compilation with older CMake? If so, I can instead modify the BUILD_COMMAND when CMake is too old for BUILD_ENVIRONMENT_MODIFICATION.

> Raising the minimum from 3.5 to 3.24 drops Ubuntu 20.04 (3.16) and Debian 11 (3.18). Intentional? It cuts against the packaging benefit of the find_package rework, since those are the distros most likely to want system dependencies.

Yes. CMake 3.24 adds integration between FetchContent and find_package, restbed requires CMake 3.22, and support for CMake <3.10 is deprecated. Sorry, Ubuntu 20.04...maybe it's time to update.

The per-target Windows/GCC 15 fixups for BitDHT are moved to the BitDHT project (which I forgot to push). The fixups are not needed for miniupnpc because that target is not built within the RetroShare super-project. (On master, miniupnpc is an external dependency and the target does not exist.) The fixups are irrelevant for libsam3 because that dependency is added with ExternalProject and hence uses its own build system (GNU make).

> Since SOURCE_DIR points at the submodules when present, the sed -i patches rewrite the developer's own working tree in place rather than a downloaded copy.

I am aware :/ Ideally, I'd like to avoid using PATCH_COMMAND altogether because it is very hacky. (For sam3, I actually decided to use ExternalProject instead of FetchContent just to avoid this issue. I toyed with using PATCH_COMMAND to drop in a whole CMakeLists.txt, and it worked, but I didn't like that it changed the the working tree.) For udp-discovery, ExternalProject won't fix the issue because the project is broken upstream ¯\\_(ツ)_/¯ The way to fix this is submit a patch upstream.

jolavillette:
Thanks for pushing the BitDHT fork — `51aebb6e` resolves from upstream now.

The CI will still be red though, and it's not the pin: all eleven jobs fail with
`error: pathspec 'libbitdht/' did not match any file(s) known to git`. Moving the submodule to
`supportlibs/libbitdht` left every build workflow — eleven of them — asking for the old path. The
deploy scripts have it too (`deploy-macos.sh:137`, `deploy-windows.sh:171`, `makeOSXPackage.sh:45`
copy `libbitdht/src/bitdht/bdboot.txt`).

Yes, GNU sed 4.9 — but it isn't sed, sed never runs. CMake refuses to parse the sub-build file it
generates from the command. I bisected your exact line, same sed throughout, only the CMake binary
changing: 3.28 and 3.29 fail with `Invalid character escape '\1'`, 3.30 onward work. So it breaks on
3.24-3.29, which is inside the minimum you declare.

On `BUILD_ENVIRONMENT_MODIFICATION`: yes, the older-CMake fallback matters. 4.2 is available nowhere
we build — Ubuntu 24.04 ships 3.28, and `prepare-toolchain-clang.sh` uses the host cmake — so right now
the property is silently ignored on every Android and mingw build. libsam3's Makefile doesn't set `CC`
either, so it falls back to make's default `cc`.

You're right on miniupnpc and sam3, withdrawn. Agreed on 3.24 and on patching udp-discovery upstream.

One separate thing while you're in there: `supportlibs/CMakeLists.txt:177` looks for
`supportlibs/openpgpsdk/`, but `.gitmodules` still has it at the repo root — libbitdht was moved,
openpgpsdk wasn't. So the `EXISTS` test always fails and it silently clones OpenPGP-SDK master instead
of using the pinned submodule. Exactly the case that `message(WARNING)` would have caught.

dbear496:
> it isn't sed, sed never runs. CMake refuses to parse the sub-build file it
generates from the command.

This doesn't make sense because if sed never runs, then the sub-build CMakeLists.txt file will not be modified.

If a literal `\1` appears in supportlibs/udp-discovery-cpp/CMakeLists.txt, then that must mean CMake is preserving the double backslash from line 306 and invoking the sed command with literally `\\1`. But this should be impossible because the [CMake documentation](https://cmake.org/cmake/help/v3.24/manual/cmake-language.7.html#quoted-argument) states, "A `\` followed by a non-alphanumeric character simply encodes the literal character". Alternatively, maybe sed is invoked with `\1` but your version of sed does not interpret it as a capture group reference, but this couldn't be the case if the behavior changes between CMake version 3.29 and 3.30.

Does it work for you if you put the `--posix` option in the sed command?

jolavillette:
No — I tried it. On 3.28 `sed --posix -i -e …` gives exactly the same `Invalid character escape '\1'`,
and on 4.2 it configures fine and patches the file correctly with or without the flag. GNU sed with
`--posix` also handles `\1` perfectly well on its own (back-references are standard POSIX BRE; `--posix`
only disables GNU extensions). If sed were the variable here, that flag would have moved at least one of
those results. It moves neither.

The reason it can't is that sed is never reached in the failing case. You're right that nothing gets
modified — configure aborts first. The `\1` isn't in `supportlibs/udp-discovery-cpp/CMakeLists.txt`,
it's in the file FetchContent *generates*, which is the one the error names:
`_deps/udp-discovery-cpp-subbuild/CMakeLists.txt:30`. Straight out of my build tree on 3.28:

```cmake
ExternalProject_Add(udp-discovery-cpp-populate
   "UPDATE_DISCONNECTED" "False" "GIT_REPOSITORY" "…" "EXTERNALPROJECT_INTERNAL_ARGUMENT_SEPARATOR"
   "GIT_TAG" "origin/master" "GIT_SHALLOW" "TRUE" "GIT_PROGRESS" "TRUE" "PATCH_COMMAND" "sed" "-i"
   "-e" "s/^\(cmake_minimum_required(VERSION 2.8\))$/\1...4.3.4)/" "CMakeLists.txt"
```

Single backslashes, re-emitted as a quoted argument — so it goes through quoted-argument escape rules
when CMake configures that sub-build. And the grammar is on the very page you linked, immediately above
the sentence you quoted:

```
escape_sequence  ::= escape_identity | escape_encoded | escape_semicolon
escape_identity  ::= '\' <match '[^A-Za-z0-9;]'>
escape_encoded   ::= '\t' | '\r' | '\n'
escape_semicolon ::= '\;'
```

`\(` matches `escape_identity`, so it's fine — that's your sentence. `\1` matches none of the three, so
it's an invalid escape, which is verbatim what the error says. Nothing preserved a double backslash, and
no regex dialect is involved.

I was imprecise about the 3.29/3.30 boundary, though — it isn't the version, it's **CMP0168** (added in
3.30), which makes FetchContent implement the steps directly instead of through a sub-build. Under NEW
the command goes to `CMakeFiles/fc-tmp/…/patch.cmake` as bracket arguments, which interpret no escapes:

```cmake
COMMAND [====[sed]====] [====[-i]====] [====[-e]====]
        [====[s/^\(cmake_minimum_required(VERSION 2.8\))$/\1...4.3.4)/]====] [====[CMakeLists.txt]====]
```

What sets that policy NEW is the `...4.4.1` in your `cmake_minimum_required`. On the same project with
CMake **4.2.0**, plain `cmake_minimum_required(VERSION 3.24)` fails with `Invalid character escape '\1'`
and the call stack goes through `__FetchContent_populateSubbuild`; adding `cmake_policy(SET CMP0168 NEW)`
to that same file makes it configure cleanly. So it's not an old-CMake problem: on 3.24-3.29 there's no
direct mode to fall back on, and on 3.30+ it holds only as long as CMP0168 stays NEW.

Dropping the capture group is immune to all of that, which is why I'd still go that way.

dbear496:
Aha, that makes sense. It appears then that the only way to be compatible with both the NEW and OLD behavior is to not use backslashes. I'm surprised this isn't mentioned in the docs for CMP0168.

jolavillette:
The new `PATCH_COMMAND` works — I ran your line on CMake 3.28: configures cleanly and the file comes out
as `cmake_minimum_required(VERSION 2.8...4.3.4)`. The openpgpsdk move and the restbed pin-by-commit look
right too.

The `BUILD_COMMAND` fallback works as well — `EnvironmentModification` expands to
`cmake -E env --modify CC=set:/usr/bin/gcc --`, and I confirmed that passes `CC` through on 3.28. One
gap: `cmake -E env --modify` was added in **3.25**, so on 3.24 — the minimum this PR declares — the
fallback itself fails. Same shape as the sed issue. The plain `NAME=VALUE` and `--unset=NAME` forms carry
no `versionadded` in the docs and go back to the introduction of `cmake -E env`, if you want to cover
that last version.

On the CI: moving openpgpsdk added a second stale path. Eleven workflows still ask for `libbitdht/`, and
nine of those also ask for `openpgpsdk/` — e.g. `ubuntu-cmake.yml:42` reads
`libbitdht/ libretroshare/ openpgpsdk/ retroshare-webui/`. The jobs stop at the first one:

```
error: pathspec 'libbitdht/' did not match any file(s) known to git
```

The deploy scripts have the same problem and no CI covers them —
`build_scripts/OSX/deploy-macos.sh:137` and `Windows-msys2/deploy-windows.sh:171` copy
`libbitdht/src/bitdht/bdboot.txt` behind an `if [ -f ]`, so they'd quietly ship packages without the DHT
bootstrap file; `build_scripts/OSX/makeOSXPackage.sh:45` does the same copy unguarded.

Two smaller ones while you're in the submodules: `supportlibs/librnp` moves from Dec 2024 to Apr 2026 and
`supportlibs/rapidjson` from the 1.1.0 tag (2016) to a Dec 2024 master commit. Both are real dependency
changes worth a line in the description, or reverting if they came in with a `--remote` update. And
there's a single `message(WARNING)` in `supportlibs/CMakeLists.txt` — worth one per dependency, since
each `EXISTS` test can fall through the same way openpgpsdk did.

jolavillette:
The `EnvironmentModification` fallback doesn't do what it's meant to — two problems.

**The version threshold is a typo:** `elseif(CMAKE_VERSION VERSION_GREATER_EQUAL 2.25)` should be
`3.25`. Since every 3.x is `>= 2.25`, CMake 3.24 takes the `--modify` branch, and on a real 3.24.4
binary that gives:

```
$ cmake -E env --modify CC=set:/usr/bin/gcc -- sh -c 'echo $CC'
cmake -E env: unknown option '--modify'
```

So the one version the fallback exists for is the one it breaks on, and the compatible `else()` branch
is unreachable in practice.

**The regexes in that unreachable branch wouldn't work either:** `^([^=])=set:(.*)$` matches only a
**one-character** variable name. Running the transform on a realistic list:

```
legacy transform result: [CC=set:/usr/bin/gcc;LDFLAGS=set:-lws2_32;FOO=unset:]
with [^=]+ instead:      [CC=/usr/bin/gcc]
```

and the untransformed form does exactly what you'd expect — `cmake -E env CC=set:/usr/bin/gcc` sets
`CC` to the literal string `set:/usr/bin/gcc`. `([^=]+)` fixes it. The plain `NAME=VALUE` form does work
on 3.24, so the legacy branch is the right idea once it's reachable and the regex matches.

On the CI: moving openpgpsdk added the second stale path I expected — the jobs now report both

```
error: pathspec 'libbitdht/' did not match any file(s) known to git
error: pathspec 'openpgpsdk/' did not match any file(s) known to git
```

Eleven workflows reference `libbitdht/` and nine of them also reference `openpgpsdk/`.

dbear496:
I noticed pegmarkdown seems to be insufficiently incorporated into the build. I can see it is added as an include directory, but it doesn't look like the source files are ever compiled. I'm also having trouble finding the source repository. Is this dependency even worth keeping around?

PR #3297, opened 2026-08-30
Distant chat window: close it when the conversation is closed from another client

jolavillette:
Depends on libretroshare `fix/distant-chat-locally-closed-event` (new `RsDistantChatEventCode::TUNNEL_STATUS_LOCALLY_CLOSED`), itself stacked on libretroshare #378. Does not build against a libretroshare without that enum value.

A distant chat started from the web UI pops the desktop chat window for its tunnel (the core sends itself a "[Starting distant chat" message). Leaving that conversation from the web UI then left the window as it was: green LED, "You can talk", and on close the offer to end a conversation the core no longer had — `closeEvent()` ignored the false answer of `getDistantChatStatus()` and read a default-initialised status.

On the new `TUNNEL_STATUS_LOCALLY_CLOSED` event the window closes, as it does when the conversation is ended from the window itself. And `closeEvent()` asks no confirmation and calls no close for a tunnel the core does not know any more.

### DIFF
--- a/retroshare-gui/src/gui/chat/PopupDistantChatDialog.cpp
+++ b/retroshare-gui/src/gui/chat/PopupDistantChatDialog.cpp
@@ -217,16 +217,30 @@ void PopupDistantChatDialog::handleEvent_main_thread(std::shared_ptr<const RsEve
 		getChatWidget()->unblockSending();
         setPeerStatus(RsStatusValue::RS_STATUS_ONLINE);
 		break;
+
+    case RsDistantChatEventCode::TUNNEL_STATUS_LOCALLY_CLOSED:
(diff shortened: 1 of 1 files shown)


csoler:
When the other side closes the conversation, you may want to keep your window open, for instance to copy/save some info that was sent, click on some links, etc. Also the other side may close by mistake (wrong click, etc) and then will open again. The "peer has closed the conversation" was meant for that already. Am I missing something?

jolavillette:
The problem is not that the chat window stays opened on the friend side. As you said that is the normal behaviour.
The problem is that when you close the chat windows in the webui it stays open in the UI, and vice versa.
3297 and 383 fix that.

PR #3303, opened 2026-09-01
🐞 Windows multi-client autostart fix

defnax:
Changes:
- Each RetroShare executable path now gets its own registry value: `RetroShare-<path-hash>`.
- Multiple installations can independently enable or disable autostart.
- Existing legacy RetroShare entries remain recognized.
- Legacy entries are migrated/removed only when they belong to the current executable.
- Another installation’s autostart entry is never modified.
- Minimized startup (-m) remains supported.

### DIFF
--- a/retroshare-gui/src/gui/settings/rsharesettings.cpp
+++ b/retroshare-gui/src/gui/settings/rsharesettings.cpp
@@ -20,6 +20,7 @@
  *******************************************************************************/
 
 #include <math.h>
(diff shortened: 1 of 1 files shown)


defnax:
@thunder2 @csoler 
This is important, very old issue when you have more rs clients on windows with autostart
Or when running test client on last time the next windows start its starts the wrong retroshare

defnax:
@csoler bugfix

PR #3304, opened 2026-09-01
🐞Windows-only channel-row height correction that removes the extra spa…

defnax:
…ce below thumbnails.

**before this fix on Windows:(Master)**
image

with the fix:
image

smooth image scaling:
image


defnax:
@csoler bug fix

csoler:
There's something inconsistent here: the size of the cell should be in accordance to what is drawn in paint() and the same code should be used on all OS, unless Qt has a bug on windows. The size problem that you're correcting is probably coming from a discrepancy between the code in sizeHint() and the code in paint().

PR #3310, opened 2026-09-08
submodules: bump libretroshare to current master (fixes red CMake CI)

jolavillette:
The six CMake workflows are currently red on master and on every PR, all failing on the same line:

```
PopupDistantChatDialog.cpp:221: error: 'TUNNEL_STATUS_LOCALLY_CLOSED' is not a member of 'RsDistantChatEventCode'
```

Cause: the GUI code merged via #3297 uses `RsDistantChatEventCode::TUNNEL_STATUS_LOCALLY_CLOSED`, introduced by RetroShare/libretroshare#383 — but the recorded libretroshare gitlink (`cafaaa19`) predates that merge. The CMake workflows init submodules at the pinned commit (without `--remote`), so they build a library that lacks the enum value. The legacy workflows use `--remote` and build libretroshare's tip, which is why they stay green while the CMake ones fail.

This PR bumps the pin to libretroshare current master (`2f093fd7b`, which contains libretroshare#383). One gitlink change, no code.


jolavillette:
Superseded by f6ed95ee2, which moved libretroshare to the same commit. Closing.

PR #3311, opened 2026-09-10
🐞Fix: to use available screen geometry for toasters and fix stack spacing

defnax:
The bottom of the desktop is now the top of the taskbar, so even with the default value, the toaster sits cleanly above the taskbar.


defnax:
@csoler

PR #3312, opened 2026-09-11
gui: stop pre-creating the download destination folder tree

SchlauFuchs:
## Problem

Three call sites eagerly `mkpath()` the full destination directory tree the instant a directory/collection download is requested - before any file content has arrived:

- `RetroshareDirModel::downloadDirectory()` (`RemoteDirModel.cpp`)
- `RsCollection::autoDownloadFiles()` (`RsCollection.cpp`)
- `RsCollectionDialog::download()`'s `recursDL` lambda (`RsCollectionDialog.cpp`)
- `SearchDialog::downloadDirectory()` (`SearchDialog.cpp`) - added in a follow-up commit, per review below

All real download bytes are staged in the Partials directory regardless (`ftController::FileRequest()`), and the destination is only touched once a file actually completes, via `RsDirUtil::moveFile()` - which creates any missing destination directories lazily at that point.

So the upfront `mkpath()` calls are redundant, and their only visible effect is leaving empty folders sitting in the download location for the entire duration of the transfer, before any of their contents exist.

## Fix

Drop the redundant `mkpath()` calls. The destination folder tree now only materializes as individual files actually finish downloading, matching how single-file downloads already behave.

## Review update

@jolavillette found the removed calls were incidentally masking two real bugs in `libretroshare`'s directory-creation path (a zero-byte file in a downloaded sub-folder being lost, and an uncaught throwing `create_directories()` on the transfer thread). Both are now fixed in RetroShare/libretroshare#391. The missed `SearchDialog::downloadDirectory()` call site is also now included above.

## Testing

Full clean rebuild of `libretroshare.a` and `retroshare-gui` - links and runs with no new warnings from the touched files.

https://claude.ai/code/session_0137V8RpdwiKisg9SSrMEiBt


jolavillette:
*Pre-review done with an AI under my supervision.*

Worth keeping, but not mergeable alone: the removed `mkpath()` calls were masking two lib holes.

- Zero-byte files in a downloaded sub-folder are lost: the `size == 0` path of `ftController::FileRequest()` writes the stub directly into `dest/fname` and never creates the directory (`ftcontroller.cc:948-966`). #388 keeps that path.
- `RsDirUtil::moveFile()` uses the throwing `std::filesystem::create_directories` (`rsdir.cc:327`) with no catch on the ft thread, so a destination that cannot be created (e.g. a directory name invalid on Windows) aborts the process at completion, and again at every restart. The early `return` this PR removes in `downloadDirectory()` was the only guard on that path.

Both fixes are two lines each in libretroshare (create the directory in the stub, `error_code` overload in `moveFile`). With that companion lib PR, this one is fine. `SearchDialog::downloadDirectory()` keeps the same `mkpath`, it should follow.

SchlauFuchs:
Good catches, thanks. Companion fixes are up in RetroShare/libretroshare#391 (zero-byte-in-subfolder + non-throwing `moveFile()`), and the missed `SearchDialog::downloadDirectory()` call site is now included here too.

jolavillette:
*Pre-review done with an AI under my supervision.*

From my side this is good now: the four eager `mkpath()` sites are gone (incl. `SearchDialog`), nothing else in the download paths pre-creates directories, and the two lib holes this exposes are fixed in RetroShare/libretroshare#388. Two notes for whoever merges: #388 has to land first and the libretroshare pin bumped past it before this one, otherwise master loses zero-byte files in sub-folders and can abort at completion; and the `mirror-to-gitlab.yml` commit is unrelated to the PR, fine by me but csoler may want it separate.

Needs a human review.

PR #3313, opened 2026-09-11
gui: fix friend file list "downloading" color duplicating "already have"

SchlauFuchs:
## Pivoting this PR per review

Thanks @csoler and @jolavillette for the review - dropping the hide-based filter entirely in favor of the color-coding approach you both preferred, and fixing the bug you found along the way.

## What

`RetroshareDirModel::data()`'s `Qt::ForegroundRole` handler for a friend's file list already existed, but returned the **same red** for a file we already have (`rsFiles->alreadyHaveFile()`) and for one currently downloading (present in `rsFiles->FileDownloads()`) - both branches shared one `local_file_color` variable.

That collides with the convention `SearchDialog` already establishes via its `textColorLocal`/`textColorDownloading` properties (red/green by default - `default.qss:285-286`): "already have" should be red, "downloading" should be green, so the two states are visually distinguishable at a glance, consistent with search results.

## Fix

Give the downloading case its own color (`QColor(0, 128, 0)`, matching `SearchDialog`'s own commented-out literal for CSS "green") instead of reusing the already-have one.

Note: the previous commit in this PR (the "Hide files I have" filter) has been dropped entirely and force-pushed out, per the review discussion. This PR now contains only the color fix above.

## Testing

Full clean rebuild of `retroshare-gui`, no new warnings.

https://claude.ai/code/session_0137V8RpdwiKisg9SSrMEiBt


csoler:
There is a color coding scheme already in searched files for this: green = files you have, red = files being transfered. I think we should keep that in friend file lists too for consistency. It's pretty easy to do as it just needs modifying RemoteDirModel::foregroundRole(). Btw that color coding needs fixing in searched files because some columns dont use it whereas all should do.

Hiding the files you have is actually misleading because not seeing a file in one friends' list also means that the friend does not have it, and you cannot know whether this is because you clicked the hide button or not.

jolavillette:
*Pre-review done with an AI under my supervision.*

I agree with the colour approach, and the implementation has a blocking bug anyway: the per-folder totals are keyed by `details.path`, which for a friend's tree is relative to that friend's root (`directory_storage.cc:265`). Every friend's root node has path `""` and same-named top-level folders collide, so the last friend walked decides for all of them: with the box checked the whole friends list can disappear. Toggling also resets the model (tree collapses), and the walk reruns on every `update()` but never on `FRIEND_DIR_LIST_UPDATED`, so folders go stale.

On the colours: the model already returns one for friends' files you have or are downloading (`RetroshareDirModel::data()`, ForegroundRole in remote mode). It is invisible because the Standard/retroclassic skins set `color` on the plain `QTreeView::item` rule, which overrides ForegroundRole for the default delegate. Same reason search results only keep their colours in Size/Age. That is a stylesheet fix, not a model one; I can send it.

Note this PR carries #3312 as its first commit.

SchlauFuchs:
Agreed on both counts - dropped the hide-based filter entirely and fixed the actual bug you found in the existing color-coding instead (the downloading case was reusing the same red as the already-have case, instead of green).

jolavillette:
*Pre-review done with an AI under my supervision.*

Looks good. One thing so the green actually shows: with the default skins the colour returned by the model is overridden by the plain `QTreeView::item { color: … }` rule in `Standard_Light.qss:2257-2263`, `Standard_Dark.qss:2245-2251` (`src/gui/qss/stylesheet/`) and `retroclassic.qss:123-129` (`src/qss/`). Dropping the `color` line from that rule is enough (text colour then comes from `QAbstractItemView { color }` / the palette); it also restores the colours in the Name/Sources/Type/Hash columns of the search results, which csoler mentioned. Might be worth folding into this PR.

.github/workflows/mirror-to-gitlab.yml
```diff
@@ -11,6 +11,11 @@ on:
 jobs:
   mirror:
     runs-on: ubuntu-latest
+    # This mirrors to the official project's GitLab remote and needs a
+    # secret that only exists in RetroShare/RetroShare - skip it entirely
+    # on forks instead of failing (or, worse, trying to push a fork's
+    # history over the real project's GitLab mirror).
+    if: github.repository == 'RetroShare/RetroShare'
```
csoler: Irrelevant to this PR. Remove it.

PR #3314, opened 2026-09-11
Moved Windows specific code for getting main executable path from ret…

thunder2:
…roshare-service.cc to an own file as function getMainExecutablePath


csoler:
nice! It can be merged right away imo.

